Issue #37, opened 2019-08-26
[FEATURE] Add Security support


jiazhai:
Add security support for KOP

PR #47, opened 2019-11-18
Add Security support (streamnative#37)

PierreZ:
This PR is adding Security within KoP. All auth is forwarded to Pulsar's auth mechanisms. This PR is also adding the support of `SASL_PLAIN` to inject the auth:

* user correspond to the namespace
* password corresponds to the authorization provider and the authorization payload


jiazhai:
https://github.com/streamnative/kop/issues/37 is related.

PR #139, opened 2020-04-13
Sign all release assets with bot key

zymap:
---

*Motivation*

Sign all release assets and provide GPG signature and hashsum assets.


zymap:
Close this because I will use another way to do this.

PR #158, opened 2020-05-29
Add kafka-python 1.3.3 test case

hangc0276:
Add kafka-python 1.3.3 test case

### DIFF
--- a/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaIntegrationTest.java
+++ b/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaIntegrationTest.java
@@ -269,7 +269,7 @@ public class KafkaIntegrationTest extends KopProtocolHandlerTestBase {
     @Test(timeOut = 3 * 60_000, dataProvider = "pyKafkaIntegrations")
     void simplePyKafkaPythonProduceAndConsume(final String integration, final Optional<String> topic,
                                  final boolean shouldProduce, final boolean shouldConsume,
-                                 final boolean balance_consumer, final boolean use_rdkafka,
+                                 final boolean balanceConsumer, final boolean useRdkafka,
                                  final String groupId, final String requirements,
                                  final String command) throws Exception {
         String topicName = topic.orElse(integration);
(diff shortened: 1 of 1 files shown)


sijie:
@hangc0276 Can you move the changes under `integrations/python2-kafka-python/*` a separate pull request? So we can get the dockerfile change merged and a new image can be pushed to docker hub. Then your integration test can use it.

hangc0276:
> @hangc0276 Can you move the changes under `integrations/python2-kafka-python/*` a separate pull request? So we can get the docerfile change merged and a new image can be pushed to docker hub. Then your integration test can use it.

@sijie Thanks for your reply. I have changed `kafka-python` test case dockerfile to `integrations/python2-kafka-python/` and all the versions of `kafka-python` will use the same docker image to run test cases.

sijie:
@hangc0276 correct. the issue failed with the following error `Caused by: com.github.dockerjava.api.exception.NotFoundException: {"message":"pull access denied for streamnative/kop-test-python2-kafka-python, repository does not exist or may require 'docker login': denied: requested access to the resource is denied"}`

It means we didn't have the image yet. My suggestion is to create a separate pull request that only includes the docker change. So we can merge that pull request and get the docker image published to streamnative/kop-test-python2-kafka-python. So the integration tests in this pull request can use that image.

hangc0276:
> @hangc0276 correct. the issue failed with the following error `Caused by: com.github.dockerjava.api.exception.NotFoundException: {"message":"pull access denied for streamnative/kop-test-python2-kafka-python, repository does not exist or may require 'docker login': denied: requested access to the resource is denied"}`
> 
> It means we didn't have the image yet. My suggestion is to create a separate pull request that only includes the docker change. So we can merge that pull request and get the docker image published to streamnative/kop-test-python2-kafka-python. So the integration tests in this pull request can use that image.

@sijie OK, i will use a separate PR to build the docker image.

sijie:
@hangc0276 any updates on this pull request?

hangc0276:
> @hangc0276 any updates on this pull request?

Yes, I will update the docker image soon.

hangc0276:
@sijie  I have update the `kafka-python` and `pykafka` integration test framework. Some versions are not supported, and I have exclude them in the integration test.

jiazhai:
synced with @hangc0276 , Test failure caused by the docker image, maybe we should fix the image first:
image

Issue #166, opened 2020-07-06
[BUG] java.lang.IllegalArgumentException: Invalid version for API key API_VERSIONS: 3


sido420:
**Describe the bug**
Pulsar starts fine with Kafka protocol handler installed however I see error messages in logs.

```
13:12:48.091 [ForkJoinPool.commonPool-worker-3] INFO  org.apache.pulsar.broker.service.ServerCnx - [/127.0.0.1:56110] Created subscription on topic persistent://public/default/exclamation-input / public/default/exclamation
13:12:57.442 [pulsar-io-25-13] INFO  io.streamnative.pulsar.handlers.kop.KafkaRequestHandler - channel inactive [id: 0x17ea1bee, L:/127.0.0.1:9092 ! R:/127.0.0.1:56113]
13:12:57.443 [pulsar-io-25-13] INFO  io.streamnative.pulsar.handlers.kop.KafkaRequestHandler - close channel [id: 0x17ea1bee, L:/127.0.0.1:9092 ! R:/127.0.0.1:56113]
13:12:58.158 [pulsar-io-25-3] INFO  io.streamnative.pulsar.handlers.kop.KafkaRequestHandler - channel active: [id: 0x62a5187c, L:/127.0.0.1:9092 - R:/127.0.0.1:56128]
13:12:58.159 [pulsar-io-25-3] ERROR io.streamnative.pulsar.handlers.kop.KafkaCommandDecoder - error to get Response ByteBuf:
java.lang.IllegalArgumentException: Invalid version for API key API_VERSIONS: 3
	at org.apache.kafka.common.protocol.ApiKeys.schemaFor(ApiKeys.java:312) ~[?:?]
	at org.apache.kafka.common.protocol.ApiKeys.responseSchema(ApiKeys.java:286) ~[?:?]
	at org.apache.kafka.common.requests.ApiVersionsResponse.toStruct(ApiVersionsResponse.java:129) ~[?:?]
	at org.apache.kafka.common.requests.ResponseUtils.serializeResponse(ResponseUtils.java:40) ~[?:?]
	at io.streamnative.pulsar.handlers.kop.KafkaCommandDecoder.responseToByteBuf(KafkaCommandDecoder.java:112) ~[?:?]
	at io.streamnative.pulsar.handlers.kop.KafkaCommandDecoder.writeAndFlushResponseToClient(KafkaCommandDecoder.java:238) ~[?:?]
[...]

Issue #153, opened 2020-05-20
[FEATURE] Pulsar Kop should support Kafka 2.1+ version 


fenghao007:
Currently Kop Only Support Kafka 2.0.  The latest Kafka version has come to 2.5. 
The most important is since Kafka 2.1,  there are some changes  for the schema used to store consumer offsets  which makes current Kop cannot support Kafka 2.1 and all subsequent incremental version .   [Kafka schema changed related docs . ](https://kafka.apache.org/21/documentation.html#upgrade)
I hava perfomed some  kop tests with @codelipenghui  , Using  Kafka 2.3 and Kafka 2.0 clients .   Kafka 2.0 client work well .
While  Kafka 2.3 clients , it will always give a following error which Verified my conjecture. 

`06:48:16.138 [pulsar-io-22-14] ERROR io.streamnative.pulsar.handlers.kop.KafkaRequestHandler - Caught error in handler, closing channel
io.netty.handler.codec.TooLongFrameException: Adjusted frame length exceeds 104857600: 369295621 - discarded
        at io.netty.handler.codec.LengthFieldBasedFrameDecoder.fail(LengthFieldBasedFrameDecoder.java:513) ~[io.netty-netty-codec-4.1.43.Final.jar:4.1.43.Final]
`

Now we're going to quickly  immigrate our production to pulsar with kop . It can really reduce the cost of immigration esp. it  can be  transparent to business.  Kop not support Kafka 2.3 has blocked us . 
What's more , Kop need this support  Kafka 2.1+  , it may be more popular and used  .  @jiazhai @sijie

sijie:
@fenghao007 noted. we will prioritize to add the support for Kafka 2.1+ clients.

thinker0:
+1

jsyvanen:
+1

jiazhai:
This is already done, will close this issue

Issue #99, opened 2020-02-12
[FEATURE] Add support for APIVersions 3


PierreZ:
**Is your feature request related to a problem? Please describe.**
[This commit in librdkakfa](https://github.com/edenhill/librdkafka/commit/be484c6cce82dfb052ccfd145225c4d2f14962f2#diff-d66fa525591da8385b21fcab72e24f28) is removing support of older versions of APIVersions to use only v3. As APIVersions is almost the first frame sent by clients, we need to support it to enable support of future librdkafka's version.

**Describe the solution you'd like**
We need to bump [kafka-clients version](https://github.com/streamnative/kop/blob/master/pom.xml#L37) to at least 2.4 as encoding and decoding is handled through this dep.

PR #176, opened 2020-08-26
Support for Kafka-Client versions 2.4.x and above

ptirupac-tibco:
**Fixes:**
 #166 #153 #99 

**Issue:**
KoP uses [Kafka-2.0.0](https://github.com/streamnative/kop/blob/78d9ba3487d4d7c85a5d667d45d9b38aaa7c824f/pom.xml#L46) which supports [API_VERSION's](https://kafka.apache.org/protocol.html#The_Messages_ApiVersions) **0** --> **2**

When **_Kafka-Clients-2.4.x+_**(using `API_VERSION:  3`) connects to KoP, it panics and following error stack is observed:
`10:22:23.281 [pulsar-io-22-4] ERROR io.streamnative.pulsar.handlers.kop.KafkaCommandDecoder - error to get Response ByteBuf:
java.lang.IllegalArgumentException: Invalid version for API key API_VERSIONS: 3
    at org.apache.kafka.common.protocol.ApiKeys.schemaFor(ApiKeys.java:312) ~[?:?]
    at org.apache.kafka.common.protocol.ApiKeys.responseSchema(ApiKeys.java:286) ~[?:?]
    at org.apache.kafka.common.requests.ApiVersionsResponse.toStruct(ApiVersionsResponse.java:129) ~[?:?]
    at org.apache.kafka.common.requests.ResponseUtils.serializeResponse(ResponseUtils.java:40) ~[?:?]`

**Resolved By:**
Returning an `UNSUPPORTED_VERSION` [error-code: 35](https://kafka.apache.org/protocol.html#protocol_error_codes), which would notify the **_kafka-client_** to lower it's `API_VERSION`. As no list of `ApiKeys & versions` were available for the **kafka-clients** to refer, it safely falls-back to using `API_VERSION:  0` and KoP continues processing the kafka-messages using `API_VERSION:  0`.

**Tested producing/consuming with Kafka-Clients:**

> 2.0.0
2.2.2
2.3.1
2.4.0
2.4.1
2.5.0
2.5.1
2.6.0

**More...**
KoP could have returned the list of supported `ApiKeys & versions` while sending the `UNSUPPORTED_VERSION` error-code, which would have made the **_kafka-client_** select the **_latest_** supported `API_VERSION` and use `API_VERSION:  2` instead of falling all the way back and using `API_VERSION:  0` 

Notes on how **_Kafka-Brokers_** is supposed to handle this scenario: 
> 2. On receiving ApiVersionsRequest, a broker returns its full list of supported ApiKeys and versions regardless of current authentication state (e.g., before SASL authentication on an SASL listener, do note that no Kafka protocol requests may take place on an SSL listener before the SSL handshake is finished). If this is considered to leak information about the broker version a workaround is to use SSL with client authentication which is performed at an earlier stage of the connection where the ApiVersionRequest is not available. Also, note that broker versions older than 0.10.0.0 do not support this API and will either ignore the request or close connection in response to the request.
> 
> 3. If multiple versions of an API are supported by broker and client, clients are recommended to use the latest version supported by the broker and itself.

_Reference: [Kafka-Protocol Guide](https://kafka.apache.org/protocol.html#api_versions)_

We analyzed how various **_Kafka-Brokers_** respond to a similar situation where the **_kafka-client's_** `API_VERSION` is higher than what is supported by the **_Kafka-Broker_**.

Kafka-Broker-Client-Wireshark-Results
_Reference: Wireshark packet captures - [Kafka-Protocol-Study.zip](https://github.com/streamnative/kop/files/5127018/Kafka-Protocol-Study.zip)_

From the study we can infer that, in a similar `API_VERSION` mismatch scenario the **_Kafka-Brokers_** doesn't return the list of supported `ApiKeys & versions` when notifying the **_kafka-client_** with the `UNSUPPORTED_VERSION` [error-code: 35](https://kafka.apache.org/protocol.html#protocol_error_codes). Thus, forcing the **_kafka-clients_** to fall-back to using `API_VERSION:  0`.

To keep KoP working in sync with the **_Kafka-Broker_** working, we decided not to return the list of supported `ApiKeys & versions`. 

Thoughts and suggestions?

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
@@ -108,9 +108,17 @@ public abstract class KafkaCommandDecoder extends ChannelInboundHandlerAdapter {
     protected ByteBuf responseToByteBuf(AbstractResponse response, KafkaHeaderAndRequest request) {
         try (KafkaHeaderAndResponse kafkaHeaderAndResponse =
                  KafkaHeaderAndResponse.responseForRequest(request, response)) {
-
+            // Lowering Client API_VERSION request to the oldest API_VERSION KoP supports, this is to make \
+            // Kafka-Clients 2.4.x and above compatible and prevent KoP from panicking \
+            // when it comes across a higher API_VERSION.
+            short apiVersion = kafkaHeaderAndResponse.getApiVersion();
+            if (request.getHeader().apiKey() == API_VERSIONS){
+                if (!ApiKeys.API_VERSIONS.isVersionSupported(apiVersion)) {
+                    apiVersion = ApiKeys.API_VERSIONS.oldestVersion();
+                }
+            }
             return ResponseUtils.serializeResponse(
-                kafkaHeaderAndResponse.getApiVersion(),
+                apiVersion,
                 kafkaHeaderAndResponse.getHeader(),
                 kafkaHeaderAndResponse.getResponse()
             );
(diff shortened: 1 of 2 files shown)


jiazhai (approved):

jiazhai:
@ptirupac-tibco Thanks for the great help. lgtm.

PR #183, opened 2020-09-23
Support Pulsar 2.6.1

BewareMyPower:
Current version of KoP is not compatible with Pulsar 2.6.1, update the dependency to support it.


BewareMyPower:
There're some compatibility problems, I'll handle them soon.

jiazhai (approved):

Issue #184, opened 2020-09-23
[BUG] DistributedClusterTest CI failed but local tests passed


BewareMyPower:
See #183 

The DistributedClusterTest CI failed with Github Action:

```
2020-09-23T08:54:33.4397994Z [INFO] Running io.streamnative.pulsar.handlers.kop.DistributedClusterTest
2020-09-23T08:54:55.9560950Z [ERROR] Tests run: 3, Failures: 3, Errors: 0, Skipped: 0, Time elapsed: 22.468 s <<< FAILURE! - in io.streamnative.pulsar.handlers.kop.DistributedClusterTest
2020-09-23T08:54:55.9563904Z [ERROR] testMutiBrokerAndCoordinator(io.streamnative.pulsar.handlers.kop.DistributedClusterTest)  Time elapsed: 0.225 s  <<< FAILURE!
2020-09-23T08:54:55.9565737Z java.lang.AssertionError: expected [2] but found [1]
2020-09-23T08:54:55.9566540Z 	at org.testng.Assert.fail(Assert.java:96)
2020-09-23T08:54:55.9567449Z 	at org.testng.Assert.failNotEquals(Assert.java:776)
2020-09-23T08:54:55.9568858Z 	at org.testng.Assert.assertEqualsImpl(Assert.java:137)
2020-09-23T08:54:55.9569889Z 	at org.testng.Assert.assertEquals(Assert.java:118)
2020-09-23T08:54:55.9570766Z 	at org.testng.Assert.assertEquals(Assert.java:652)
2020-09-23T08:54:55.9571619Z 	at org.testng.Assert.assertEquals(Assert.java:662)
2020-09-23T08:54:55.9575151Z 	at io.streamnative.pulsar.handlers.kop.DistributedClusterTest.testMutiBrokerAndCoordinator(DistributedClusterTest.java:293)
2020-09-23T08:54:55.9578305Z 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
2020-09-23T08:54:55.9580208Z 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
2020-09-23T08:54:55.9582275Z 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
2020-09-23T08:54:55.9584712Z 	at java.lang.reflect.Method.invoke(Method.java:498)
[...]

PR #187, opened 2020-09-24
use kafkaTenant and kafkaNamespace when OffsetAcker create consumer

### LINKED ISSUE #186: [BUG] OffserAcker always use topics under public/default namespaces
**Describe the bug**

`OffsetAcker` always use the `public/default`  as namespaces name when create consumer.

**To Reproduce**
1. set `kafkaTenant`  in broker.conf to not `public`
2. set `kafkaNamespace ` in broker.conf to not `default`
3. start a kafka consumer 

**Expected behavior**
`OffsetAcker` should use the `kafkaTenant` and `kafkaNamespace` as set in the broker.conf instead of `public`
 and `defualt`

**Additional context**
  [Relative code](https://github.com/streamnative/kop/blob/master/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/OffsetAcker.java#L116)

aloyszhang:
fix #186 
use `kafkaTenant` and `kafkaNamespace` set in broker.conf instead of `public` and `default` for `OffsetAcker ` when create consumer

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
@@ -316,6 +316,7 @@ public class KafkaProtocolHandler implements ProtocolHandler {
             (PulsarClientImpl) (service.pulsar().getClient()),
             groupConfig,
             offsetConfig,
+            kafkaConfig,
             SystemTimer.builder()
                 .executorName("group-coordinator-timer")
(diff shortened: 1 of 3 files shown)


BewareMyPower:
It looks like that there's some wrong with `DistributedClusterTest`, see #184. Can you pass these tests in your local environment?

aloyszhang:
Local test runs well.

jiazhai (approved):

jiazhai:
@BewareMyPower we could merge it currently, since the test failure already tracked

Issue #138, opened 2020-04-09
[TEST] Add more integration tests to cover different versions of Kafka clients


sijie:
Currently, most of the integration tests are only verified on Kafka 2.0 clients. Can we add more integration tests to cover more Kafka clients?

jiazhai:
close for pr merged #194 #196

PR #194, opened 2020-10-16
[WIP] Add Kafka version coverage tests

BewareMyPower:
Master issue: #138 

Use one code to build multiple docker images for Kafka client version 1.x.0 and 2.y.0.

Kafka client 0.11.0.0 could also pass the `simpleProduceAndConsume` tests but from the log we can see the 0.11.0.0 consumer cannot commit offsets. Therefore it's treated as not supported by KoP.

In additions, the current integration tests cannot be run locally (at least with MacOS) because the advertised address was set to localhost. I fix this problem by replacing it with a site local address.

----

Currently, it has some problems with Kafka client tests. When the container of Kafka client exits, it doesn't print `ExitCode: 0` so exceptions would be thrown from:

https://github.com/streamnative/kop/blob/41f274ccd629c2e18edd33ccecb50c70ee593257/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaIntegrationTest.java#L204

On the other hand, it seems to have a race condition in https://github.com/streamnative/kop/blob/41f274ccd629c2e18edd33ccecb50c70ee593257/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaIntegrationTest.java#L192-L193. The race condition applies in current integration tests.

### DIFF
--- a/integrations/build.sh
+++ b/integrations/build.sh
@@ -20,11 +20,24 @@ VERSION=$(${INTR_HOME}/scripts/dev/get-project-version.py)
 TAG=${VERSION%"-SNAPSHOT"}
 IMAGE_NAME_PREFIX="kop-test-"
 
-for img_dir in `ls -d ${INTR_HOME}/integrations/*/ | grep -v dev`; do
-    BASE_NAME=$(basename ${img_dir})
-    cd ${img_dir}
-    IMAGE="streamnative/${IMAGE_NAME_PREFIX}${BASE_NAME}:${TAG}"
+build_image() {
+    IMAGE=$1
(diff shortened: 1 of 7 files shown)


BewareMyPower:
From https://github.com/streamnative/kop/pull/158#issuecomment-637199456 , I have to use a separate PR to build the new docker images, so I close this PR first.

As for the problems with Kafka client tests
- The `ExitCode` problem is the `Dockerfile`'s execute command doesn't run `echo "ExitCode: $?`, the fix will be in the new PR later.
- The race condition could be verified by a simple test, I'll open a new issue.

PR #198, opened 2020-10-20
[WIP] [DEBUG] Run CI on DistributedClusterTest only

BewareMyPower:
**DON'T MERGE!**

Try to fix issue #184 , since the tests passed in local env but failed in github CI. Here we remove other tests first for debugging quickly.


BewareMyPower:
Close it because the reason is found in https://github.com/streamnative/kop/issues/252

Issue #199, opened 2020-10-21
[BUG] java.lang.IllegalStateException: Correlation id for response (4702) does not match request (4697)


dockerzhang:
**Describe the bug**
java.lang.IllegalStateException: Correlation id for response (4702) does not match request (4697)

**To Reproduce**
produce message using:
`./bin/kafka-producer-perf-test.sh --topic lookup3 --num-records 1000000000 --record-size 100 --throughput -1 --producer-props acks=1 bootstrap.servers=127.0.0.1:9092`

**Screenshots**
![image](https://user-images.githubusercontent.com/18047329/96667159-15ad6880-138b-11eb-99d1-53df9e511fb6.png)

PR #200, opened 2020-10-21
keep requestsQueue order

dockerzhang:
related to https://github.com/streamnative/kop/issues/199, keep requestsQueue order using one thread to solve Correlation ID response in disorder.


BewareMyPower:
Could you add some explanation for why calling `writeAndFlushResponseToClient` in the channel's event loop can solve response disorder issue?

dockerzhang:
keep `requestsQueue`  in `writeAndFlushResponseToClient` peeked using the same thread.

jiazhai (approved):

Issue #202, opened 2020-10-21
[BUG] describe group failed by kafka-consumer-groups.sh


dockerzhang:
**Describe the bug**
kafka client:
`There is no leader for this topic-partition as we are in the middle of a leadership election`

kop server:
```
[2020-10-21 14:07:52,040 ERROR [pulsar-io-23-24] (kop.KafkaCommandDecoder@channelRead:222): error while handle command:
java.lang.NullPointerException: null
        at java.util.Objects.requireNonNull(Objects.java:203) ~[?:1.8.0_252]
        at java.util.Optional.<init>(Optional.java:96) ~[?:1.8.0_252]
        at java.util.Optional.of(Optional.java:108) ~[?:1.8.0_252]
        at io.streamnative.pulsar.handlers.kop.KafkaRequestHandler.handleOffsetFetchRequest(KafkaRequestHandler.java:708) ~[pulsar-protocol-handler-kafka-0.3.1-SNAPSHOT.nar-unpacked/:?]
        at io.streamnative.pulsar.handlers.kop.KafkaCommandDecoder.channelRead(KafkaCommandDecoder.java:176) [pulsar-protocol-handler-kafka-0.3.1-SNAPSHOT.nar-unpacked/:?]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:377) [io.netty-netty-transport-4.1.45.Final.jar:4.1.45.Final]](url)
```

**To Reproduce**
describe consumer group:
`bin/kafka-consumer-groups.sh --bootstrap-server 127.0.0.1:9092 --describe --group  docker-group`

**Expected behavior**
show the detail of consumer group.

**Screenshots**
![image](https://user-images.githubusercontent.com/18047329/96680048-5c10c080-13a7-11eb-9568-c18d573418b4.png)

PR #203, opened 2020-10-21
fix fetch offset when topic is null

dockerzhang:
bug fix for https://github.com/streamnative/kop/issues/202


jiazhai (approved):

jiazhai:
@dockerzhang Thanks for the help on this. it would be better to add an test case for this fix.

PR #205, opened 2020-10-21
Fix group coordinator's metadata never expired

BewareMyPower:
The `GroupCoordinator`'s startup method accepts a boolean param `enableMetadataExpiration`, which will be passed to `GroupMetadataManager#startup`. If it's true, the group metadata manager will clean up group metadata cache periodically.

However `enableMetadataExpiration` is false now. It should only be set false in tests.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
@@ -328,7 +328,7 @@ public class KafkaProtocolHandler implements ProtocolHandler {
 
     public void startGroupCoordinator() throws Exception {
         if (this.groupCoordinator != null) {
-            this.groupCoordinator.startup(false);
+            this.groupCoordinator.startup(true);
         } else {
             log.error("Failed to start group coordinator. Need init it first.");
         }


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
```diff
@@ -328,7 +328,7 @@ public void initGroupCoordinator(BrokerService service) throws Exception {
 
     public void startGroupCoordinator() throws Exception {
         if (this.groupCoordinator != null) {
-            this.groupCoordinator.startup(false);
+            this.groupCoordinator.startup(true);
```
jiazhai, line 331: Should we make it a configurable parameter?

BewareMyPower: The `startup` method only start a regular group metadata cache cleanup work whose interval is determined by offset config `offsetsRetentionCheckIntervalMs`, see https://github.com/streamnative/kop/blob/d684f4bd295ed686aa6c6f371864f29f91c02e00/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupMetadataManager.java#L239-L247

The `false` option is only used in unit test to disable the cleanup work. By the way, the group coordinator was ported from Kafka's code, it's just a mistake before.

jiazhai (approved):

PR #210, opened 2020-10-30
Update release process

BewareMyPower:
Master issue: #211

### DIFF
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -35,6 +35,8 @@ There has two type of the tags, one is stable `vX.Y.Z(.M)`, and another is unsta
 
 2. Update the project version and tag
 
+   > If the dependency of Apache Pulsar needs to change, i.e. the project version was from `X1.Y1.Z1.M` to `X2.Y2.Z2.0`, you should change the dependency of Apache Pulsar first.
+
    ```bash
    $ ./scripts/set-project-version.sh X.Y.Z.M
    $ git commit -m "Release X.Y.Z.M" -a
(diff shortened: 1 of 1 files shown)


docs/release-process.md
```diff
@@ -2,96 +2,78 @@
...
 
-We are going to create a branch from `master` to `branch-x.y`
-where the tag will be generated and where new fixes will be
-applied as part of the maintenance for the release. `x.y.z`
-is the version of the release.
+There has two type of the tags, one is stable `vX.Y.Z(.M)`, and another is unstable `vX.Y.Z(.M)-rc-$(date +%y%m%d%H%M)`. A stable tag represent this release is a verified release, and an unstable tag represent this release is not verified. 
 
-The branch needs only to be created when creating major releases,
-and not for patch releases.
+## Release workflow
```
zymap, line 17: Sorry, maybe I haven't clearly pointed out, all I mentioned in this article https://streamnative.slab.com/posts/repository-release-template-prb0lvyt is a **template**. I just want to make a template and we can apply it to any of our repositories. According to the release template, we should have a clear release process.

I see you just copy it here, does this workflow can release kop step by step?

BewareMyPower: Ok, I'll push the fix soon

BewareMyPower:
I've updated the doc, PTAL, @zymap .

By the way, I have a question that when should we use `/snbot` of streamnative-ci?

jiazhai (approved):

Issue #212, opened 2020-11-03
[FEATURE] Cursor read more entries one time when fetching message


dockerzhang:
**Is your feature request related to a problem? Please describe.**
KoP consumer performance is not so quick compared to Kafka.

**Describe the solution you'd like**
Cursor read more entries one time when fetching the message.

PR #213, opened 2020-11-03
Cursor read more entries one time when fetching message

dockerzhang:
feature related to https://github.com/streamnative/kop/issues/212, it could improve the consuming speed at least 30%.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
@@ -245,8 +245,8 @@ public class KafkaServiceConfiguration extends ServiceConfiguration {
 
     @FieldContext(
             category = CATEGORY_KOP,
-            doc = "the number of reading entries from bookie"
+            doc = "maximum number of entries that are read from cursor once per time"
     )
-    private int readEntriesNum = 5;
+    private int maxReadEntriesNum = 5;
 
 }
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
@@ -243,4 +243,10 @@ public class KafkaServiceConfiguration extends ServiceConfiguration {
     )
     private Set<String> saslAllowedMechanisms = new HashSet<String>();
 
+    @FieldContext(
+            category = CATEGORY_KOP,
+            doc = "the number of reading entries from bookie"
+    )
+    private int readEntriesNum = 10;
+
 }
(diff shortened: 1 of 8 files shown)


tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaApisTest.java
```diff
@@ -498,7 +498,7 @@ public void testBrokerRespectsPartitionsOrderAndSizeLimits() throws Exception {
 
         int messagesPerPartition = 9;
         int maxResponseBytes = 800;
-        int maxPartitionBytes = 190;
+        int maxPartitionBytes = 900;
```
BewareMyPower, line 501: Why did you do this change? Does `MessageFetchContext#readMessagesInternal` affect this test? This unit test is a known flaky test, I tried to change the `maxPartitionBytes` in my local environment but it still failed often.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
```diff
@@ -243,4 +243,10 @@
     )
     private Set<String> saslAllowedMechanisms = new HashSet<String>();
 
+    @FieldContext(
+            category = CATEGORY_KOP,
+            doc = "the number of reading entries from bookie"
```
BewareMyPower, line 248: I think it's better to change the doc to `maximum number of entries that are read from cursor once per time` to note that the entries are read from cursors and the number is a **maximum** number as `ManagedCursor#readEntries`' doc says. The same goes with the name of `readEntriesNum`. Otherwise the config may be a little confused.

BewareMyPower (changes requested):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaApisTest.java
dockerzhang: UT failed in https://github.com/streamnative/kop/runs/1345779596, and I didn't get the reason. I thought it may be entryNum * 190, and UT passed after a change to 900.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
dockerzhang: `readEntriesNum -> maxReadEntriesNum` is ok?

BewareMyPower: I think `maxReadCursorEntriesNum` is better

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaApisTest.java
BewareMyPower: It's a flaky test, the reason is still unknown. But it looks like not related to this param. So it's better to just keep the old value, I will rerun the CI if failed.

BewareMyPower:
The build failed because the relative unit tests didn't change after the config item's name changed. By the way, the doc in `*.conf` should also be updated.

BewareMyPower (approved):

Issue #222, opened 2020-11-04
[FEATURE] Support publish buffer throttle for KOP


aloyszhang:
**Describe the solution you'd like**
Currently , pulsar supports throttle on publish buffer.  KOP should support this feature too.

PR #223, opened 2020-11-04
Support publish buffer throttle for kop

aloyszhang:
#222 
Support publish buffer throttle for kop .

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/InternalProducer.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/InternalProducer.java
@@ -27,10 +27,12 @@ import org.apache.pulsar.broker.service.Topic;
  */
 @Slf4j
 public class InternalProducer extends Producer {
+    private ServerCnx serverCnx;
     public InternalProducer(Topic topic, ServerCnx cnx,
                             long producerId, String producerName) {
         super(topic, cnx, producerId, producerName, null,
             false, null, null, 0, false);
+        this.serverCnx = cnx;
     }
 
     // this will call back by bundle unload
@@ -49,5 +51,18 @@ public class InternalProducer extends Producer {
         return future;
     }
 
+    @Override
+    public ServerCnx getCnx() {
+        return serverCnx;
+    }
 
+    @Override
+    public boolean equals(Object obj) {
+        return super.equals(obj);
+    }
+
+    @Override
+    public int hashCode() {
+        return super.hashCode();
+    }
 }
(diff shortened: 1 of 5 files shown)


aloyszhang:
@BewareMyPower Is ther any way to re-run the checks?

BewareMyPower:
As far as I know there's not any script like `pulsarbot`. It requires a committer to manually rerun the workflow.

BewareMyPower:
By the way, it's better to run the simple tests in your local environment before pushing the commits:

```bash
$ mvn checkstyle:check
$ mvn spotbugs:check
# The following tests may take a few time, you can rely the CI to detect errors
$ mvn test -DfailIfNoTests=false '-Dtest=!KafkaIntegrationTest,!DistributedClusterTest'
$ mvn test '-Dtest=KafkaIntegrationTest' -pl tests
```

aloyszhang:
@BewareMyPower Thanks for you reply.  I'll check style and spotbug locally first.

BewareMyPower:
Could you add a unit test to verify the throttle like [MessagePublishThrottlingTest.java](https://github.com/apache/pulsar/pull/3986/files#diff-27c437c82514b1c1bc5f1f909b0c7e71eb5d8571ac728103157d62621de154ac)? In addition, the Pulsar version is 2.6.2 now, please rebase to the latest commit.

aloyszhang:
@BewareMyPower It's hard to test in kop for this feature since the the logic of publish buffer throttling are implement in Pulsar.

aloyszhang:
Addition,   we can not use `BrokerService#checkMessagePublishBuffer()` in kop.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/MessagePublishBufferThrottle.java
```diff
@@ -0,0 +1,114 @@
...
+        try {
+            producer.send(new ProducerRecord<>(topic, new byte[1024])).get(1, TimeUnit.SECONDS);
+            Assert.fail("Send operation should failed because of timeout.");
+        } catch (TimeoutException e) {
+            // Expected
+        }
+
+
+        internalServerCnx.setMessagePublishBufferSize(0);
+        Thread.sleep(10000);
```
BewareMyPower: You should add comments here to indicate `sleep` is just for waiting until `checkMessagePublishBuffer` was called. Here's an example: https://github.com/apache/pulsar/blob/3142456107e8bc5d3d0d7fa526fda114c0d64c42/pulsar-broker/src/test/java/org/apache/pulsar/client/api/MessageDispatchThrottlingTest.java#L262

And use `conf.getMessagePublishBufferCheckIntervalInMillis()` here is more readable than hard coded 10000. By the way, IMO, just keep the default value (100) here is enough. You can just wait a little longer if you worried about the CI failure.

BewareMyPower (changes requested):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/MessagePublishBufferThrottle.java
aloyszhang: you are right

BewareMyPower (approved):

Issue #220, opened 2020-11-04
[BUG] a lot of warnings about “Failed to getOrCreateTopic null. Topic is in loading status”  after the topic unload or broker restart


dockerzhang:
**Describe the bug**
a lot of warnings of “Failed to getOrCreateTopic null. Topic is in loading status”  after the topic unload or broker restart

**To Reproduce**
1.  deploy a kop cluster with 2-3 nodes
2.  produce message using 100 librdkafka client 
3. unload topic or restart a broker node
4. active broker and the restart broker see error

**Expected behavior**
broker and existed client recover quickly

**Screenshots**
![image](https://user-images.githubusercontent.com/18047329/98066107-f46f7080-1e90-11eb-8dbb-74897cf39a66.png)

dockerzhang:
for consuming message, a lot of exception about “topic not owned, and return null TCM in fetch” after the topic unload or broker restart.
![image](https://user-images.githubusercontent.com/18047329/98094682-90ff3600-1ec4-11eb-9a77-1a7312b4090b.png)

PR #225, opened 2020-11-05
refactor lookup cache to solve the recover of client

dockerzhang:
refactor lookup cache to solve the recovery of the client, relate to https://github.com/streamnative/kop/issues/220.

### DIFF
--- a/tests/pom.xml
+++ b/tests/pom.xml
@@ -30,6 +30,10 @@
   <name>StreamNative :: Pulsar Protocol Handler :: KoP Tests</name>
   <description>Tests for Kafka on Pulsar</description>
 
+  <properties>
+    <schema.registry.version>5.0.0</schema.registry.version>
+  </properties>
+
   <!-- include the dependencies -->
   <dependencies>
     <dependency>
@@ -52,6 +56,33 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>io.confluent</groupId>
+      <artifactId>kafka-schema-registry</artifactId>
+      <version>${schema.registry.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-log4j12</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.glassfish.jersey.core</groupId>
+          <artifactId>jersey-common</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
+    <dependency>
+      <groupId>io.confluent</groupId>
+      <artifactId>kafka-avro-serializer</artifactId>
+      <version>${schema.registry.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-log4j12</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
   </dependencies>
 
   <build>
@@ -78,4 +109,10 @@
     </plugins>
   </build>
 
+  <repositories>
+    <repository>
+      <id>io-confluent</id>
+      <url>https://packages.confluent.io/maven/</url>
+    </repository>
+  </repositories>
 </project>
--- a/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KopProtocolHandlerTestBase.java
+++ b/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KopProtocolHandlerTestBase.java
@@ -22,6 +22,9 @@ import com.google.common.collect.Sets;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
+import io.confluent.kafka.schemaregistry.avro.AvroCompatibilityLevel;
+import io.confluent.kafka.schemaregistry.rest.SchemaRegistryConfig;
+import io.confluent.kafka.schemaregistry.rest.SchemaRegistryRestApplication;
 import io.streamnative.pulsar.handlers.kop.utils.MetadataUtils;
 
 import java.io.Closeable;
@@ -73,6 +76,7 @@ import org.apache.zookeeper.CreateMode;
 import org.apache.zookeeper.MockZooKeeper;
 import org.apache.zookeeper.ZooKeeper;
 import org.apache.zookeeper.data.ACL;
+import org.eclipse.jetty.server.Server;
 
 /**
  * Unit test to test KoP handler.
@@ -96,6 +100,8 @@ public abstract class KopProtocolHandlerTestBase {
     protected int kafkaBrokerPort = PortManager.nextFreePort();
     @Getter
     protected int kafkaBrokerPortTls = PortManager.nextFreePort();
+    @Getter
+    protected int kafkaSchemaRegistryPort = PortManager.nextFreePort();
 
     protected MockZooKeeper mockZooKeeper;
     protected NonClosableMockBookKeeper mockBookKeeper;
@@ -107,6 +113,13 @@ public abstract class KopProtocolHandlerTestBase {
 
     private SameThreadOrderedSafeExecutor sameThreadOrderedSafeExecutor;
     private ExecutorService bkExecutor;
+
+    // Fields about Confluent Schema Registry
+    protected boolean enableSchemaRegistry = false;
+    private static final String KAFKASTORE_TOPIC = SchemaRegistryConfig.DEFAULT_KAFKASTORE_TOPIC;
+    protected SchemaRegistryRestApplication restApp;
+    protected Server restServer;
+    protected String restConnect;
 
     public KopProtocolHandlerTestBase() {
         resetConfig();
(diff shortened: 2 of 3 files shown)


BewareMyPower:
The spot bugs check failed.

Tips: Before pushing the PR, you can run the simple check locally for code style and spot bugs to make CI not fail before the unit tests:

```bash
$ mvn checkstyle:check
$ mvn spotbugs:check
```

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
```diff
@@ -92,24 +96,43 @@ public void onLoad(NamespaceBundle bundle) {
...
-                                KafkaTopicManager.removeLookupCache(name.toString());
 
                                 if (log.isDebugEnabled()) {
                                     log.debug("New offset partition load:  {}, broker: {}",
                                         name, service.pulsar().getBrokerServiceUrl());
                                 }
                                 groupCoordinator.handleGroupImmigration(name.getPartitionIndex());
                             }
+
+                            KafkaTopicManager.removeLookupCache(name.toString());
```
BewareMyPower, line 114: Why move it outside the if block? By the way, I found this if condition didn't check the `name`'s namespace, it should be the same with `kafkaMetadataNs/kafkaMetadataNamespace`.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -359,6 +359,9 @@ protected void handleTopicMetadataRequest(KafkaHeaderAndRequest metadataHar,
         CompletableFuture<Map<String, List<TopicName>>> pulsarTopicsFuture = new CompletableFuture<>();
 
         if (topics == null || topics.isEmpty()) {
+            // clean all cache when get all metadata for the first metadata request.
```
BewareMyPower, line 362: The official Kafka clients won't request all metadata during produce and consume. I think the old version librdkafka would request all metadata. If I remember correctly, librdkafka also dosn't request all metadata from 1.0.x. So this behavior may affect other clients.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
BewareMyPower: Ok, I see the lookup cache now records not only `__consumer_offsets` but all topics now.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
dockerzhang: What you say is right,  there is no request for all metadata for librdkafka(1.2.2) and java client(2.5.1)  I tested.

dockerzhang: the latest commit has updated:
1,  keep clean all cache when getting all metadata for librdkafka(<1.0.0)
2,  remove kop address after channel close.

BewareMyPower:
Like I commented on #233 , the unrelated changes should be put to a new PR.

It looks like that the `KOP_ADDRESS_CACHE` is an optimization for broker lookup. Could you tell me if it's related to the fix of #220 ? If not, I think it's better to open a new PR.

dockerzhang:
@BewareMyPower  I will open a new PR for `KOP_ADDRESS_CACHE `  optimization and reupload code for this PR.

BewareMyPower:
> @BewareMyPower I will open a new PR for `KOP_ADDRESS_CACHE ` optimization and reupload code for this PR.

It's merged now. Please rebase and fix the conflicts, thanks.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
```diff
@@ -110,33 +109,32 @@ public void updateCtx() {
...
+            if (topic != null) {
+                consumerManagerFuture.complete(new KafkaTopicConsumerManager(requestHandler, topic.join()));
+                consumerTopicManagers.put(topicName, consumerManagerFuture);
+            } else {
+                consumerManagerFuture.complete(null);
             }
-        );
+            return consumerManagerFuture;
+        }
     }
```
BewareMyPower, line 125: Could it be changed to `computeIfAbsent`? Because `ConcurrentHashMap#computeIfAbsent` is an atomic operation where combination of `containsKey` and `get`/`put` is not.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
```diff
@@ -254,63 +197,28 @@ private void lookupBroker(String topicName,
...
-                    // get topic broker returns null. topic should be removed from LookupCache.
-                    if (ignore == null) {
-                        removeTopicManagerCache(topicName);
-                    }
-
+        if (!topics.containsKey(topicName)) {
+            brokerService.getTopic(topicName, brokerService.isAllowAutoTopicCreation(topicName))
+                    .whenComplete((t2, throwable) -> {
+                if (throwable != null) {
+                    log.debug("getTopic error", throwable);
```
BewareMyPower, line 204: The same with my review before. It's better to use `computeIfAbsent`.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
BewareMyPower: Makes sense. But these changes make `getTopic` and `getTopicConsumerManager` not thread-safe anymore. Should we just use a normal `HashMap` and `synchronized` to make these methods thread safe? Because there're somewhere else that accesses the `topics`.

BewareMyPower: Also I found that after this change, use `CompletableFuture<KafkaTopicConsumerManager>` as a return value is meaningless because `consumerTopicManagers.put(...)` is only invoked in `getTopicConsumerManager` and the future to add is always a completed future.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
```diff
@@ -110,33 +109,32 @@ public void updateCtx() {
...
-                });
+        CompletableFuture<KafkaTopicConsumerManager> consumerManagerFuture = new CompletableFuture<>();
+        if (consumerTopicManagers.containsKey(topicName)) {
+            return consumerTopicManagers.get(topicName);
+        } else {
+            CompletableFuture<PersistentTopic> topic = getTopic(topicName);
+            if (topic != null) {
+                consumerManagerFuture.complete(new KafkaTopicConsumerManager(requestHandler, topic.join()));
+                consumerTopicManagers.put(topicName, consumerManagerFuture);
+            } else {
```
BewareMyPower, line 120: I think `getTopic` never returns null, right? If any exception occurs, it returns a future that is completed with null.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
```diff
@@ -200,50 +173,20 @@ public InternalServerCnx getInternalServerCnx() {
...
-                    }
-                    return null;
-                });
+                    .getBroker(TopicName.get(topicName))
+                    .thenAccept(pair -> {
+                        checkState(pair.getLeft().equals(pair.getRight()));
+                        retFuture.complete(pair.getLeft());
+                    })
+                    .exceptionally(th -> {
+                        return null;
```
BewareMyPower, line 185: `exceptionally` here is meaningless, should you use `retFuture.completeExceptionally(th)` or `retFuture.complete(null)` here? And it's better to add warn logs here. By the way, I think use `whenComplete` is better than `thenAccept` and `exceptionally` here, because we don't care the future returned by `thenAccept`/`exceptionally`.

BewareMyPower (changes requested):

BewareMyPower:
After looking into these code in detail, as far as I comprehend, the main problem is that there could be some null future that have been put into the cache (`topics`, `consumerTopicManagers`) before. So the caller may get a null future from the cache.

But is there any better way to fix it? i.e. the current code from caller's side may be:

```java
topicManager.getTopicConsumerManager(topic).whenComplete((tcm, ex) -> {
    if (tcm == null) {
        // throw NPE
    } else {
```

Could it be changed to code like:

```java
topicManager.getTopicConsumerManager(topic).thenApply(tcm -> {
    if (tcm != null) {
        return tcm;
    } else {
        topicManager.getConsumerTopicManagers().remove(topic);
        // TODO: brokerService.getTopic(topic) and do something with the returned future
```

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
dockerzhang: yes, it returns a future that is completed with null, it would never return null. And I will reconsider `getTopicConsumerManager`.

dockerzhang:
@BewareMyPower I will split this PR into more PRs.

Issue #226, opened 2020-11-11
[BUG] the producer metrics of kop topic is not changed when producing 


dockerzhang:
**Describe the bug**
the producer metrics of kop topic is not changed when producing, still is 0

**To Reproduce**
1. create a partitioned topic
`pulsar-admin persistent create-partitioned-topic  t-kop/ns-kop/lookup2 --partitions 30`
2. producing message using the created topic 
3. show the status of the topic
`bin/pulsar-admin topics partitioned-stats t-kop/ns-kop/lookup2`

**Screenshots**
![image](https://user-images.githubusercontent.com/18047329/98804309-fc12b480-2450-11eb-8eb1-39671d0a911a.png)

PR #228, opened 2020-11-11
use producer to publish message for metrics collected by pulsar

dockerzhang:
to solve https://github.com/streamnative/kop/issues/226, and this patch can improve producing performance about 5 times.

### DIFF
--- a/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestTypeTest.java
+++ b/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestTypeTest.java
@@ -16,6 +16,7 @@ package io.streamnative.pulsar.handlers.kop;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotEquals;
 import static org.testng.Assert.assertNotNull;
 import static org.testng.Assert.assertNull;
 import static org.testng.Assert.assertTrue;
@@ -34,14 +35,11 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import lombok.Cleanup;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.kafka.clients.ApiVersions;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.producer.ProducerRecord;
-import org.apache.kafka.clients.producer.RecordMetadata;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.header.Header;
-import org.apache.kafka.common.record.AbstractRecords;
 import org.apache.kafka.common.serialization.StringSerializer;
 import org.apache.pulsar.client.api.CompressionType;
 import org.apache.pulsar.client.api.Consumer;
@@ -234,11 +232,6 @@ public class KafkaRequestTypeTest extends KopProtocolHandlerTestBase {
         KProducer kProducer = new KProducer(kafkaTopicName, false, getKafkaBrokerPort());
 
         int totalMsgs = 10;
-        int totalBytes = 0;
-        byte maxUsableMagic = new ApiVersions().maxUsableProduceMagic();
-        org.apache.kafka.common.record.CompressionType compression =
-                org.apache.kafka.common.record.CompressionType.NONE;
-        RecordMetadata recordMetadata = null;
 
         String messageStrPrefix = "Message_Kop_KafkaProducePulsarConsume_"  + partitionNumber + "_";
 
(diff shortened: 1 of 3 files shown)


BewareMyPower:
The main issue is that Pulsar consumer cannot consume messages produced from Kafka producer. It looks like the original implementation does some extra work.

The following is just an experiment. Let's change `doPublishMessages` to 

```java
                        // 1. the old way
                        persistentTopic.publishMessage(
                                headerAndPayload,
                                MessagePublishContext.get(
                                        offsetFuture, persistentTopic, System.nanoTime()));
                        // 2. the new way
                        topicManager.registerProducerInPersistentTopic(topic.toString(), persistentTopic);
                        topicManager.getReferenceProducer(topic.toString()).publishMessage(0, 0,
                                headerAndPayload, size, false);
                        offsetFuture.complete(Long.valueOf(size));
```

to produce 2 same messages once.

And change the `testKafkaProducePulsarConsumeMessageOrder` to receive twice:

```
            msg = consumer.receive(1000, TimeUnit.MILLISECONDS);
            assertNotNull(msg);
            msg = consumer.receive(1000, TimeUnit.MILLISECONDS);
            assertNotNull(msg);
```

Then the unit test passed. So maybe we need to figure out what the extra work `persistentTopic.publishMessage` does.

By the way, the NPE of `Producer$MessagePublishContext.completed(Producer.java:377)` seems to be not relative to this issue.

BewareMyPower:
It looks like that the original implementation with a stats update could solve the problem:

```java
                        topicManager.getReferenceProducer(topic.toString())
                                .getTopic().incrementPublishCount(size, headerAndPayload.readableBytes());
                        persistentTopic.publishMessage(
                                headerAndPayload,
                                MessagePublishContext.get(
                                        offsetFuture, persistentTopic, System.nanoTime()));
```

In addition, if the problem was solved, a unit test should be added to verify the stats has been updated:

1. Use kafka producer to produce some messages;
2. Use pulsar admin to get topic stats to verify  `bytesInCounter` and `msgInCounter`.

BewareMyPower:
Also I found the possible reason. It still may be related to the NPE of Producer$MessagePublishContext.completed(Producer.java:377).

When use `service.Producer#publishMessage`, it would finally call the same method `PersistentTopic#publishMessage`:

```java
    public void publishMessage(ByteBuf headersAndPayload, PublishContext publishContext) {
        pendingWriteOps.incrementAndGet(); // ** [1] pendingWriteOps++ here, we need to -- later **/ 
        /* ... */
        switch (status) {
            case NotDup:
                // this is a PersistentTopic instance that has implemented addComplete/addFailed methods
                ledger.asyncAddEntry(headersAndPayload, this, publishContext);
                break;
            /* other cases... */
        }
    }
```

The only difference is the `publishContext`. The original implemenation uses a `io.streamnative.pulsar.handlers.kop.MessagePublishContext` whose `complete` method only update the publish latency and complete the future.

while `Producer#publishMessage` uses `Producer.MessagePublishContext` whose `complete` method is:

```java
        public void completed(Exception exception, long ledgerId, long entryId) {
            if (exception != null) {
                /* ... */
            } else {
                /* ... */
                this.ledgerId = ledgerId;
                this.entryId = entryId;
                // NPE happens here, producer.cnx.ctx() == null
                producer.cnx.ctx().channel().eventLoop().execute(this);
            }
        }
```

And see the `PersistentTopic#addComplete`:

```java
    public void addComplete(Position pos, Object ctx) {
        PublishContext publishContext = (PublishContext) ctx;
        PositionImpl position = (PositionImpl) pos;

        messageDeduplication.recordMessagePersisted(publishContext, position);
        // NPE caused by Producer.MessagePublishContext#complete, so the following
        // pendingWriteOps-- would be skipped.
        publishContext.completed(null, position.getLedgerId(), position.getEntryId());
        // ** [2] pendingWriteOps-- here, it has ++ in publishMessage  **/ 
        decrementPendingWriteOpsAndCheck();
    }
```

I'm not sure if the reference count leak is the reason, but the NPE here is really an issue. See the example stack:

```
java.lang.NullPointerException: null
	at org.apache.pulsar.broker.service.Producer$MessagePublishContext.completed(Producer.java:377) ~[pulsar-broker-2.6.2.jar:2.6.2]
	at org.apache.pulsar.broker.service.persistent.PersistentTopic.addComplete(PersistentTopic.java:370) ~[pulsar-broker-2.6.2.jar:2.6.2]
	at org.apache.bookkeeper.mledger.impl.OpAddEntry.safeRun(OpAddEntry.java:192) ~[managed-ledger-2.6.2.jar:2.6.2]
```

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestTypeTest.java
```diff
@@ -214,6 +215,56 @@ public void testKafkaProducePulsarConsume(int partitionNumber, boolean isBatch)
...
+                    .add(key2 + i, (value2 + i).getBytes(UTF_8));
+
+            if (isBatch) {
+                kProducer.getProducer()
+                        .send(record).get();
+            } else {
+                kProducer.getProducer()
+                        .send(record)
+                        .get();
+            }
```
BewareMyPower, line 255: The if-else is redundant.

BewareMyPower (changes requested):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestTypeTest.java
```diff
@@ -214,6 +215,56 @@ public void testKafkaProducePulsarConsume(int partitionNumber, boolean isBatch)
...
+
+        for (int i = 0; i < totalMsgs; i++) {
+            String messageStr = messageStrPrefix + i;
+            ProducerRecord record = new ProducerRecord<>(
+                    kafkaTopicName,
+                    i,
+                    messageStr);
+            record.headers()
+                    .add(key1 + i, (value1 + i).getBytes(UTF_8))
+                    .add(key2 + i, (value2 + i).getBytes(UTF_8));
```
BewareMyPower, line 246: Could you explain why add headers here? I didn't find anywhere to check the headers.

dockerzhang: I planed to compute the byte size of the record with headers, it can remove now.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestTypeTest.java
dockerzhang: I will fix it

BewareMyPower (approved):

BewareMyPower:
Add an explanation for

> improve producing performance about 5 times.

After discussing with @dockerzhang , the reason is that the original commit used `Producer#publishMessage`, which has a default `PublishContext` that doesn't parse `MessageId` but simply set offset to 0. Though the parsed offset is not used anywhere now, see #230 .

PR #231, opened 2020-11-12
add back press for kop server

yuanboliu:
kop data flow works like this:

kafka client(1) -> kop/broker (2)-> bookkeeper

Right now we can use throttler to slow down (2) data flood, but (1) doesn't have any protection. This pr is used to enable/disable auto read for (1) channel. With this, we can controll the speed of channel in case of potential oom issue.

### DIFF
--- a/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandlerTest.java
+++ b/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandlerTest.java
@@ -115,39 +115,33 @@ public class KafkaRequestHandlerTest extends KopProtocolHandlerTestBase {
     }
 
     @Test
-    public void testAutoReadEnableDisAble() {
+    public void testAutoReadEnableDisable() {
         final ByteBuf buffer = Unpooled.copiedBuffer("Test", CharsetUtil.US_ASCII);
         final EmbeddedChannel channel = new EmbeddedChannel(handler);
         ByteBuf[] buffers = new ByteBuf[1000000];
         for (int i = 0; i < buffers.length; i++) {
             buffers[i] = buffer.duplicate().retain();
         }
-        try {
-            AtomicBoolean isServiceFinished = new AtomicBoolean(false);
-            AtomicBoolean isChannelWritable = new AtomicBoolean(true);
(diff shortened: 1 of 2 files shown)


tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandlerTest.java
```diff
@@ -107,6 +114,42 @@ protected void cleanup() throws Exception {
...
+                        isChannelWritable.set(false);
+                        break;
+                    }
+                    count += 1;
+                    try {
+                        TimeUnit.MILLISECONDS.sleep(50);
+                    } catch (Exception ignored) {}
+
+                }
+                isServiceFinished.set(true);
```
BewareMyPower, line 144: Is `isServiceFinished` meaningless here?

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandlerTest.java
```diff
@@ -107,6 +114,42 @@ protected void cleanup() throws Exception {
         super.internalCleanup();
     }
 
+    @Test
+    public void testAutoReadEnableDisAble() {
```
BewareMyPower, line 118: Just a typo, `DisAble` -> `Disable`

BewareMyPower:
LGTM for the implementation, just left a few comments on the test.

Another suggest for the test is that you need not to catch exceptions in the unit test. If a method that throws a checked exception was called, add it to the method signature like other tests do.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandlerTest.java
yuanboliu: yes. "writeOutbound" is a sync operation which won't finish until all data have been sent. I was thinking that main thread would finished quickly before detecting writable of channel in an independent thread.

yuanboliu:
@BewareMyPower 
update pr for your comments.

BewareMyPower (approved):

BewareMyPower:
Good job! Thanks for your contribution!

PR #234, opened 2020-11-13
Perform authentication before processing other requests

### LINKED ISSUE #208: [FEATURE] Support authentication in a proper way
The KoP authentication is introduced by #47 . However it's not a proper way to implement it. Because if user didn't configure the SASL related properties, the Kafka client would bypass authentication mechanism.

It works well before because KoP disallow the client access any namespace except for the default configured namespace, i.e. `<kafkaTenant>/<kafkaNamespace>`. The flaw is Kafka client could also bypass the authentication of default namespace.

However, if multi-tenancy was supported, all namespaces could be accessed, so Kafka clients could just bypass 

Add an authenticator to `KafkaRequestHandler` like what Kafka does. Before the authenticator completed, only `API_VERSIONS` and `SASL_HANDSHAKE`/`SASL_AUTHENTICATION` requests could be received. The other requests would just fail.

BewareMyPower:
Fixes #208 

This PR ports Kafka's `SaslServerAuthenticator` into KoP with some limits:

- Don't handle authentication corner cases for the Kafka Client 0.9.0.x, it's too troublesome and KoP doesn't support Kafka Client 0.9.0.x now.
-  Don't handle `SaslHandshakeRequest` v0 because this request has no header that it takes a lot of efforts to refactor current code and test.
- Keep the old authentication way, i.e. user pass a `username` field to represent the namespace and a `data` field to represent the token, then Pulsar broker will use JWT authentication to do the actual authentication. See `docs/security.md` for details.
- Add a unit test to cover the case that client doesn't configure authentication.

More works to do:
- Currently it just does a simple check for permissions in the authentication step. The authorization should be performed before each request is processed.
- The authenticator exposes a getter for `AuthenticationState`, the `isExpired` method should be called to check if the token is expired and trigger the reconnection for clients.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
@@ -2,9 +2,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * <p>
- * http://www.apache.org/licenses/LICENSE-2.0
- * <p>
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
@@ -19,10 +19,12 @@ import io.streamnative.pulsar.handlers.kop.SaslAuth;
 import io.streamnative.pulsar.handlers.kop.utils.SaslUtils;
 
 import java.io.IOException;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import javax.naming.AuthenticationException;
 
+import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
 import org.apache.kafka.common.protocol.ApiKeys;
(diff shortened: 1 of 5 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
```diff
@@ -0,0 +1,265 @@
...
+                responseFuture.complete(request.getErrorResponse(e));
+                throw e;
+            }
+            handleHandshakeRequest((SaslHandshakeRequest) request, responseFuture);
+        } else {
+            throw new AuthenticationException("Unexpected Kafka request of type " + apiKey + " during SASL handshake");
+        }
+    }
+
+    private void handleSaslToken(RequestHeader header,
```
jiazhai, line 159: How about change its name to handleAuthenticate?  
token is not related with SASL.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
```diff
@@ -0,0 +1,265 @@
...
+            authState = authenticationProvider.newAuthState(
+                    AuthData.of(saslAuth.getAuthData().getBytes(UTF_8)), null, null);
+        } catch (AuthenticationException e) {
+            responseFuture.complete(new SaslAuthenticateResponse(Errors.SASL_AUTHENTICATION_FAILED, e.getMessage()));
+            return;
+        }
+        authDataSource = authState.getAuthDataSource();
+        authRole = authState.getAuthRole();
+
+        // TODO: we need to let KafkaRequestHandler do the authorization works. Here we just check the permissions
```
jiazhai, line 202: How a bout open a github issue, and linked it here?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
```diff
@@ -0,0 +1,265 @@
...
+        }
+    }
+
+    private void handleHandshakeRequest(SaslHandshakeRequest request,
+                                        CompletableFuture<AbstractResponse> responseFuture)
+            throws AuthenticationException {
+
+        final String mechanism = request.mechanism();
+        if (mechanism != null && !mechanism.equals("PLAIN")) {
+            // TODO: support more mechanisms
```
jiazhai, line 246: seems not very clear what to do here, how about open an issue, and describe the detailed requests?

tests/src/test/java/io/streamnative/pulsar/handlers/kop/SaslPlainTest.java
```diff
@@ -193,4 +197,24 @@ void badNamespaceProvided() throws Exception {
...
+        props.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, StringSerializer.class.getName());
+        props.put(ProducerConfig.MAX_BLOCK_MS_CONFIG, Integer.toString(metadataTimeoutMs));
+
+        KafkaProducer<String, String> producer = new KafkaProducer<>(props);
+        producer.send(new ProducerRecord<>(TOPIC, "", "hello"), (recordMetadata, e) -> {
+            assertNotNull(e);
+            assertTrue(e.getMessage().contains("Failed to update metadata"));
+        });
+
+        Thread.sleep(metadataTimeoutMs + 300);
```
jiazhai, line 217: why it need a sleep here?

jiazhai: Oh, I see, the assert is in the callback.

jiazhai: Do we have a better way to avoid sleep? sleep will cause a flaky test, because the ci env is not stable.

jiazhai (commented):
Thanks for the work @BewareMyPower , left some minor comments, overall lgtm.

BewareMyPower:
Fixes were added according to the comments, PTAL, @jiazhai

jiazhai (approved):

Issue #237, opened 2020-11-17
[FE] performance improvement discussion


yuanboliu:
Discussion about this piece of code

```
Queue<Pair<CompletableFuture<ByteBuf>, CompletableFuture<PartitionResponse>>> topicQueue =
                transQueue.get(topic);
        // loop from first responseFuture.
        while (topicQueue != null && topicQueue.peek() != null
                && topicQueue.peek().getLeft().isDone() && isActive.get()) {
            CompletableFuture<Long> offsetFuture = new CompletableFuture<>();
```

I've found the code above has big performance impact on publishing message. Since we make a queue for partition writing, the next partition will never get messages until the last partition has finished. This may lead to performance degression when the system gets a lot of requests. I'd propose to remove the queue for patition-level writing since we have a queue for request-level. But I'm not sure whether this change would bring side effection. I open this thread and hope get feedback from community, thanks for any comment in advance.

jiazhai:
@yuanboliu Agree on the performance penalty, we could make it parallel.

yuanboliu:
ok, I'll create a pr for it.

PR #150, opened 2020-05-04
Issue 134: produce optimize

ltamber:
optimize for #134  including the following aspects:
1. recordToEntry paralleled
2. publish message to bookie asynchronous
In our environment, produce delay is about one-fifth of the previous

PR #238, opened 2020-11-17
#237 remove partition queue to improve performance

yuanboliu:


BewareMyPower:
I think @jiazhai 's option in #237 is not to remove the queue.

This PR reverts #150 , so I think the performance improvement needs to be verified. Could you provide any performance comparison between before and after this fix?

Though from my perspective, `transQueue` is redundant because `persistentTopic.publishMessage` calls `ManagedLedger#asyncAddEntry` which is already an asynchronous operation.

BewareMyPower:
In addition, I think it would be helpful to have @ltamber 's option.

yuanboliu:
@BewareMyPower 
Thanks for your comments and sorry for late response.
After testing this pr in private envrionment, there is not much evidence showing this pr has huge performance improvements. I'll close this pr, go through our inner repo again and hope to come up with more prs for kop project.

yuanboliu:
oops, seems I cannot close this pr.

BewareMyPower:
Ok, let me close the PR.

Issue #248, opened 2020-12-02
[FEATURE] kop reuse pulsar consumer state and metric


dockerzhang:
**Is your feature request related to a problem? Please describe.**
After https://github.com/streamnative/kop/issues/226 fixed, the topic of kop could get producer state and metric by pulsar command. And it's better to reuse pulsar consumer metric for kop at the same time.

**Describe the solution you'd like**
we can get the topic of kop state/metric using the pulsar command.

All subscription and consumer metrics is invalid now:
![image](https://user-images.githubusercontent.com/18047329/100843458-58e41680-34b5-11eb-9ff6-e31ba984141b.png)

PR #263, opened 2020-12-11
kop using pulsar consumer metrics

dockerzhang:
fix for #248

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -39,9 +39,11 @@ import io.streamnative.pulsar.handlers.kop.utils.CoreUtils;
 import io.streamnative.pulsar.handlers.kop.utils.KopTopic;
 import io.streamnative.pulsar.handlers.kop.utils.MessageIdUtils;
 import io.streamnative.pulsar.handlers.kop.utils.OffsetFinder;
+import io.streamnative.pulsar.handlers.kop.utils.ZooKeeperUtils;
 
 import java.net.InetSocketAddress;
 import java.nio.ByteBuffer;
+import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -176,6 +178,8 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
     private final String advertisedListeners;
     private final int defaultNumPartitions;
     public final int maxReadEntriesNum;
+    public final ConcurrentHashMap<String, String> currentConnectedGroup;
+    public final String groupIdStoredPath;
     @Getter
     private final EntryFormatter entryFormatter;
 
@@ -208,6 +212,8 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
         this.defaultNumPartitions = kafkaConfig.getDefaultNumPartitions();
         this.maxReadEntriesNum = kafkaConfig.getMaxReadEntriesNum();
         this.entryFormatter = EntryFormatterFactory.create(kafkaConfig.getEntryFormat());
+        this.currentConnectedGroup = new ConcurrentHashMap<>();
+        this.groupIdStoredPath = kafkaConfig.getGroupIdZooKeeperPath();
     }
 
     @Override
@@ -236,6 +242,11 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
             writeAndFlushWhenInactiveChannel(ctx.channel());
             ctx.close();
             topicManager.close();
+            String clientHost = ctx.channel().remoteAddress().toString();
+            if (currentConnectedGroup.containsKey(clientHost)){
+                log.info("currentConnectedGroup remove {}", clientHost);
+                currentConnectedGroup.remove(clientHost);
+            }
         }
     }
 
@@ -664,6 +675,13 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
             throw new NotImplementedException("FindCoordinatorRequest not support TRANSACTION type "
                 + request.coordinatorType());
         }
+        // store group name to zk for current client
+        String groupId = request.coordinatorKey();
+        String zkSubPath = ZooKeeperUtils.groupIdPathFormat(findCoordinator.getClientHost(),
+                findCoordinator.getHeader().clientId());
+        byte[] groupIdBytes = groupId.getBytes(Charset.forName("UTF-8"));
+        ZooKeeperUtils.createPath(pulsarService.getZkClient(), groupIdStoredPath,
+                zkSubPath, groupIdBytes);
 
         findBroker(TopicName.get(pulsarTopicName))
                 .whenComplete((node, t) -> {
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -179,8 +179,8 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
     private final int defaultNumPartitions;
     public final int maxReadEntriesNum;
     // store the group name for current connected client.
-    public final ConcurrentHashMap<String, String> currentConnectedGroup;
-    public final String groupIdStoredPath;
+    private final ConcurrentHashMap<String, String> currentConnectedGroup;
+    private final String groupIdStoredPath;
     @Getter
     private final EntryFormatter entryFormatter;
 
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -178,6 +178,7 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
     private final String advertisedListeners;
     private final int defaultNumPartitions;
     public final int maxReadEntriesNum;
+    // store the group name for current connected client.
     public final ConcurrentHashMap<String, String> currentConnectedGroup;
     public final String groupIdStoredPath;
     @Getter
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
@@ -382,7 +382,7 @@ public class KafkaTopicManager {
         // make sure internal consumer existed
         CompletableFuture<Consumer> consumerFuture = new CompletableFuture<>();
         if (!requestHandler.getGroupCoordinator()
-                .getoffsetAcker().getConsumer(groupId, kafkaPartition).isDone()) {
+                .getOffsetAcker().getConsumer(groupId, kafkaPartition).isDone()) {
             log.warn("not get consumer for group {} this time", groupId);
             consumerFuture.complete(null);
             return consumerFuture;
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
@@ -15,6 +15,7 @@ package io.streamnative.pulsar.handlers.kop;
 
 import static com.google.common.base.Preconditions.checkState;
 
+import io.streamnative.pulsar.handlers.kop.utils.KopTopic;
 import java.net.InetSocketAddress;
 import java.util.Map;
 import java.util.Optional;
@@ -26,12 +27,15 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.kafka.common.TopicPartition;
 import org.apache.pulsar.broker.PulsarServerException;
 import org.apache.pulsar.broker.PulsarService;
 import org.apache.pulsar.broker.service.BrokerService;
+import org.apache.pulsar.broker.service.Consumer;
 import org.apache.pulsar.broker.service.Producer;
 import org.apache.pulsar.broker.service.persistent.PersistentTopic;
 import org.apache.pulsar.client.impl.PulsarClientImpl;
+import org.apache.pulsar.common.naming.NamespaceBundle;
 import org.apache.pulsar.common.naming.TopicName;
 
 /**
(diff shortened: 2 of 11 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MessageRecordUtils.java
```diff
@@ -419,6 +431,8 @@ public static MemoryRecords entriesToRecords(List<org.apache.bookkeeper.mledger.
                 payload.release();
                 entry.release();
             });
+            consumer.updateStats(batchSizes, batchIndexesAcks, totalMessages.get(),
```
dockerzhang, line 434: pulsar `consumer` add `updateStats ` for updating metric

dockerzhang (commented):
it's the first version for kop using pulsar consumer metrics.

dockerzhang:
@jiazhai @BewareMyPower PTAL

BewareMyPower:
Please fix the license check first. I'll review this PR soon.

As for now, I just have a question, `org.apache.pulsar.zookeeper`, `org.apache.bookkeeper.util` both have `ZKUtils`. Is it necessary to implement the ZK utils again? And any change of utils should have tests to verify it no matter how minor it is.

dockerzhang:
@BewareMyPower I will try to check if `ZKUtils` has the needed methods in `org.apache.bookkeeper.util`.

dockerzhang:
@BewareMyPower the PR which exposed `updateStats` with param is at https://github.com/apache/pulsar/pull/8951

BewareMyPower:
Got it.

dockerzhang:
@BewareMyPower https://github.com/apache/pulsar/pull/8951 has been merged, how to trigger this PR using the latest method？

BewareMyPower:
@zymap I think we may need to change the pulsar dependency to sn-pulsar. Could you give some help on @dockerzhang 's question?

BewareMyPower:
@dockerzhang I'll release a self-maintained pulsar soon. Then we can use it as KoP's dependency release. Before it, I may release KoP 2.7.0 first to make KoP work with pulsar 2.7.0.

BewareMyPower:
Now the latest pulsar dependency (2.8.0-rc-202012200040) is ready, please rebase to latest master.

BewareMyPower:
Please rebase to latest and some changes are needed. Because now we use `EntryFormatter` to do the encode/decode, see #280 . You need to decouple the `entriesToRecords` API in this PR to two methods that are `EntryFormatter#decode` and an extra method like `updateStats`:

```java
public void updateStats(List<Entry> entries, Consumer consumer)
```

dockerzhang:
ok, I will add it tomorrow.

BewareMyPower:
Could you rebase and solve the conflict?

jiazhai:
ping @dockerzhang for help with the conflict

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -138,7 +140,7 @@
 public class KafkaRequestHandler extends KafkaCommandDecoder {
     public static final long DEFAULT_TIMESTAMP = 0L;
 
-    private final PulsarService pulsarService;
+    public final PulsarService pulsarService;
```
BewareMyPower, line 143: You should change it to `private`, and access it by `getBrokerService()` (generated from `@Getter`) in `KafkaTopicManager`.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
dockerzhang: ok, I will fix it.

BewareMyPower:
Though the transaction implementation is not stable that these two tests are easy to fail:

```
Error:    TransactionTest.readCommittedTest:94->basicProduceAndConsumeTest:141 » ThreadTimeout
Error:    TransactionTest.readUncommittedTest:99->basicProduceAndConsumeTest:143 » ThreadTimeout
```


I think this PR's code still has some problems, but not sure whether it's related to the unstable transaction implementation. @gaoran10 Please take a look when you're free.

BewareMyPower:
The transaction tests have been fixed by #343, could you rebase again and try to fix the tests? It looks like the code still has some problems.

BewareMyPower:
Before fixing your code correctness, you should first pass the static code analysis by following https://app.codacy.com/gh/streamnative/kop/pullRequest?prid=6829145

dockerzhang:
@BewareMyPower the static code analysis has fixed, I am looking at the failed UT now.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupCoordinator.java
```diff
@@ -200,6 +200,10 @@ public void shutdown() {
         log.info("Shutdown group coordinator completely.");
     }
 
+    public OffsetAcker getoffsetAcker() {
+        return offsetAcker;
+    }
```
BewareMyPower, line 205: It's redundant because we already has `@Getter` on field `offsetAcker`.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/OffsetAcker.java
```diff
@@ -158,6 +159,7 @@ public void close() {
         KopTopic kopTopic = new KopTopic(topicPartition.topic());
         return consumerBuilder.clone()
                 .topic(kopTopic.getPartitionName(topicPartition.partition()))
+                .subscriptionType(SubscriptionType.Shared)
```
BewareMyPower, line 162: Could you explain why use shared subscription mode here?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -174,6 +176,8 @@
     private final String advertisedListeners;
     private final int defaultNumPartitions;
     public final int maxReadEntriesNum;
+    public final ConcurrentHashMap<String, String> currentConnectedGroup;
```
BewareMyPower, line 179: It's better to add comments to indicate what does the key and value string mean here.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/OffsetAcker.java
dockerzhang: each broker will create a consumer for the consumer group(subscription).

BewareMyPower: But I think for a specific partition of `__consumer_offsets`, only one broker will access it. Did you encounter the case that `createConsumer` failed because multiple consumers try to read the same partition?

BewareMyPower:
The `GlobalKTableTest` failed because `KafkaStreams` failed to start after this change. But `KStreamAggregationTest` succeed to start the `KafkaStreams`.

I think it may take some time to figure out why it failed. I'll first take a look for a while.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
```diff
@@ -387,4 +396,33 @@ public void deReference(String topicName) {
...
 
+    public CompletableFuture<Consumer> getGroupConsumers(String groupId, TopicPartition kafkaPartition) {
+        // make sure internal consumer existed
+        CompletableFuture<Consumer> consumerFuture = new CompletableFuture<>();
+        if (!requestHandler.getGroupCoordinator()
+                .getOffsetAcker().getConsumer(groupId, kafkaPartition).isDone()) {
+            log.warn("not get consumer for group {} this time", groupId);
+            consumerFuture.complete(null);
+            return consumerFuture;
+        }
```
BewareMyPower, line 407: After debugging for a while, the `GlobalKTableTests` failures may be caused by this code block. If I commented line 402 to 407, the tests would pass.

Could you take a look?

dockerzhang: it will be stuck in `createConsumer`, I think there is not a problem.
image

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/OffsetAcker.java
dockerzhang: yes, it will catch an Exclusive exception if use the `Exclusive` type in a multi-node cluster.

BewareMyPower: I'm not sure if it's the right behavior, because if multiple brokers use the same group id as the subscription name at the same time, they cannot get the whole data of `__consumer_offsets-partition-i`.

BewareMyPower: I think you can open a new PR to reproduce the problem. If in the simplest scenario the case could happen, it may be worthy to fix. Otherwise, it may be that some brokers don't close the consumer in time.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
BewareMyPower: Yeah it's strange here. I'll dig deeper into this problem when I'm free.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
```diff
@@ -348,7 +351,23 @@ private void readMessagesInternal(KafkaHeaderAndRequest fetch,
...
+                            String clientHost = fetch.getClientHost();
+                            String groupName = requestHandler
+                                    .getCurrentConnectedGroup().computeIfAbsent(clientHost, ignored -> {
+                                String zkSubPath = ZooKeeperUtils.groupIdPathFormat(clientHost,
+                                        fetch.getHeader().clientId());
+                                String groupId = ZooKeeperUtils.getData(requestHandler.getPulsarService().getZkClient(),
+                                        requestHandler.getGroupIdStoredPath(), zkSubPath);
+                                log.info("get group name from zk for current connection:{} groupId:{}",
+                                        clientHost, groupId);
+                                return groupId;
```
BewareMyPower, line 364: The `groupId` here may be an empty String. Then `OffsetAcker#createConsumer` will subscribe the partition with an empty subscription name and `subscribeAsync` stuck. This is the reason why `GlobalKTableTest` failed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/ZooKeeperUtils.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public static String getData(ZooKeeper zooKeeper, String zkPath, String subPath) {
+        String data = "";
+        try {
+            String addSubPath = zkPath + subPath;
+            Stat zkStat = zooKeeper.exists(addSubPath, true);
+            if (zkStat != null) {
+                data = new String(zooKeeper.getData(addSubPath, false, zkStat), StandardCharsets.UTF_8);
+            }
```
BewareMyPower, line 56: We need to handle the case when `zkStat == null` and `getData` returns an empty String group id, which is not accepted.

BewareMyPower (changes requested):

BewareMyPower:
I found the reason why `GlobalKTableTest` failed and gave two comments, PTAL. A simple fix could be ignoring the empty group id like

```java
    public CompletableFuture<Consumer> getGroupConsumers(String groupId, TopicPartition kafkaPartition) {
        if (groupId == null || groupId.isEmpty()) {
            return CompletableFuture.completedFuture(null);
        }
```

But I think getting an empty group id is not a proper behavior.

After resolving the problem, you can rebase to latest to include PR #361 .

BewareMyPower:
In addition, should we avoid `public` fields?

```java
    public final ConcurrentHashMap<String, String> currentConnectedGroup;
    public final String groupIdStoredPath;
```

And the internal methods like `getConsumers` should also be `private` or make it visible only for package.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/ZooKeeperUtils.java
dockerzhang: it will return null.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/OffsetAcker.java
dockerzhang: Each broker owns different partitions for a topic, they can get the whole data if use `Share` type. and I think the subscription type is topic level, so it will catch an Exclusive exception if use `Exclusive`.

dockerzhang:
@BewareMyPower  thanks for your help about this pr. the latest commit fixed next problem
1, rebased from the master
2, fix the group null issue to avoid `GlobalKTableTest` failed.
3, avoid public fields for `currentConnectedGroup` and `groupIdStoredPath`. but `getConsumers` will keep `public` because it used in `KafkaTopicManager`

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/OffsetAcker.java
BewareMyPower: Did you mean when `consumer-0` already subscribes `my-topic-partition-0`, if `consumer-1 ` want to subscribe `my-topic-partition-1` with the same subscription name, it will fail? But I just have a test that it works. Consumers can subscribe different partitions with the same subscription name in Exclusive subscription mode.

BewareMyPower:
Please rebase to master again since #368 fixes the broken kop integration tests.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/OffsetAcker.java
dockerzhang: I tested again about this PR on 3 nodes cluster, both `Exclusive` and `Share` works well, I will recover `subscriptionType`.

BewareMyPower (approved):

PR #267, opened 2020-12-16
remove unnecessary topics cache for Topic Manager

### LINKED ISSUE #266: [BUG] redundant topic cache for Topic Manager
**Describe the bug**
`topics` in `KafkaTopicManager` will cache `PersistentTopic` by `brokerService.getTopic`, it's unnecessary because `PersistentTopic` is cached in `brokerService.getTopic`. we should remove it to avoid getting a `null` topic.

dockerzhang:
fixes #266 
`topics` in `KafkaTopicManager` will cache `PersistentTopic` by `brokerService.getTopic`, it's unnecessary because `PersistentTopic` is cached in `brokerService.getTopic`. we should remove it to avoid getting a `null` topic.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
@@ -51,7 +51,7 @@ public class KafkaTopicManager {
     @Getter
     private final ConcurrentHashMap<String, CompletableFuture<KafkaTopicConsumerManager>> consumerTopicManagers;
 
-    // cache for topics: <topicName, persistentTopic>
+    // cache for topics: <topicName, persistentTopic>, for removing producer
     private final ConcurrentHashMap<String, CompletableFuture<PersistentTopic>> topics;
     // cache for references in PersistentTopic: <topicName, producer>
     private final ConcurrentHashMap<String, Producer> references;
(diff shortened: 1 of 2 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
```diff
@@ -358,22 +330,6 @@ public synchronized void close() {
                 manager.get().close();
             }
             consumerTopicManagers.clear();
-
-            for (Map.Entry<String, CompletableFuture<PersistentTopic>> entry : topics.entrySet()) {
-                String topicName = entry.getKey();
-                removeTopicManagerCache(topicName);
```
BewareMyPower, line 364: Should `removeTopicManagerCache` be retained?

jiazhai:
@dockerzhang Thanks for the change. Would you please help add the dedicated test cases to test the new code path? 
So the code could be protected by the test cases.

dockerzhang:
@jiazhai there is existed `KafkaTopicManager` UT for this PR.
@BewareMyPower the newest topic cache only serve to remove the producer after the connection close.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
```diff
@@ -371,8 +347,8 @@ public synchronized void close() {
                     topicFuture.get().removeProducer(references.get(topicName));
```
BewareMyPower, line 347: `topicFuture.get()` needs a null check because the topic future that was put to `topics` may complete with null.

BewareMyPower (changes requested):

BewareMyPower (approved):

BewareMyPower:
LGTM. PTAL again @jiazhai

PR #280, opened 2020-12-21
Add entry formatter to do conversions between MemoryRecords and ByteBuf

BewareMyPower:
Currently, KoP has some significant performance overhead in the conversion between Kafka's records and Pulsar's messages. For pure Kafka users, i.e. no Pulsar client is involved, the conversion is unnecessary.

So this PR provides a config `entry.format` to specify if the entry's format can be accepted by Pulsar clients. And a `EntryFormatter` is used to do the format conversion instead of static methods of `MessageRecordUtils`.

In future, we'll implement a new `EntryFormatter` to avoid this conversion so that the performance can be improved a lot. After that, users can choose whether to support Pulsar clients for KoP.


BewareMyPower:
Since this PR already includes a lot of changes and the existed tests need refactors as well to apply the new config, I'll open a new PR for the `KafkaEntryFormatter`.

codelipenghui (approved):

PR #288, opened 2020-12-24
Add KafkaEntryFormatter

BewareMyPower:
It's a follow PR of [#280](https://github.com/streamnative/kop/pull/280). This PR adds an optional value to `entry.format` config so that user can improve the performance because there's little entry conversion between `MemoryRecords` and entry's `ByteBuf`, while the cost is that Kafka client won't be compatible with Pulsar client.

Because the entry format of `kafka` still adds a `MessageMetadata` header to record the `entry.format`, it will be possible for Pulsar broker/client to decode these messages later.

This PR adds a `KafkaEntryFormatter` class associated to `entry.format=kafka` and apply this config to existed tests that have relation to Kafka client but not Pulsar client. An exception is `SaslPlainTest`, it will fail when multiple test instances are created because `LOOKUP_CACHE` is a static variable, the `findBroker` call of other test instances may be influenced. So we create two separated derived classes for test.

Finally this PR adds a performance test that can be run to see the time cost of different `EntryFormatter`s, a simple run result is:

```
--- fixed records for 1000 times ---
PulsarEntryFormatter encode time: 7287 ms
KafkaEntryFormatter encode time: 79 ms
NoHeaderKafkaEntryFormatter encode time: 0 ms
--- random records for 1000 times ---
PulsarEntryFormatter encode time: 6947 ms
KafkaEntryFormatter encode time: 78 ms
NoHeaderKafkaEntryFormatter encode time: 1 ms
```

The default batch is 2048 messages and each message contains 1024 bytes. We can see the behavior of `KafkaEntryFormatter`  has some overhead caused by adding header and copying the original bytes from `MemoryRecords`. However it's not significant much. But `PulsarEntryFormatter` takes about 7 ms for each batch, which may be not acceptable.

Another performance test that compares `PulsarEntryFormatter` and `KafkaEntryFormatter` is done by running KoP with Pulsar 2.8.0-SNAPSHOT (commit 11b9359) standalone on my LapTop.

```bash
$ ./bin/kafka-producer-perf-test.sh --topic my-topic --producer-props bootstrap.servers=127.0.0.1:9092 \
 --num-records 100000000 --record-size 128 --throughput -1
```

Sampling from 5 lines from the output after the throughput/latency are stable, and calculating the average value. The result is:

| entryFormat | Throughput (MB/sec) | Avg. Latency (ms) | Max Latency (ms) |
| - | - | - | - |
| pulsar | 39.768 | 740.74 | 802.8 |
| kafka | 44.212 | 667.98 | 756.6 |

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatter.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatter.java
@@ -1,0 +1,70 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.streamnative.pulsar.handlers.kop.format;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import io.streamnative.pulsar.handlers.kop.utils.ByteBufUtils;
+import io.streamnative.pulsar.handlers.kop.utils.MessageIdUtils;
+import java.nio.ByteBuffer;
+import java.util.List;
+import org.apache.bookkeeper.mledger.Entry;
+import org.apache.kafka.common.record.CompressionType;
+import org.apache.kafka.common.record.MemoryRecords;
+import org.apache.kafka.common.record.MemoryRecordsBuilder;
+import org.apache.kafka.common.record.Record;
+import org.apache.kafka.common.record.TimestampType;
+import org.apache.pulsar.common.protocol.Commands;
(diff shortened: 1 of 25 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatter.java
```diff
@@ -0,0 +1,70 @@
...
+        for (Entry entry : entries) {
+            size += entry.getLength();
+        }
+        final MemoryRecordsBuilder builder = MemoryRecords.builder(
+                ByteBuffer.allocate(size),
+                magic,
+                CompressionType.NONE,
+                TimestampType.CREATE_TIME,
+                MessageIdUtils.getOffset(entries.get(0).getLedgerId(), entries.get(0).getEntryId()));
+        entries.forEach(entry -> {
```
codelipenghui, line 58: After convert the entries to the MemoryRecords, we should release the entry?

codelipenghui (commented):
The change looks good to me, nice work.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatter.java
BewareMyPower: It looks like a problem. The `entry` needs to be released, I'll add it soon.

codelipenghui (approved):

jiazhai (approved):

PR #293, opened 2020-12-28
Fix broken tests caused by static lookup cache

BewareMyPower:
The tests for different `EntryFormatter` may fail because `LOOKUP_CACHE` is a static variable. If two test instances were executed in cross, tests might fail like https://github.com/streamnative/kop/runs/1617432032.

This PR replace TestNG `@factory` with two different test classes to avoid two KoP instances share the same `LOOKUP_CACHE`.


BewareMyPower:
@jiazhai PTAL. The broken tests seems to be caused by the issue that we've discussed before. But it somehow didn't fail in #288 which may introduce the broken tests. From the commit history it started to fail from #267 , which may expose the problem before.

jiazhai (approved):

Issue #287, opened 2020-12-24
[FEATURE] Broken or flaky tests tracker issue


BewareMyPower:
This issue is tracking the existed broken or flaky tests. If some test

Broken tests:
- [x] DistributedClusterTest#testMutiBrokerUnloadReload: #252 

Flaky tests:
- [x] KafkaApisTest#testBrokerRespectsPartitionsOrderAndSizeLimits  #246 
- [ ] KafkaIntegrationTest
  These tests are hardly to fail on CI environment but easily to fail on local environment. The reason seems to be related to `TestContainers`. When some docker containers started too quick (it usually happened with Golang containers because the native binaries start faster), the waited condition might never meet because it relies on the output that may not be captured by `TestContainers`.
- [x] KafkaMessageOrderTest#testKafkaProduceMessageOrder
- [x] KafkaRequestHandlerTest#testProduceCallback

Issue #241, opened 2020-11-25
[FEATURE] Support running KoP with Confluent Schema Registry


BewareMyPower:
**Is your feature request related to a problem? Please describe.**
Currently [Confluent Schema Registry](https://github.com/confluentinc/schema-registry) cannot run with KoP because KoP doesn't support `CreateTopics` and `DescribeConfigs` requests.  Also, it causes that Kafka's `AdminClient` doesn't work currently.

P.S. since KoP is based on Kafka 2.0.0, the related Confluent Schema Registry version is 5.0.0.

**Describe the solution you'd like**

- [x] Support CreateTopics: #244 
- [x] Support DescribeConfigs: #251 
- [x] Fix Kafka producer's send callback has no offset: #245 
- [x] Fix KafkaAdmin listTopics error: #254 
----

The DeleteTopics request is not related to this issue. So we'll do the task later.

BewareMyPower:
After #244 , the Confluent Schema Registry still failed because it will simply send a record and consume until the offset:

```java
  public void waitUntilKafkaReaderReachesLastOffset(int timeoutMs) throws StoreException {
    long offsetOfLastMessage = getLatestOffset(timeoutMs);
    log.info("Wait to catch up until the offset of the last message at " + offsetOfLastMessage);
    kafkaTopicReader.waitUntilOffset(offsetOfLastMessage, timeoutMs, TimeUnit.MILLISECONDS);
    log.debug("Reached offset at " + offsetOfLastMessage);
  }
```

```java
      // in getLatestOffset()
      Future<RecordMetadata> ack = producer.send(producerRecord);
      RecordMetadata metadata = ack.get(timeoutMs, TimeUnit.MILLISECONDS);
      this.lastWrittenOffset = metadata.offset();
      log.trace("Noop record's offset is " + this.lastWrittenOffset);
      return this.lastWrittenOffset;
```

However, KoP doesn't return the offset field to producer now. It's simply set -1. So we need to fix it first.

BewareMyPower:
Confluent Schema Registry will still fail if `_schemas` already exist. It's caused by:

```java
      Set<String> allTopics = admin.listTopics().names().get(initTimeout, TimeUnit.MILLISECONDS);
      if (allTopics.contains(topic)) {
        verifySchemaTopic(admin);
      } else {
        createSchemaTopic(admin);
      }
```

[...]

PR #296, opened 2020-12-30
Support continuous offset for kop

### LINKED ISSUE #290: [FEATURE] Support continues offset for kop
As described in #175, the design of map between kakfa offset and pulsar <LedgerId, EntryId> has potential risk of using up ledgerId which leads to cluster can't work.

And also cause problem in #218 

Now, pulsar has support continuous index at message level, we can use this feature to refacotr the implement of offset management in KOP.

**Additional context**
After upgrade the latest dependency of pulsar, see #289, I will work on this feature.
And this feature will resolve #175 and #218  once for all.

aloyszhang:
Fix #290 
This PR supports continuous offset for kop.
Since this PR disable the orginal design of mapping between Pulsar `MessageId` and Kafka `offset`. So I just ignore some UnitTest based on the original desing. 
Maybe we can raise up another issue to track how we deal with these ignored tests.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -703,29 +703,15 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
             }
 
             ManagedLedgerImpl managedLedger = (ManagedLedgerImpl) perTopic.getManagedLedger();
+            PositionImpl lac = (PositionImpl) managedLedger.getLastConfirmedEntry();
             if (timestamp == ListOffsetRequest.LATEST_TIMESTAMP) {
                 PositionImpl position = (PositionImpl) managedLedger.getLastConfirmedEntry();
                 if (log.isDebugEnabled()) {
                     log.debug("Get latest position for topic {} time {}. result: {}",
                         perTopic.getName(), timestamp, position);
                 }
-
-                // no entry in ledger, then entry id could be -1
-                long entryId = position.getEntryId();
-
-                if (legacyMode) {
-                    partitionData.complete(new ListOffsetResponse.PartitionData(
-                            Errors.NONE,
-                            Collections.singletonList(MessageIdUtils
-                                    .getOffset(position.getLedgerId(), entryId == -1 ? 0 : entryId))));
-
-                } else {
-                    partitionData.complete(new ListOffsetResponse.PartitionData(
-                            Errors.NONE,
-                            DEFAULT_TIMESTAMP,
-                            MessageIdUtils
-                                    .getOffset(position.getLedgerId(), entryId == -1 ? 0 : entryId)));
-                }
+                long offset = MessageIdUtils.getCurrentOffset(managedLedger);
+                fetchOffsetForTimestampSuccess(partitionData, legacyMode, offset);
 
             } else if (timestamp == ListOffsetRequest.EARLIEST_TIMESTAMP) {
                 PositionImpl position = OffsetFinder.getFirstValidPosition(managedLedger);
@@ -734,17 +720,18 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
                     log.debug("Get earliest position for topic {} time {}. result: {}",
                         perTopic.getName(), timestamp, position);
                 }
-
-                if (legacyMode) {
-                    partitionData.complete(new ListOffsetResponse.PartitionData(
-                            Errors.NONE,
-                            Collections.singletonList(MessageIdUtils.getOffset(position.getLedgerId(),
-                                    position.getEntryId()))));
+                if (position.compareTo(lac) > 0 || MessageIdUtils.getCurrentOffset(managedLedger) < 0) {
+                    long offset = Math.max(0, MessageIdUtils.getCurrentOffset(managedLedger));
+                    fetchOffsetForTimestampSuccess(partitionData, legacyMode, offset);
                 } else {
-                    partitionData.complete(new ListOffsetResponse.PartitionData(
-                            Errors.NONE,
-                            DEFAULT_TIMESTAMP,
-                            MessageIdUtils.getOffset(position.getLedgerId(), position.getEntryId())));
+                    MessageIdUtils.getOffsetOfPosition(managedLedger, position).whenComplete((offset, throwable) -> {
+                        if (throwable != null) {
+                            log.error("[{}] Failed to get offset for position {}", perTopic, position, throwable);
+                            fetchOffsetForTimestampFailed(partitionData, legacyMode);
+                            return;
+                        }
+                        fetchOffsetForTimestampSuccess(partitionData, legacyMode, offset);
+                    });
                 }
 
             } else {
@@ -760,40 +747,33 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
                             if (finalPosition == null) {
                                 log.warn("Unable to find position for topic {} time {}. get NULL position",
                                     perTopic.getName(), timestamp);
-
-                                if (legacyMode) {
-                                    partitionData.complete(new ListOffsetResponse
-                                            .PartitionData(
-                                            Errors.UNKNOWN_SERVER_ERROR,
-                                            Collections.emptyList()));
-                                } else {
-                                    partitionData.complete(new ListOffsetResponse
-                                            .PartitionData(
-                                            Errors.UNKNOWN_SERVER_ERROR,
-                                            ListOffsetResponse.UNKNOWN_TIMESTAMP,
-                                            ListOffsetResponse.UNKNOWN_OFFSET));
-                                }
+                                fetchOffsetForTimestampFailed(partitionData, legacyMode);
                                 return;
                             }
                         } else {
                             finalPosition = (PositionImpl) position;
                         }
 
+
                         if (log.isDebugEnabled()) {
                             log.debug("Find position for topic {} time {}. position: {}",
                                 perTopic.getName(), timestamp, finalPosition);
                         }
-                        if (legacyMode) {
-                            partitionData.complete(new ListOffsetResponse.PartitionData(
-                                    Errors.NONE,
-                                    Collections.singletonList(
-                                            MessageIdUtils.getOffset(
-                                                    finalPosition.getLedgerId(), finalPosition.getEntryId()))));
+
+                        if (finalPosition.compareTo(lac) > 0 || MessageIdUtils.getCurrentOffset(managedLedger) < 0) {
+                            long offset = Math.max(0, MessageIdUtils.getCurrentOffset(managedLedger));
+                            fetchOffsetForTimestampSuccess(partitionData, legacyMode, offset);
                         } else {
-                            partitionData.complete(new ListOffsetResponse.PartitionData(
-                                    Errors.NONE,
-                                    DEFAULT_TIMESTAMP,
-                                    MessageIdUtils.getOffset(finalPosition.getLedgerId(), finalPosition.getEntryId())));
+                            MessageIdUtils.getOffsetOfPosition(managedLedger, finalPosition)
+                                    .whenComplete((offset, throwable) -> {
+                                        if (throwable != null) {
+                                            log.error("[{}] Failed to get offset for position {}",
+                                                    perTopic, finalPosition, throwable);
+                                            fetchOffsetForTimestampFailed(partitionData, legacyMode);
+                                            return;
+                                        }
+                                        fetchOffsetForTimestampSuccess(partitionData, legacyMode, offset);
+                                    });
                         }
                     }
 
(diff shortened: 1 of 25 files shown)


jiazhai:
Thanks @aloyszhang for the great work. looks there were test errors. and would you please also help add a test for the continuous offset?

aloyszhang:
@jiazhai I'll try to add some test cases for this PR.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -779,22 +793,42 @@ public void findEntryComplete(Position position, Object ctx) {
...
+                                partitionData.complete(new ListOffsetResponse.PartitionData(
+                                        Errors.NONE,
+                                        Collections.singletonList(offset)));
+                            } else {
+                                partitionData.complete(new ListOffsetResponse.PartitionData(
+                                        Errors.NONE,
+                                        DEFAULT_TIMESTAMP,
+                                        offset));
+                            }
+                        });
```
BewareMyPower, line 831: Should we add a method to do the `get offset -> fill PartitionData` logic since these two changes of `fetchOffsetForTimestamp` are totally same?

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicConsumerManager.java
```diff
@@ -157,7 +156,7 @@ public void deleteCursorFailed(ManagedLedgerException exception, Object ctx) {
 
     private Pair<ManagedCursor, Long> createCursorIfNotExists(long offset) {
         // This is for read a new entry, first check if offset is from a batched message request.
-        offset = offsetAfterBatchIndex(offset);
+        //offset = offsetAfterBatchIndex(offset);
```
BewareMyPower, line 159: Remove this line

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/PendingProduce.java
```diff
@@ -59,6 +60,7 @@ public PendingProduce(CompletableFuture<PartitionResponse> responseFuture,
             log.error("Failed to compute ByteBuf for partition '{}': {}", partitionName, e);
             return null;
         });
+        this.numberOfMessages = EntryFormatter.parseNumMessages(memoryRecords);
```
BewareMyPower, line 63: It's duplicated because we already have a `numMessages` field.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessagePublishContext.java
```diff
@@ -68,6 +69,19 @@ public static MessagePublishContext get(CompletableFuture<Long> offsetFuture,
         return callback;
     }
```
BewareMyPower, line 70: I think the old `get` method could be removed, right?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatterHeader.java
```diff
@@ -34,6 +34,19 @@
...
+                .setKey("entry.format")
+                .setValue(EntryFormatterFactory.EntryFormat.KAFKA.name().toLowerCase())
+                .build());
+        builder.setProducerName("");
+        builder.setSequenceId(0L);
+        builder.setPublishTime(0L);
+        builder.setNumMessagesInBatch(numMessages);
+        return builder.build();
+    }
+
```
BewareMyPower, line 49: `KafkaEntryFormatterHeader` is a singleton class because we only need to initialize once for the header that is just used to indicate the entry format is kafka.

Is `setNumMessagesInBatch` necessary here? If yes, the other code of `KafkaEntryFormatterHeader` could be removed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MessageIdUtils.java
```diff
@@ -91,4 +106,66 @@ public static final long offsetAfterBatchIndex(long offset) {
...
+    public static long peekBaseOffsetFromEntry(Entry entry) {
+
+        return peekOffsetFromEntry(entry)
+                - Commands.peekMessageMetadata(entry.getDataBuffer(), null, 0)
+                    .getNumMessagesInBatch() + 1;
+    }
+
+    public static long getMockOffset(long ledgerId, long entryId) {
+        return ledgerId + entryId;
+    }
```
BewareMyPower, line 170: Could you explain what's the meaning of `ledgerId + entryId`? I found `MessagePublishContext#complete` use it as the offset which would be sent back to the producer.

BewareMyPower:
Just leave a few comments, PTAL. In addition, I found some tests are ignored, could you explain why?

BewareMyPower:
The integration tests also failed in my local test environment. Only `node-kafka-node` passed but it doesn't test for consumers. I think there's something wrong with the handler for FETCH requests, please also take a look.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
aloyszhang: I'll add a new method for this logic.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicConsumerManager.java
aloyszhang: will remove

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/PendingProduce.java
aloyszhang: Ok, will remove this field and use the origin filed `numMessages `

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessagePublishContext.java
aloyszhang: will remove the legacy code

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatterHeader.java
aloyszhang: Yes，I think `setNumMessagesInBatch` is nessacery here since we need to caculate the start offset in an entry when handling fetch request and now the an entry only has the end offset.

BewareMyPower: OK, got it.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MessageIdUtils.java
aloyszhang: This a mock method to provide a simulate offset after append one entry successfully. It's invoked from two places :
1. from `MessagePublishContext` when append entry to bookie 
2. from `GroupMetadataManager` when  commited offset write to bookie
as far as I can see, the only requirement for this method is this method should generate a monotone increasing result.

aloyszhang:
> Just leave a few comments, PTAL. In addition, I found some tests are ignored, could you explain why?

These ignored tests all has the code of the legacy logic of `MessageIdUtils#getOffset` or `MessageIdUtils#getMessageId`， so in this PR, just skip thest test first.

aloyszhang:
> The integration tests also failed in my local test environment. Only `node-kafka-node` passed but it doesn't test for consumers. I think there's something wrong with the handler for FETCH requests, please also take a look.

I'll check this later.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MessageIdUtils.java
BewareMyPower: You're probably right. I also think a monotonically increasing offset is enough. If existed tests could pass, I think the precise offset could be implemented later.

BewareMyPower:
> > Just leave a few comments, PTAL. In addition, I found some tests are ignored, could you explain why?
> 
> These ignored tests all has the code of the legacy logic of `MessageIdUtils#getOffset` or `MessageIdUtils#getMessageId`， so in this PR, just skip thest test first.

OK, these tests should be fixed in a following PR.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KopProtocolHandlerTestBase.java
```diff
@@ -622,6 +625,12 @@ public void close() {
         }
     }
 
+    public static void addBrokerEntryMetadataInterceptors(ServiceConfiguration configuration) {
+        Set<String> interceptorNames = new HashSet<>();
+        interceptorNames.add("org.apache.pulsar.common.intercept.AppendBrokerTimestampMetadataInterceptor");
+        interceptorNames.add("org.apache.pulsar.common.intercept.AppendIndexMetadataInterceptor");
+        configuration.setBrokerEntryMetadataInterceptors(interceptorNames);
```
BewareMyPower, line 632: Since after this PR KoP will require an extra config `brokerEntryMetadataInterceptors` for broker, it would be better to update the README: https://github.com/streamnative/kop#install-kop-protocol-handler

BewareMyPower (changes requested):

BewareMyPower:
There're 2 test failures:

> Error:    CommitOffsetBacklogTest.testOffsetCommittedBacklogCleared:214->verifyBacklogInTopicStats:127 expected [3000] but found [2900]
> Error:    KafkaApisTest.testBrokerRespectsPartitionsOrderAndSizeLimits:594->checkFetchResponse:374->lambda$checkFetchResponse$2:394 expected [true] but found [false]

The `testOffsetCommittedBacklogCleared` may be caused by something wrong with `handleOffsetCommitRequest`, so it should be fixed.

As for `testBrokerRespectsPartitionsOrderAndSizeLimits`, it's a flaky test as #287 said. We can ignore it first.

BewareMyPower:
The `SchemaRegistryTest` failed in setup. You may have to check why it failed.

Here's my track issue to support Confluent Schema Registry: https://github.com/streamnative/kop/issues/241. Maybe it could give some help.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/OffsetAcker.java
```diff
@@ -74,8 +88,31 @@ public void ackOffsets(String groupId, Map<TopicPartition, OffsetAndMetadata> of
                     log.warn("Error when get consumer for offset ack:", throwable);
                     return;
                 }
-                MessageId messageId = MessageIdUtils.getMessageId(offsetAndMetadata.offset());
-                consumer.acknowledgeCumulativeAsync(messageId);
+                KopTopic kopTopic = new KopTopic(topicPartition.topic());
+                String partitionTopicName = kopTopic.getPartitionName(topicPartition.partition());
+                brokerService.getTopic(partitionTopicName, false).whenComplete((topic, error) -> {
```
BewareMyPower, line 93: Should we add some logs to handle the case that `error` is not null or `topic` is `Optional.empty()`?

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
```diff
@@ -245,9 +245,9 @@
 
     @FieldContext(
             category = CATEGORY_KOP,
-            doc = "Maximum number of entries that are read from cursor once per time"
+            doc = "Maximum number of entries that are read from cursor once per time, default is 1"
     )
-    private int maxReadEntriesNum = 5;
+    private int maxReadEntriesNum = 1;
```
BewareMyPower, line 250: Is the change here necessary?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatterHeader.java
```diff
@@ -23,33 +23,18 @@
 
     private static volatile PulsarApi.MessageMetadata messageMetadata = null;
```
BewareMyPower, line 24: Remove this line

BewareMyPower (changes requested):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/PulsarAuthEnabledKafkaTest.java
```diff
@@ -13,10 +13,23 @@
...
+    @BeforeClass
+    @Override
+    protected void setup() throws Exception {
+        super.setup();
+    }
+    @AfterClass
+    @Override
+    protected void cleanup() throws Exception {
+        super.cleanup();
+    }
```
BewareMyPower, line 32: I think these lines could be removed, as well as `PulsarAuthEnabledPulsarTest`. The fix should be adding `@AfterClass` annotation to `PulsarAuthEnabledTestBase#cleanup`.

BewareMyPower (changes requested):

BewareMyPower:
LGTM on the whole, great jobs! Just left a few comments, PTAL.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
aloyszhang: Yes, this fix the problem of CommitOffsetBacklogTest.testOffsetCommittedBacklogCleared

BewareMyPower (approved):

Issue #303, opened 2021-01-04
[BUG] consumer will be cleaned quickly


dockerzhang:
**Describe the bug**
 consumer group be cleaned quickly

**To Reproduce**
1. produce messages using a new topic.
2. consume this topic using a new group name.
3. stop consumer
4. consumer group will be removed in about 10/15 minutes.

**Expected behavior**
consumer group could retain longer.

**Screenshots**
![image](https://user-images.githubusercontent.com/18047329/103508063-69fcba80-4e9b-11eb-91e8-d998e09efbb9.png)

WJL3333:
i found some infomation about it.
when call GroupMetadataManager.cleanupGroupMetadata
the GroupMetadata.removeExpiredOffsets filter offsets that can expire. but when the OffsetAndMetadata.expireTimestamp = -1. GroupMetadata.removeExpiredOffsets remove all the offset for this group. so the group transition to Dead state and being removed by GroupMetadataManager.cleanupGroupMetadata

WJL3333:
the problem is that when we process OffsetCommitRequest. we just use OffsetCommitRequest.PartitionData.timestamp as OffsetAndMetadata.expireTimestamp which is -1. so the logic of GroupMetadataManager.cleanupGroupMetadata can remove all the offset in the coordinator.https://github.com/apache/kafka/blob/ac7b5d3389fddf46bc53ab656de1fa7e2562efdb/core/src/main/scala/kafka/server/KafkaApis.scala#L550

BewareMyPower:
@WJL3333 LGTM. The KoP's `handleOffsetCommitRequest` doesn't handle well for the preprocess of PartitionData.

PR #304, opened 2021-01-04
change offsetsRetentionMs for cleanupGroupMetadata

dockerzhang:
bug fix for #303

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupMetadataManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupMetadataManager.java
@@ -240,7 +240,7 @@ public class GroupMetadataManager {
         if (enableMetadataExpiration) {
             scheduler.scheduleAtFixedRate(
                 this::cleanupGroupMetadata,
-                offsetConfig.offsetsRetentionCheckIntervalMs(),
+                offsetConfig.offsetsRetentionMs(),
                 offsetConfig.offsetsRetentionCheckIntervalMs(),
                 TimeUnit.MILLISECONDS
             );


BewareMyPower:
Could you add some explanation? Actually the previous code here is wrong, but the original Kafka code is:

```scala
      // only fun, period, unit are configured, the other params of schedule keep the default
      scheduler.schedule(name = "delete-expired-group-metadata",
        fun = cleanupGroupMetadata,
        period = config.offsetsRetentionCheckIntervalMs,
        unit = TimeUnit.MILLISECONDS)
```

```scala
trait Scheduler {
  /* ... */
  // NOTE: the default delay is 0
  def schedule(name: String, fun: ()=>Unit, delay: Long = 0, period: Long = -1, unit: TimeUnit = TimeUnit.MILLISECONDS)
}
```

```scala
  // class KafkaScheduler(/* ... */) extends Scheduler
  def schedule(name: String, fun: () => Unit, delay: Long, period: Long, unit: TimeUnit) {
    /* ... */
    this synchronized {
      ensureRunning()
      val runnable = CoreUtils.runnable {/* ... */}
      if(period >= 0)
        executor.scheduleAtFixedRate(runnable, delay, period, unit)
      else
        executor.schedule(runnable, delay, unit)
    }
  }
```

So the equivalent Java code should be

```java
            scheduler.scheduleAtFixedRate(
                this::cleanupGroupMetadata,
                0L, // initialDelay the time to delay first execution
                offsetConfig.offsetsRetentionCheckIntervalMs(),
                TimeUnit.MILLISECONDS
            );
```

I think setting `offsetsRetentionMs`  as the `initialDelay` is meaningless.

dockerzhang:
@BewareMyPower  you are right, `offsetsRetentionMs` as the `initialDelay ` is meaningless, here should set to 0.

I think the root reason for this issue is caused by `cleanupGroupMetadata`.
In kop, group will be transitioned to `Dead` at `offsetsRetentionCheckIntervalMs` period, and group will be removed, default time is 10 minutes.
``` java
                if (group.is(GroupState.Empty) && !group.hasOffsets()) {
                    log.info("Group {} transitioned to Dead in generation {}",
                        groupId, group.generationId());
                    group.transitionTo(GroupState.Dead);
                }
```

but in kafka, `cleanupGroupMetadata` only remove expired offsets.
``` scala
  private[group] def cleanupGroupMetadata(): Unit = {
    val currentTimestamp = time.milliseconds()
    val numOffsetsRemoved = cleanupGroupMetadata(groupMetadataCache.values, group => {
      group.removeExpiredOffsets(currentTimestamp, config.offsetsRetentionMs)
    })
    offsetExpiredSensor.record(numOffsetsRemoved)
    info(s"Removed $numOffsetsRemoved expired offsets in ${time.milliseconds() - currentTimestamp} milliseconds.")
  }
```

so, do you think we should remove the group dead transition for this issue?

BewareMyPower:
@dockerzhang Sorry I didn't find any difference between Kafka and KoP for `cleanupGroupMetadata`. As far as I see, KoP also remove expired offsets only.

The corresponding Kafka code of your 1st code block:

```scala
  def cleanupGroupMetadata(groups: Iterable[GroupMetadata], selector: GroupMetadata => Map[TopicPartition, OffsetAndMetadata]): Int = {
    var offsetsRemoved = 0

    groups.foreach { group =>
      val groupId = group.groupId
      val (removedOffsets, groupIsDead, generation) = group.inLock {
        val removedOffsets = selector(group)
        if (group.is(Empty) && !group.hasOffsets) {
          info(s"Group $groupId transitioned to Dead in generation ${group.generationId}")
          group.transitionTo(Dead)
        }
        (removedOffsets, group.is(Dead), group.generationId)
      }
```

And the corresponding KoP code of your 2nd code block:

```java
    CompletableFuture<Void> cleanupGroupMetadata() {
        final long startMs = time.milliseconds();
        return cleanGroupMetadata(groupMetadataCache.values().stream(),
            group -> group.removeExpiredOffsets(time.milliseconds())
        ).thenAcceptAsync(offsetsRemoved ->
                log.info("Removed {} expired offsets in {} milliseconds.",
                    offsetsRemoved, time.milliseconds() - startMs)
            , scheduler);
    }
```

dockerzhang:
@BewareMyPower I got it,  here we should check where will remove the group in detail.

jiazhai:
@WJL3333 would you please also take a look at this PR?

WJL3333:
@jiazhai ok, i'll try

dockerzhang:
this will close for fixed by https://github.com/streamnative/kop/pull/311.

PR #306, opened 2021-01-07
bump to 2.7.0.3-rc-202101071443

Renkai:


BewareMyPower (changes requested):
You should merge to branch-2.7.0. And you only updated the project version using `./scripts/set-project-version.sh`, but didn't update the pulsar version which could be done buy `./scripts/set-pulsar-version.sh`.

PR #317, opened 2021-01-13
Add advertised listeners as the listeners exposed to client

### LINKED ISSUE #315: [BUG] listener address should not always be the same as advertised address
**Describe the bug**
Currently KoP's listener address must be the same as the `advertisedAddress` of Pulsar broker, see https://github.com/streamnative/kop/blob/96a832d501a7bfafea8390dff17d35240cb86502/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java#L479-L480

However the advertised address could be used as an external address, while the listener address is an internal address. They're totally different concepts. The current wrong design may force user to bind to the advertised address.

However the worse is, the Pulsar config `advertisedAddress` doesn't contain the port, so the full advertised address uses the listener port, which may not be exposed, see https://github.com/streamnative/kop/blob/96a832d501a7bfafea8390dff17d35240cb86502/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java#L1495-L1505

Which means when Kafka Client send metadata requests, the controller id returned by KoP uses the internal port and there's no way to send external port to client.

**To Reproduce**
Apply the following diff and rerun the tests, all tests will fail at setup phase.

```diff
diff --git a/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KopProtocolHandlerTestBase.java b/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KopProtocolHandlerTestBase.java
index 0ad4113..05850fc 100644
--- a/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KopProtocolHandlerTestBase.java
+++ b/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KopProtocolHandlerTestBase.java
@@ -149,6 +149,7 @@ public abstract class KopProtocolHandlerTestBase {
         kafkaConfig.setDefaultNumberOfNamespaceBundles(1);
         kafkaConfig.setZookeeperServers("localhost:2181");
         kafkaConfig.setConfigurationStoreServers("localhost:3181");
+        kafkaConfig.setAdvertisedAddress("0.0.0.0");
 
         kafkaConfig.setAuthenticationEnabled(false);
         kafkaConfig.setAuthorizationEnabled(false);
```

**Expected behavior**
Kafka's [advertised.listeners](https://kafka.apache.org/documentation/#brokerconfigs_advertised.listeners) should be ported to KoP as the real advertised listeners. The Pulsar broker's `advertisedAddress` should never used in KoP.

BewareMyPower:
Fixes #315 

The original `listeners` config's semantics is wrong that it mixed the `listeners` and `advertised.listeners` semantics of Kafka. So this PR adds a `kafkaAdvertisedListeners` as the listeners exposed to client, and only use `listeners` as the bind address. To avoid conflict with other protocol handlers, mark `listeners` as deprecated and use `kafkaListeners` instead.

For convenience, this PR adds an `EndPoint` class to do the listener url parse work, which can be applied to both `kafkaListeners` and `kafkaAdvertisedListeners`. It also handles `SASL_XXX` protocols which were not handled before. And the related tests are added.

The existed `KafkaApisTest#testBrokerHandleTopicMetadataRequest` could verify the `kafkaAdvertisedListeners` because the tests' `kafkaAdvertisedListeners` is `127.0.0.1:<port>` while `kafkaListeners` is `localhost:<port>`.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -16,7 +16,8 @@ The following table lists all KoP configurations.
 |Name|Description|Default|
 |---|---|---|
 |messagingProtocols|  Messaging protocols available for being loaded by Pulsar Broker |kafka|
-|listeners| Listener properties for the Kafka service (The host should follow the  `advertisedAddress`.) |PLAINTEXT://localhost:9092,SSL://localhost:9093|
+|kafkaListeners|Comma-separated list of URIs we will listen on and the listener names.<br>e.g. PLAINTEXT://localhost:9092,SSL://localhost:9093.<br>If hostname is not set, bind to the default interface.||
+|listeners|Deprecated - use `kafkaListeners` instead. ||
 |kafkaTenant| The default tenant of Kafka's topics |public|
 |kafkaNamespace| The default namespace of Kafka's topics |default|
 |kafkaMetadataTenant| Tenant used for storing Kafka metadata topics |public|
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -17,6 +17,7 @@ The following table lists all KoP configurations.
 |---|---|---|
 |messagingProtocols|  Messaging protocols available for being loaded by Pulsar Broker |kafka|
 |kafkaListeners|Comma-separated list of URIs we will listen on and the listener names.<br>e.g. PLAINTEXT://localhost:9092,SSL://localhost:9093.<br>If hostname is not set, bind to the default interface.||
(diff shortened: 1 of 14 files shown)


BewareMyPower:
In addition, this PR could be cherry picked to branch-2.7.0.

jiazhai (approved):

PR #320, opened 2021-01-14
Fix NPE if kafkaAdvertisedListeners was not set

BewareMyPower:
If `kafkaAdvertisedListeners` that was introduced from #317 was not set, the KoP would fail to start. It's because in `KafkaProtocolHandler#start` we use `kafkaConfig.getKafkaAdvertisedListeners()` to get the configured `kafkaAdvertisedListeners` which is null. But we need to call `KafkaProtocolHandler#getKafkaAdvertisedListeners` to get the `kafkaListeners`/`listeners` if `kafkaAdvertisedListeners` is null.

This PR overwrites the `KafkaServiceConfiguration#getKafkaAdvertisedListeners` to fix this issue.

This issue was not exposed because the `kafkaAdvertisedListeners` has been set in `KopProtocolHandlerTestBase#resetConfig` before. So this PR removes the default config and only set it in `KafkaApisTest` for verification of the `kafkaAdvertisedListeners`.


jiazhai (approved):

BewareMyPower:
The integration tests cannot pass currently, it may be caused by something with docker network. I'll fix it later.

PR #327, opened 2021-01-16
support publishtime for message of kafka entry format

### LINKED ISSUE #326: [BUG] Message of kakfa entry format has invalid publishtime
**Describe the bug**
Currently, messages of kakfa entry format has invalid publishtime of 0.
See https://github.com/streamnative/kop/blob/master/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatter.java#L82

aloyszhang:
fix #326 
support publishtime for message of kakfa entry format

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatter.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatter.java
@@ -19,6 +19,8 @@ import io.streamnative.pulsar.handlers.kop.utils.ByteBufUtils;
 import io.streamnative.pulsar.handlers.kop.utils.MessageIdUtils;
 import java.nio.ByteBuffer;
 import java.util.List;
+import java.util.stream.StreamSupport;
+
 import org.apache.bookkeeper.mledger.Entry;
 import org.apache.kafka.common.record.CompressionType;
 import org.apache.kafka.common.record.MemoryRecords;
@@ -39,7 +41,7 @@ public class KafkaEntryFormatter implements EntryFormatter {
         final ByteBuf recordsWrapper = Unpooled.wrappedBuffer(records.buffer());
         final ByteBuf buf = Commands.serializeMetadataAndPayload(
                 Commands.ChecksumType.None,
-                getMessageMetadataWithNumberMessages(numMessages),
+                getMessageMetadataWithNumberMessages(numMessages, records),
                 recordsWrapper);
         recordsWrapper.release();
         return buf;
@@ -71,7 +73,8 @@ public class KafkaEntryFormatter implements EntryFormatter {
         return builder.build();
     }
 
-    private static PulsarApi.MessageMetadata getMessageMetadataWithNumberMessages(int numMessages) {
+    private static PulsarApi.MessageMetadata getMessageMetadataWithNumberMessages(int numMessages,
+                                                                                  MemoryRecords records) {
         final PulsarApi.MessageMetadata.Builder builder = PulsarApi.MessageMetadata.newBuilder();
         builder.addProperties(PulsarApi.KeyValue.newBuilder()
                 .setKey("entry.format")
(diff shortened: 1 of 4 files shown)


BewareMyPower (changes requested):
I think for `entryFormat=kafka`, setting publish time in Pulsar header is meaningless. Because the `MemoryRecords` contains a publish time field, and Pulsar header is invisible for Kafka client. If you want to set publish time field, I think you'd better change `MemoryRecords`.

In addition, what does a published message without publish time affect?

aloyszhang:
Without setting a publishtime, how could we locate  position for a given timestamp when `handleListOffsetRequest`?

BewareMyPower:
> Without setting a publishtime, how could we locate position for a given timestamp when `handleListOffsetRequest`?

Okay, so could you add a test for list offset by timestamp? I found there's no tests affected by the `KafkaEntryFormatter`'s wrong behavior before.

BewareMyPower:
Please add the license header

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatter.java
```diff
@@ -71,15 +73,25 @@ public MemoryRecords decode(List<Entry> entries, byte magic) {
...
         builder.setProducerName("");
         builder.setSequenceId(0L);
-        builder.setPublishTime(0L);
+        boolean res = StreamSupport.stream(records.records().spliterator(), true).anyMatch(record -> {
+            if (record.timestamp() > 0) {
+                builder.setPublishTime(record.timestamp());
+                return true;
+            }
+            return false;
+        });
```
BewareMyPower, line 91: Even if it's a parallel stream, the `builder.setPublishTime` will be invoked for each record that has timestamp. If a `MemoryRecords` has 10000 records, it will be set for 10000 times but only one random (because it's parallel stream) timestamp will be set.

Why not just set it to `System.currentTimeMillis()`? The publish time should be associated to the whole records, not any single record.

BewareMyPower (changes requested):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/EntryPublishTimeKafkaFormatTest.java
```diff
@@ -0,0 +1,135 @@
...
+                .setTargetTimes(targetTimes);
+
+        KafkaCommandDecoder.KafkaHeaderAndRequest request = buildRequest(builder);
+        CompletableFuture<AbstractResponse> responseFuture = new CompletableFuture<>();
+        kafkaRequestHandler.handleListOffsetRequest(request, responseFuture);
+
+        AbstractResponse response = responseFuture.get();
+        ListOffsetResponse listOffsetResponse = (ListOffsetResponse) response;
+        assertEquals(listOffsetResponse.responseData().get(tp).error, Errors.NONE);
+        assertTrue(listOffsetResponse.responseData().get(tp).offset == 0);
```
BewareMyPower, line 114: Use `assertEquals` here.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/EntryPublishTimeKafkaFormatTest.java
```diff
@@ -0,0 +1,135 @@
...
+        RequestHeader header = RequestHeader.parse(serializedRequest);
+
+        ApiKeys apiKey = header.apiKey();
+        short apiVersion = header.apiVersion();
+        Struct struct = apiKey.parseRequest(apiVersion, serializedRequest);
+        AbstractRequest body = AbstractRequest.parseRequest(apiKey, apiVersion, struct);
+        return new KafkaCommandDecoder.KafkaHeaderAndRequest(header, body, byteBuf, serviceAddress);
+    }
+
+}
```
BewareMyPower, line 135: Can you reuse the `KafkaApisTest`? I found it's the same with `KafkaApisTest#buildRequest`. Also I think the tests can be put into `KafkaApisTest` as well.

BewareMyPower: > Also I think the tests can be put into KafkaApisTest as well.

Oh I just noticed that the `KafkaApiTest` only supports pulsar entry format.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatter.java
aloyszhang: `System.currentTimeMillis()` make sense

BewareMyPower (approved):

Issue #276, opened 2020-12-18
[BUG] KafkaConsumer.position() Method should give the offset one larger than the last record


WalterMa:
**Describe the bug**
According to Kafka Conusmer API Doc:
https://kafka.apache.org/22/javadoc/index.html?org/apache/kafka/clients/consumer/KafkaConsumer.html

> The position of the consumer gives the offset of the next record that will be given out. It will be **one larger than the highest offset the consumer has seen** in that partition. It automatically advances every time the consumer receives messages in a call to poll(Duration).

While currently, KoP will return the offset of the last record, without adding one to it

**To Reproduce**
1. Suscribe to a Kafka topic or a KoP topic
2. Call `conusmer.seekToEnd()` and then print `conusmer.position()` result
3. Kafka will return the offset one larger than last record, while KoP not.

**Expected behavior**
KafkaConsumer.position() Method should give the offset one larger than the last record

BewareMyPower:
Could you give your code example? I accidentally found KoP doesn't support `KafkaConsumer#seekEnd` yet. The error is

> java.lang.IllegalStateException: You can only check the position for partitions assigned to this consumer.

BewareMyPower:
Sorry it's my fault. I've set the wrong partition in my tests before. I'll take a look at this issue.

WalterMa:
> Sorry it's my fault. I've set the wrong partition in my tests before. I'll take a look at this issue.

That's great!
Do you still need some code examples for reproducing?

BewareMyPower:
@WalterMa no, I've reproduced the problem in a unit test. Also I found there're no tests for `KafkaConsumer#seek`. If the bug was fixed, the related tests would be added as well.

BewareMyPower:
I've a question. What does the **wrong** behavior affects?

Currently the KoP's offset is just mapped to Pulsar's **message id**, i.e. the `(ledger id, entry id, batch index)` triple. So it makes offset not continuous.

It's different from Kafka which uses a HW(High Watermark, **the next uncommitted offset**) to fetch messages. KoP just fetches messages from the latest message id, i.e., **the latest committed offset**. It's because KoP or Pulsar doesn't know what the next message's id is. But Kafka knows the next message's offset is just `lastOffset + 1`.

In conclusion, if there's not a big issue, we should change the current behavior after continuous offset is implemented.

WalterMa:
[...]

PR #330, opened 2021-01-16
fix highwatermark and logendoffset error

### LINKED ISSUE #329: [BUG] HighWaterMark and LogEndOffset error
**Describe the bug**
For kafka, HighWaterMark and LogEndOffset should be the max offset + 1

aloyszhang:
fix #329 
fix highwatermark and logendoffset error

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MessageIdUtils.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MessageIdUtils.java
@@ -36,6 +36,14 @@ public class MessageIdUtils {
         return ((ManagedLedgerInterceptorImpl) managedLedger.getManagedLedgerInterceptor()).getIndex();
     }
 
+    public static long getHighWatermark(ManagedLedger managedLedger) {
+        return getCurrentOffset(managedLedger) + 1;
+    }
+
+    public static long getLogEndOffset(ManagedLedger managedLedger) {
+        return getCurrentOffset(managedLedger) + 1;
+    }
+
     public static CompletableFuture<Long> getOffsetOfPosition(ManagedLedgerImpl managedLedger, PositionImpl position) {
         final CompletableFuture<Long> future = new CompletableFuture<>();
         managedLedger.asyncReadEntry(position, new AsyncCallbacks.ReadEntryCallback() {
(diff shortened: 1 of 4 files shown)


BewareMyPower:
Could you add some tests? IMO this PR may fix #276

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MessageIdUtils.java
```diff
@@ -107,6 +107,14 @@ public static final long offsetAfterBatchIndex(long offset) {
         return offset;
     }
 
+    public static long getHighWaterMark(ManagedLedger managedLedger) {
```
jiazhai, line 110: Hi @aloyszhang Seems these 2 methods are the same, we provided a wrapper on getCurrentOffset, but the inside code not changed, Would you please help provide more info on how could this fix the issue?

BewareMyPower:
Please rebase to latest master to resolve the conflicts and these two lines need to fix: https://github.com/streamnative/kop/blob/a59d745d1602373c4279fab73530ee072c92df9e/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaApisTest.java#L320-L321 by just removing the comment and modifying the expected value in `assertEqual`.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MessageIdUtils.java
aloyszhang: `getCurrentOffset`  gives the max offset for the current time, and `getHighWaterMark` and `getLogEndOffset` will give  `getCurrentOffset` + 1.

aloyszhang:
@BewareMyPower  I have rebase the master branch, PTAL.   
And for #276 , I found there is something wrong with the logic of `ListOffset` when doing binary search.
I'll open an issue later and test for `ListOffset` will be added with the new issue.

BewareMyPower:
@aloyszhang As I said before, `testConsumerListOffsetLatest` needs to fix.

BewareMyPower (approved):

aloyszhang:
@BewareMyPower  I noticed your comment about the ListOffset test , but the test is not cover all condition.  You can see #334

PR #338, opened 2021-01-21
fix listOffset returns wrong value

### LINKED ISSUE #334: [BUG] ListOffset return wrong value
**Describe the bug**
Currently, KoP return wrong value when process `ListOffsetRequest`.

**To Reproduce**
1. start pulsar broker with KoP
2. send some message to broker
3. `ListOffsetRequest` for specified timstamp

**Expected behavior**
KoP should return the right value for `ListOffsetRequest`

**Additional context**
For example, we produce 4 message with timestamp
```bash
 timestamp for message 0 is 2
 timestamp for message 1 is 4
 timestamp for message 2 is 6
 timestamp for message 3 is 8
```
and do ListOffsetRequest test ， offset result are following
```bash
offset for earliest 0
 offset for latest 4
 offset for time  2 is 0
 offset for time  3 is 0
 offset for time  4 is 1
 offset for time  5 is 1
 offset for time  6 is 2
 offset for time  7 is 2
 offset for time  8 is 3
 offset for time  9 is 3
```

As we can see from the result,  when timestamp is 3, 5, 7, 9, the offset result should be 1, 2, 3, 4, but found 0, ,1, 2,3

aloyszhang:
fix #334 
This PR fix listOffset returns wrong value

### DIFF
--- a/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaApisTest.java
+++ b/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaApisTest.java
@@ -321,6 +321,123 @@ public class KafkaApisTest extends KopProtocolHandlerTestBase {
         assertEquals(listOffsetResponse.responseData().get(tp).timestamp, Long.valueOf(0));
     }
 
+    @Test(timeOut = 80000)
+    public void testConsumerListOffset() throws Exception {
+        String topicName = "listOffset";
+        TopicPartition tp = new TopicPartition(topicName, 0);
+
+        // create partitioned topic.
+        admin.topics().createPartitionedTopic(topicName, 1);
+
+        // 1. prepare topic:
+        @Cleanup
+        KProducer kProducer = new KProducer(topicName, false, getKafkaBrokerPort());
+        int totalMsgs = 10;
+        String messageStrPrefix = topicName + "_message_";
+
+        // produce 10 message with offset and timestamp :
+        //  message               timestamp        offset
+        // listOffset_message_0   2                0
+        // listOffset_message_1   4                1
+        // listOffset_message_2   6                2
+        // listOffset_message_3   8                3
+        // listOffset_message_4   10               4
+        // listOffset_message_5   12               5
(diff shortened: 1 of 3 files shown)


aloyszhang:
@BewareMyPower can you help to check the test `TransactionTest`. I'm not familiar with transaction.

BewareMyPower:
The transaction test is not stable now, I think if the tests only fail with

```
Error:    TransactionTest.readCommittedTest:94->basicProduceAndConsumeTest:141 » ThreadTimeout
Error:    TransactionTest.readUncommittedTest:99->basicProduceAndConsumeTest:143 » ThreadTimeout**
```

we can treat it as success.

BewareMyPower:
I've opened #334 to track the transaction failures. And I'll review this PR later.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaApisTest.java
```diff
@@ -321,6 +321,123 @@ public void testConsumerListOffsetLatest() throws Exception {
...
+        assertEquals(listOffsetResponse.responseData().get(tp).error, Errors.NONE);
+        assertEquals(listOffsetResponse.responseData().get(tp).offset.intValue(), totalMsgs);
+
+        listOffsetResponse = listOffset(0, tp);
+        System.out.println("offset for timestamp=0 " + listOffsetResponse.responseData().get(tp).offset.intValue());
+        assertEquals(listOffsetResponse.responseData().get(tp).error, Errors.NONE);
+        assertEquals(listOffsetResponse.responseData().get(tp).offset.intValue(), 0);
+
+        listOffsetResponse = listOffset(1, tp);
+        System.out.println("offset for timestamp=0 " + listOffsetResponse.responseData().get(tp).offset.intValue());
```
BewareMyPower, line 385:
```suggestion
        System.out.println("offset for timestamp=1 " + listOffsetResponse.responseData().get(tp).offset.intValue());
```

BewareMyPower (approved):
LGTM, good job!

PR #343, opened 2021-01-21
Fix transaction ci fail problem

congbobo184:
Thread sync problem.
Simple change. In order to keep CI stable.
It can be optimized later.


congbobo184:
> It looks like that the tests still failed, I'll try to rerun for about at most 2 times to check.

I will find the problem again.

BewareMyPower (approved):

PR #346, opened 2021-01-25
Fix offset in send callback may be wrong

### LINKED ISSUE #332: [BUG] Offset in send callback is not continuous when batching is enabled
**Describe the bug**
#296 implemented continuous offset, but in producer's send callback, the offset is not continuous when batching is enabled.

**To Reproduce**

Modify `KafkaRequestHandlerTest#testProduceCallbck` to following code and rerun.

```java
    public void testProduceCallback() throws Exception {
        final String topic = "test-produce-callback";
        final int numMessages = 20;
        final String messagePrefix = "msg-";

        final Properties props = new Properties();
        props.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:" + getKafkaBrokerPort());
        props.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, IntegerSerializer.class);
        props.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, StringSerializer.class);
        props.put(ProducerConfig.BATCH_SIZE_CONFIG, "100"); // avoid all messages being in the same batch

        @Cleanup
        KafkaProducer<Integer, String> producer = new KafkaProducer<>(props);

        Map<Integer, Long> indexToOffset = new ConcurrentHashMap<>();
        final CountDownLatch latch = new CountDownLatch(numMessages);
        for (int i = 0; i < numMessages; i++) {
            final int index = i;
            producer.send(new ProducerRecord<>(topic, i, messagePrefix + i), (recordMetadata, e) -> {
                if (e != null) {
                    log.error("Failed to send {}: {}", index, e);
                    fail("Failed to send " + index + ": " + e.getMessage());
                }
                assertEquals(recordMetadata.topic(), topic);
                assertEquals(recordMetadata.partition(), 0);
                indexToOffset.put(index, recordMetadata.offset());
                latch.countDown();
            }); // asynchronous send
        }
        latch.await();
        indexToOffset.forEach((index, offset) -> {
            log.info("{} => {} (delta = {})", index, offset, offset - index);
        });
```

The output:

```
1 => 10 (delta = 9)
2 => 9 (delta = 7)
3 => 10 (delta = 7)
4 => 9 (delta = 5)
5 => 10 (delta = 5)
6 => 9 (delta = 3)
7 => 10 (delta = 3)
8 => 9 (delta = 1)
9 => 10 (delta = 1)
10 => 19 (delta = 9)
11 => 20 (delta = 9)
12 => 19 (delta = 7)
13 => 20 (delta = 7)
14 => 19 (delta = 5)
15 => 20 (delta = 5)
16 => 19 (delta = 3)
17 => 20 (delta = 3)
18 => 19 (delta = 1)
19 => 20 (delta = 1)
```

We can see the offset is not continuous, even not monodic increasing. And the first offset is not zero.

However, if we use synchronous send, i.e. `producer.send(/* ... */).get()`, the first offset will be 0, and the offset will be continuous.

**Expected behavior**
The offset in producer's send callback is continuous.

BewareMyPower:
Fixes #332 

The offset in send callback was calculated with the index of `AppendIndexMetadataInterceptor` when `MessagePublishContext#completed` was invoked. However, the index of `AppendIndexMetadataInterceptor` was updated every time a message was sent while `PublishContext#completed` was called every time a message's send was done. So the interceptor's index may be overwritten by a newer message's offset.

This PR fixes this issue by implementing the new method `PublishContext#setMetadataFromEntryData` that was introduced in https://github.com/apache/pulsar/pull/9257.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessagePublishContext.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessagePublishContext.java
@@ -59,7 +59,9 @@ public final class MessagePublishContext implements PublishContext {
             topic.recordAddLatency(System.nanoTime() - startTimeNs, TimeUnit.MICROSECONDS);
 
             if (offset < 0) {
-                log.error("offset is {} (< 0) but no exception was thrown, use the offset from managed ledger", offset);
+                if (offset != -1L) {
+                    log.error("offset was changed to {} (< 0) but no exception was thrown", offset);
+                }
                 offset = MessageIdUtils.getCurrentOffset(managedLedger);
             }
 
@@ -72,6 +74,10 @@ public final class MessagePublishContext implements PublishContext {
 
     @Override
     public void setMetadataFromEntryData(ByteBuf entryData) {
+        if (entryData == null) {
+            // When the managed ledger was closed, the `entryData` is null.
+            return;
+        }
         final BrokerEntryMetadata metadata = Commands.peekBrokerEntryMetadataIfExist(entryData);
         if (metadata != null && metadata.hasIndex()) {
             offset = metadata.getIndex();
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessagePublishContext.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessagePublishContext.java
@@ -13,6 +13,7 @@
  */
 package io.streamnative.pulsar.handlers.kop;
 
+import io.netty.buffer.ByteBuf;
 import io.netty.util.Recycler;
 import io.netty.util.Recycler.Handle;
 import io.streamnative.pulsar.handlers.kop.utils.MessageIdUtils;
@@ -22,6 +23,8 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.bookkeeper.mledger.ManagedLedger;
 import org.apache.pulsar.broker.service.Topic;
 import org.apache.pulsar.broker.service.Topic.PublishContext;
+import org.apache.pulsar.common.api.proto.BrokerEntryMetadata;
+import org.apache.pulsar.common.protocol.Commands;
 
 /**
  * Implementation for PublishContext.
@@ -34,6 +37,7 @@ public final class MessagePublishContext implements PublishContext {
     private long startTimeNs;
     private long numberOfMessages;
     private ManagedLedger managedLedger;
+    private long offset = -1L;
 
     /**
      * Executed from managed ledger thread when the message is persisted.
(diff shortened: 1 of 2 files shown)


BewareMyPower:
All changes are done, PTAL @jiazhai

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessagePublishContext.java
```diff
@@ -54,16 +58,32 @@ public void completed(Exception exception, long ledgerId, long entryId) {
...
-            // In asynchronously send mode, there's a possibility that some messages' offsets are larger than the actual
-            // offsets. For example, two messages are sent concurrently, if the 1st message is completed while the 2nd
-            // message is already persisted, `MessageIdUtils.getCurrentOffset` will return the 2nd message's offset.
-            final long baseOffset = MessageIdUtils.getCurrentOffset(managedLedger) - (numberOfMessages - 1);
+            if (offset < 0) {
+                if (offset != -1L) {
+                    log.error("offset was changed to {} (< 0) but no exception was thrown", offset);
+                }
+                offset = MessageIdUtils.getCurrentOffset(managedLedger);
+            }
```
jiazhai, line 66: what is the value for offset in the else path?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessagePublishContext.java
```diff
@@ -54,16 +58,32 @@ public void completed(Exception exception, long ledgerId, long entryId) {
...
+    @Override
+    public void setMetadataFromEntryData(ByteBuf entryData) {
+        if (entryData == null) {
+            // When the managed ledger was closed, the `entryData` is null.
+            return;
+        }
+        final BrokerEntryMetadata metadata = Commands.peekBrokerEntryMetadataIfExist(entryData);
+        if (metadata != null && metadata.hasIndex()) {
+            offset = metadata.getIndex();
+        }
```
jiazhai, line 84: Should we add some log, or return some error for the else path?

jiazhai (approved):
overall lgtm. left a minor comments

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessagePublishContext.java
BewareMyPower: I added the logs in `completed` to handle the case that `BrokerEntryData` is null or has no index for the specific entry, PTAL again.

BewareMyPower:
It looks like `KafkaRequestHandlerTest.testCreateInvalidTopics` is very unstable in CI environment, please review #361 first, @jiazhai

PR #354, opened 2021-01-26
fix can not consume all message if topics has > 1000 partitions

### LINKED ISSUE #353: [BUG] can not consume all messages if the topic has lots of partitions(>1000)
**Describe the bug**
can not consume all messages if the topic has lots of partitions(>1000).

this issue may be related to #318

**To Reproduce**
1. create a topic with 1000 partitions.
2. produce messages.
3. consume messages using a group.
4. describe consumer group using Kafka client, See an error

**Expected behavior**
LAG is 0 for all partitions.

**Screenshots**
image

image

dockerzhang:
fixes #353

1, `responseData` is `LinkedHashMap`, which implementation is not synchronized, therefore `responseValues` should not parallel.
2, `getExecutor schedule` may not process so fast for lots of partitions.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
@@ -303,7 +303,6 @@ public final class MessageFetchContext {
                             fetch.getHeader(), entriesRead.get(), allSize);
                     }
 
-                    AtomicBoolean allPartitionsNoEntry = new AtomicBoolean(true);
                     responseValues.entrySet().forEach(responseEntries -> {
                         final PartitionData partitionData;
                         TopicPartition kafkaPartition = responseEntries.getKey();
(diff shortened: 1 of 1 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
```diff
@@ -305,7 +304,7 @@ private void readMessagesInternal(KafkaHeaderAndRequest fetch,
                     }
 
                     AtomicBoolean allPartitionsNoEntry = new AtomicBoolean(true);
```
BewareMyPower, line 306:
```suggestion
```

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
```diff
@@ -29,7 +29,6 @@
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
```
BewareMyPower, line 32:
```suggestion
```

BewareMyPower (changes requested):

BewareMyPower (approved):

PR #356, opened 2021-01-26
revert 9a4be21 for close configuring channel auto-read

### LINKED ISSUE #355: [BUG] the logic of configuring channel auto-read is redundant for lots of partitions
**Describe the bug**
the logic of configuring channel auto-read is redundant for lots of partitions, we should revert https://github.com/streamnative/kop/commit/9a4be2112dfd0cb7c0dbbc3c506ca083506310bd

**To Reproduce**
1. create a 1000 partitions topic
2. produce and consume messages.
3. auto-read will be disabled according to the log.

**Expected behavior**
auto-read should not be disabled

**Screenshots**
image

dockerzhang:
revert https://github.com/streamnative/kop/commit/9a4be2112dfd0cb7c0dbbc3c506ca083506310bd, fixes #355


BewareMyPower:
Confirmed with @dockerzhang and @yuanboliu offline, the OOM issue is not related to the issue that #231 want to resolve. So  merge this PR first.

BewareMyPower (approved):

PR #358, opened 2021-01-26
optimize lookupBroker to avoid get null cache

### LINKED ISSUE #357: [BUG] get a lot of TimeoutException when produce/consume for a topic with lots of partitions
**Describe the bug**
when produce/consume for a topic with lots of partitions, it may get a lot of TimeoutException.

**To Reproduce**
1. create a topic with 1000 or 10000 partitions.
2. produce/consume messages.
3. get a lot of TimeoutException.

**Screenshots**
image

dockerzhang:
fixes #357

1,  remove `retFuture.complete(null)` to avoid get null cache.
2,  remove `backoff schedule `  because it will look up next time if can not get the `persistentTopic`.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
@@ -31,7 +31,6 @@ import org.apache.pulsar.broker.PulsarService;
 import org.apache.pulsar.broker.service.BrokerService;
 import org.apache.pulsar.broker.service.Producer;
 import org.apache.pulsar.broker.service.persistent.PersistentTopic;
-import org.apache.pulsar.client.impl.Backoff;
 import org.apache.pulsar.client.impl.PulsarClientImpl;
 import org.apache.pulsar.common.naming.TopicName;
 
@@ -184,12 +183,7 @@ public class KafkaTopicManager {
                     requestHandler.ctx.channel(), topicName);
             }
             CompletableFuture<InetSocketAddress> returnFuture = new CompletableFuture<>();
-            Backoff backoff = new Backoff(
-                100, TimeUnit.MILLISECONDS,
-                30, TimeUnit.SECONDS,
-                30, TimeUnit.SECONDS
-                );
-            lookupBroker(topicName, backoff, returnFuture);
+            lookupBroker(topicName, returnFuture);
             return returnFuture;
         });
     }
@@ -201,7 +195,6 @@ public class KafkaTopicManager {
     // this method do the real lookup into Pulsar broker.
     // retFuture will be completed with null when meet error.
     private void lookupBroker(String topicName,
-                              Backoff backoff,
                               CompletableFuture<InetSocketAddress> retFuture) {
         try {
             // If channel is closing, complete a null to avoid brings in further handling.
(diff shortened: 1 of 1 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
```diff
@@ -226,26 +218,14 @@ private void lookupBroker(String topicName,
...
-                        log.warn("[{}] getBroker for topic {} failed, retried too many times {}, return null."
-                                        + " throwable: ", requestHandler.ctx.channel(), topicName, waitTimeMs, th);
-                        retFuture.complete(null);
-                    } else {
-                        log.warn("[{}] getBroker for topic failed, will retry in {} ms. throwable: ",
-                            topicName, waitTimeMs, th);
-                        requestHandler.getPulsarService().getExecutor()
-                            .schedule(() -> lookupBroker(topicName, backoff, retFuture),
-                                waitTimeMs,
-                                TimeUnit.MILLISECONDS);
```
BewareMyPower, line 241: How can you ensure that it will look up next time if these code was removed? IMO, If lookup failed, only a warn log would be printed and the `retFuture` would never be completed.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
dockerzhang: 1, if `retFuture` can not get the topic address, the client will get an error and retry.
2,  I tested that if `retFuture`  can not be completed, `whenComplete` will not execute. 
if `retFuture`  be completed with `null`,  `LOOK_CACHE` will cache a `retFuture`  with `null` value. here we can add `retFuture.complete(null);` back, because `LOOK_CACHE` remove this topic cache if the address is `null`.

dockerzhang: the latest commit only removes the schedule thread for lookup.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
```diff
@@ -226,26 +219,14 @@ private void lookupBroker(String topicName,
...
-                    }
+                    log.warn("[{}] getBroker for topic failed. throwable: ",
+                            topicName, th);
+                    retFuture.complete(null);
                     return null;
                 });
         } catch (PulsarServerException e) {
             log.error("[{}] getTopicBroker for topic {} failed get pulsar client, return null. throwable: ",
                 requestHandler.ctx.channel(), topicName, e);
-            retFuture.complete(null);
```
BewareMyPower, line 248: So if the code reached here, the `retFuture` would be a never completed `CompletableFuture`? I'm afraid it's not a good behavior.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
dockerzhang: fixed

BewareMyPower (approved):

PR #359, opened 2021-01-26
change default offsets topic number partitions to 8

hangc0276:
### Motivation
For `DefaultOffsetsTopicNumPartitions` in code is set to `1`, however the document setting is `8` . `DefaultOffsetsTopicNumPartitions` should be greater than `1` to improve the throughput.

### Modification
1. Sync with documentation and set to `8`


BewareMyPower:
It looks like that there's something wrong with the test. If we increased num partitions of `__consumer_offsets` to 8, `GroupMetadataManagerTest#testGroupNotExits` would fail.

I think you can add `@Ignore` to this test and then open an issue to track it.

hangc0276:
> It looks like that there's something wrong with the test. If we increased num partitions of `__consumer_offsets` to 8, `GroupMetadataManagerTest#testGroupNotExits` would fail.
> 
> I think you can add `@Ignore` to this test and then open an issue to track it.

OK

BewareMyPower (approved):

PR #371, opened 2021-02-07
Clean log and return highWaterMark if active transactions not exist

gaoran10:
1. remove log if the active transaction map is empty.
2. return the `highWaterMark` if there is no active transaction.


BewareMyPower:
@hangc0276 could you take a look?

BewareMyPower (approved):

PR #384, opened 2021-02-25
there is no need to find position if last confirmed entry is equal to first valid entry

yuanboliu:
in some  case, it may lead to NPE issue.


BewareMyPower:
Could you give an example of NPE caused by the original behavior? Or just explain why it could cause NPE? As far as I see, before this change, when `position.compareTo(lac) == 0`, it went to the else branch

```java
MessageIdUtils.getOffsetOfPosition(managedLedger, position, false, timestamp)
```

jiazhai:
> Could you give an example of NPE caused by the original behavior? Or just explain why it could cause NPE? As far as I see, before this change, when `position.compareTo(lac) == 0`, it went to the else branch
> 
> ```java
> MessageIdUtils.getOffsetOfPosition(managedLedger, position, false, timestamp)
> ```

@yuanboliu to help on this question.

yuanboliu:
it seems not to be reproduced again, close it.

PR #386, opened 2021-02-26
Bump pulsar to 2.8.0-rc-202102252222

BewareMyPower:
This version update is convenient for tests in real environment since there's no binary download url for original pulsar `2.8.0-rc-202101252233`.

This PR fixes the API incompatibility problems that are introduced by https://github.com/apache/pulsar/pull/9397 and https://github.com/apache/pulsar/pull/9302.

Another significant change between these two versions is https://github.com/apache/pulsar/pull/9338, which introduced metadata-store API for cluster resources. This PR fixed the test failure caused by it as well. Since KoP `tests` module only uses one `MockZooKeeper` to manage z-nodes, see `KopProtocolHandlerTestBase#createMockZooKeeper`, the mocked `createConfigurationMetadataStore` method returns `mockedZooKeeper` here instead of a `mockedZooKeeperGlobal` like what Pulsar did in `MockedPulsarServiceBaseTest`.

Besides, there's a test bug in `testBrokerHandleTopicMetadataRequest` that was not exposed by the previous Pulsar. This PR fixes it.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaService.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaService.java
@@ -116,8 +116,9 @@ public class KafkaService extends PulsarService {
 
             BookKeeperClientFactory bkClientFactory = newBookKeeperClientFactory();
             setBkClientFactory(bkClientFactory);
-            setManagedLedgerClientFactory(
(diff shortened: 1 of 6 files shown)


BewareMyPower:
Some deleted classes should be recovered first.

BewareMyPower:
It looks like Pulsar introduced some changes that make integration tests fail. I'll figure out the reason.

jiazhai (approved):

PR #387, opened 2021-02-26
Remove outdated files for standalone KoP service

BewareMyPower:
There're some code and configs from a long time ago. They're unused now and may cause confusion. For example, we should document KoP config items in `docs/configuration.md` instead of `kafka-impl/conf/kop.conf`.


BewareMyPower:
@jiazhai PTAL

jiazhai (approved):

PR #390, opened 2021-03-03
Avoid unexpected ConflictException when metadata is created

BewareMyPower:
Currently KoP created `__consumer_offsets` repeatedly, the `ConflictException` will be thrown after https://github.com/apache/pulsar/pull/9342. Though it doesn't affect the usage because the `ConflictException` is treated as the concurrent issue and will be swallowed, it could cause some unnecessary logs. So this PR remove the redundant partition creation code.

Another issue is that when `__consumer_offsets` already exists, the existed partition will be created because the partition returned by broker contains `persistent://` prefix but the topics in `offsetPartitionSet` don't. This PR use `KopTopic` to avoid the topic name dismatch.


dockerzhang (commented):
LGTM

jiazhai (approved):

PR #391, opened 2021-03-03
change pulsar dependency jars to provided scope to reduce nar package size

hangc0276:
### Motivation
KOP put pulsar dependency jar into nar package, which lead the nar size up to 130MB. However, the kop nar run with pulsar broker, the pulsar dependency is no need to be packed into KOP nar package.

### Changes
1. change pulsar dependency jar scope to provided in pom.xml


dockerzhang (commented):
LGTM

BewareMyPower (approved):

PR #394, opened 2021-03-08
[ENHANCEMENT] perf tuning in decode phrase

wuzhanpeng:
### Motivation
The decode method in `KafkaEntryFormatter` is time-consuming among which the process of constructing `MemoryRecordsBuilder` takes the highest proportion of time. 

### Designing
Therefore our aims are:
1. overwrites `BaseOffset` field in byte buffer directly to avoid time-consuming iteration which is mainly used for assigning offsets
2. concatenates multiple record batches into one `MemoryRecords` object to better simulate kafka's behavior

### Modifications
- remove the iteration of continuously appending records in the `MemoryRecordsBuilder`
- change to one-to-one correspondence between batch and entry

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatter.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatter.java
@@ -13,21 +13,20 @@
  */
 package io.streamnative.pulsar.handlers.kop.format;
 
+import static org.apache.kafka.common.record.Records.MAGIC_OFFSET;
+import static org.apache.kafka.common.record.Records.OFFSET_OFFSET;
+
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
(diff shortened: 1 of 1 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatter.java
```diff
@@ -17,18 +17,18 @@
...
 import org.apache.pulsar.common.api.proto.MessageMetadata;
 import org.apache.pulsar.common.protocol.Commands;
 
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.apache.kafka.common.record.Records.MAGIC_OFFSET;
+import static org.apache.kafka.common.record.Records.OFFSET_OFFSET;
+
```
BewareMyPower, line 31: Please change the imports order to make `check-style` plugin succeed.

BewareMyPower (changes requested):

jiazhai (approved):

BewareMyPower (approved):

PR #403, opened 2021-03-15
add produce message validation

hangc0276:
### Motivation
KOP receives Kafka messages produced by all kinds of Kafka Producer, and packed into pulsar entry and stored in bookkeeper without validation. If Kafka producer produced a broker message to KOP, the consumer will parse failed and may cause consumer failed.

### Changes
1. Add message validation in KOP `handleProduce` stage. If it validate failed, return error to Kafka Producer and request resend to ensure valid message stored in bookkeeper.


dockerzhang (commented):
LGTM

BewareMyPower (approved):

PR #400, opened 2021-03-11
Support OAUTHBEARER mechanism for KoP

BewareMyPower:
This PR introduces OAUTHBEARER mechanism to KoP.

Two config items are added. One is the server callback handler to validate OAUTHBEARER token, the other is the properties configuration file that contains the server callback handler's configs. The default server callback handler is the same as Kafka, which validates unsecured JSON Web Tokens.

In addition, the authenticator is refactored that now it uses different `SaslServer` for different mechanisms to perform authentication.

Related tests are added to verify these configs and the default server callback handler:
- `KafkaServerConfiguration#testGetKopOauth2Configs`: unit test for `getKopOauth2Properties` method.
- `SaslOAuthBearerTest`: test `OAUTHBEARER` SASL mechanism with Kafka's default server callback handler.
- `CustomOAuthBearerCallbackHandlerTest`: test custom `AuthenticateCallbackHandler`.

PR #405, opened 2021-03-18
Add KoP's own callback handlers for OAuth 2.0 authentication

BewareMyPower:
Based on the framework from https://github.com/streamnative/kop/pull/400, this PR adds two callback handlers and their associated configs:
- `OauthLoginCallbackHandler` and `ClientConfig`: They are in an independent module and for Kafka client to get access token from a third-party OAuth 2.0 authorization server. It requires issuer URI, credential file's URI and andience, just like what Pulsar client does.
- `OauthValidatorCallbackHandler` and `ServerConfig`: It validates the client's access token using Pulsar's `AuthenticationProvider` whose `authMethod` is determined by a config of `ServerConfig`.

Since the validate callback handler is created by `SaslServer`, we cannot pass construct params to it. So this PR makes `AuthenticationService` static so that the callback handler could access it.

Also this PR exposed the role (or authorization id) for the implementation of authorization in future.

Unit tests are added for `ServerConfig` and `ClientConfig` and an integration test is added for OAuth 2.0 authentication with the new added callback handlers in this PR.


BewareMyPower:
There're some issues left that will be processed in new PRs.
- The docs for enabling OAuth 2.0 authentication.
- Remove unnecessary dependencies of login callback handler.
- The `iat` and `exp` claims are not handled in callback handlers since Pulsar's `Authentication` interface doesn't expose related methods. We need to assume token is JWT and parse it again.

jiazhai (approved):

PR #424, opened 2021-04-04
[Transaction] Add transaction log key and transaction log value

gaoran10:
# Motivation

Transaction log key and value are the preconditions of the transaction log.

The transaction log key will be encoded to bytes and set in message key bytes.
The transaction log value will be encoded to bytes and set in message value bytes.


BewareMyPower:
Could you explain where will the transaction log key/value be used?

gaoran10:
> Could you explain where will the transaction log key/value be used?

The transaction log key is used to store the transactional id, the transaction log value is used to store the transaction metadata, such as state, topic partitions, start time, etc. If the broker stopped or crashed, we could recover the transaction metadata from the transaction log.

I'll add the store&recover logic in another PR.

BewareMyPower (approved):

Issue #229, opened 2020-11-12
[BUG] the topic can not remove after the producer and consumer finished


dockerzhang:
**Describe the bug**
the topic can not remove after the producer and consumer finished.

**To Reproduce**
1, create a partitioned topic
`pulsar-admin persistent create-partitioned-topic t-kop/ns-kop/lookup2 --partitions 30`
2, produce `100` message using the created topic
3, consume all produced messages.
4, delete the used topic `lookup2` with pulsar command
`./bin/pulsar-admin topics delete-partitioned-topic  t-kop/ns-kop/lookup2`
5, See error:
`Failed to perform http delete request: javax.ws.rs.ClientErrorException: HTTP 412 Precondition Failed
Topic has active producers/subscriptions`

**Expected behavior**
the topic can remove after the producer and consumer finished.

**Screenshots**
![image](https://user-images.githubusercontent.com/18047329/98894624-22336580-24e0-11eb-95fd-32fe5d213768.png)

BewareMyPower:
Yeah, it's an existed issue for a long time because the `OffsetAcker` holds consumers for acknowledgement. I think there's a better way to do it.

jiazhai:
added comments in the PR

PR #425, opened 2021-04-05
Close offset consumers when channel is inactive

BewareMyPower:
Fixes https://github.com/streamnative/kop/issues/229

### Motivation
The consumers of `OffsetAcker` are used to commit offsets only. However, they would never be closed. Therefore the topics that are consumed by Kafka clients cannot be deleted because topic has active producers/subscriptions.

### Modifications
- Record all group IDs before group coordinator handles `SYNC_GROUP` request, which will call `OffsetAcker#addOffsetsTracker` to add offset consumers if success.
- Close and remove the consumers associated with the recorded group IDs before during the close phase of channel. It's okay because `OffsetAcker#getConsumer` (called by `ackOffsets`) will recreate the consumer if it doesn't exist.
- Add tests to delete a topic after all Kafka clients are closed.


BewareMyPower:
Mark it as WIP since some tests failed with unknown reason, and integration tests all failed.

jiazhai (approved):

PR #426, opened 2021-04-05
Avoid creating offset consumer before updating consumer metrics

BewareMyPower:
### Motivation
Sometimes multiple brokers could subscribe the same topic partition and fail with `ConsumerBusyException`:

```
WARN  io.streamnative.pulsar.handlers.kop.coordinator.group.OffsetAcker - Error when get consumer for offset ack:
  java.util.concurrent.CompletionException: org.apache.pulsar.client.api.PulsarClientException$ConsumerBusyException: Exclusive consumer is already connected
```

This problem was mentioned in https://github.com/streamnative/kop/pull/263 before. However, switching the subscription type to shared or failover is not correct because a partition should have only one consumer at the same time. The problem is before updating consumer metrics, it uses `OffsetAcker#getConsumer` to check if the consumer has been created. This check is wrong.

Firstly, `OffsetAcker#getConsumer` may create an offset consumer, while the actual offset consumer could have been created by another broker.
Secondly, it only checks if the future of consumer is done and returns immediately if not. It could cause a situation that this metrics update was skipped.

### Modifications

- Modify the condition check in `KafkaTopicConsumer#getGroupConsumers`. Because the offset consumer could be created by `GroupCoordinator#handleSyncGroup` before, here we just check if the consumer existed. Then update the consumer metrics when the future is completed.
- Remove the failed future of offset consumers.
- Add consumer stats tests to guarantee the modification works.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
@@ -15,8 +15,10 @@ package io.streamnative.pulsar.handlers.kop;
 
 import static com.google.common.base.Preconditions.checkState;
 
+import io.streamnative.pulsar.handlers.kop.coordinator.group.OffsetAcker;
 import io.streamnative.pulsar.handlers.kop.utils.KopTopic;
 import java.net.InetSocketAddress;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
@@ -27,6 +29,7 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.pulsar.broker.PulsarServerException;
 import org.apache.pulsar.broker.PulsarService;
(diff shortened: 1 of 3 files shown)


jiazhai (approved):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
```diff
@@ -379,14 +382,21 @@ public void deReference(String topicName) {
...
+            log.warn("Try to get group consumers with an empty group id");
+            return CompletableFuture.completedFuture(null);
         }
+
+        // The future of the offset consumer should be created before in `GroupCoordinator#handleSyncGroup`
+        final OffsetAcker offsetAcker = requestHandler.getGroupCoordinator().getOffsetAcker();
+        final CompletableFuture<org.apache.pulsar.client.api.Consumer<byte[]>> offsetConsumerFuture =
+                offsetAcker.getConsumer(groupId, kafkaPartition);
+        if (offsetConsumerFuture == null) {
+            log.warn("No offset consumer for [group={}] [topic={}]", groupId, kafkaPartition);
```
BewareMyPower, line 395: It may be better to change this log level to debug, I found a lot of this warning in the test environment and will check again.

dockerzhang: I agree. there will be a few of this log if the consumer is not ready.

dockerzhang (approved):
LGTM

BewareMyPower:
I've also changed the log level of empty group check to debug.

```java
    public CompletableFuture<Consumer> getGroupConsumers(String groupId, TopicPartition kafkaPartition) {
        if (StringUtils.isEmpty(groupId)) {
            if (log.isDebugEnabled()) {
                log.debug("Try to get group consumers with an empty group id");
            }
            return CompletableFuture.completedFuture(null);
        }
```

Because in my test environment, there're a lot of logs like

> 13:27:09.864 [bookkeeper-ml-workers-OrderedExecutor-6-0] INFO  io.streamnative.pulsar.handlers.kop.MessageFetchContext - get group name from zk for current connection:/10.0.2.6:49286 groupId:null
13:27:09.864 [bookkeeper-ml-workers-OrderedExecutor-6-0] WARN  io.streamnative.pulsar.handlers.kop.KafkaTopicManager - Try to get group consumers with an empty group id
...
13:31:07.859 [bookkeeper-ml-workers-OrderedExecutor-6-0] INFO  io.streamnative.pulsar.handlers.kop.MessageFetchContext - get group name from zk for current connection:/10.0.2.6:49286 groupId:null
13:31:07.859 [bookkeeper-ml-workers-OrderedExecutor-6-0] WARN  io.streamnative.pulsar.handlers.kop.KafkaTopicManager - Try to get group consumers with an empty group id

It looks like there's something wrong with https://github.com/streamnative/kop/blob/0116024651b097216bc3b3e49f592ebc1334cf4c/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java#L359-L368.

Could you help take a look? @dockerzhang

dockerzhang:
@BewareMyPower I think there is no `zk `  for UT  to store `group id`, so the `getData` returns `null`.

hangc0276 (approved):

BewareMyPower:
@dockerzhang It's the problem in the real environment not UT. The ZK node exists like:

```
ls /client_group_id
[10.0.2.10-consumer-sub-000-G0triB0-1, 10.0.2.10-consumer-sub-000-t3lUJiE-2]
```

The logs occurred in another broker whose IP is `10.0.2.4`.

BewareMyPower:
BTW, the `testDeleteClosedTopics` is very flaky, I'll open a new PR to fix it if it still fails.

BewareMyPower:
Wait for #427 merged.

PR #429, opened 2021-04-07
prevent queue too large and request timeout

### LINKED ISSUE #428: [BUG] there is a lot of close-wait connections when many clients produce or consume
**Describe the bug**
there is a lot of close-wait connections when many clients produce or consume.

**To Reproduce**
1.  create 10 topics，each topic has 6 partitions
2.  produce and consume message with 6 clients for each topic
3. See error

**Expected behavior**

**Screenshots**
image

image

dockerzhang:
fix #428

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
@@ -29,7 +29,6 @@ import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import javax.naming.AuthenticationException;
 
@@ -310,13 +309,18 @@ public abstract class KafkaCommandDecoder extends ChannelInboundHandlerAdapter {
                         kafkaConfig.getRequestTimeoutMs(),
                         TimeUnit.MILLISECONDS), response.getRequest());
                 channel.writeAndFlush(result);
-            } catch (ExecutionException | TimeoutException | InterruptedException e) {
-                // get response result may timeout.
-                log.error("request {}: error to get Response ByteBuf: ",
-                        response.getRequest().getHeader().apiKey(), e.getCause());
+            } catch (Exception e) {
+                // catch ExecutionException if this future completed exceptionally
+                if (e instanceof ExecutionException) {
+                    log.error("request {}: error to get Response ByteBuf: ",
+                            response.getRequest().getHeader(), e.getCause());
+                } else {
+                    log.error("request {} timed out or was interrupted",
+                            response.getRequest().getHeader());
+                }
                 // make sure there is response to client when timeout.
                 responseQueue.add(response);
-                writeAndFlushWhenInactiveChannelOrTimeout(channel);
+                writeAndFlushWhenInactiveChannelOrException(channel);
             } finally {
                 try {
                     if (response.getResponseFuture().get() instanceof ResponseCallbackWrapper) {
@@ -330,7 +334,7 @@ public abstract class KafkaCommandDecoder extends ChannelInboundHandlerAdapter {
     }
 
     // return all the current command before a channel close. return Error response for all pending request.
-    protected void writeAndFlushWhenInactiveChannelOrTimeout(Channel channel) {
+    protected void writeAndFlushWhenInactiveChannelOrException(Channel channel) {
         // loop from first responseFuture, and return them all
         while (responseQueue != null && responseQueue.peek() != null) {
             try {
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
@@ -29,6 +29,7 @@ import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import javax.naming.AuthenticationException;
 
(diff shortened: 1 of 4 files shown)


BewareMyPower:
After discussing with @hangc0276 yesterday, the request queue is used to keep the order of responses in each channel.

You can see some error logs in integration tests

```
org.apache.kafka.common.protocol.types.SchemaException: Error reading field 'brokers': Error reading array of size 65535, only 19 bytes available
	at org.apache.kafka.common.protocol.types.Schema.read(Schema.java:77)
	at org.apache.kafka.common.protocol.ApiKeys.parseResponse(ApiKeys.java:294)
	at org.apache.kafka.clients.NetworkClient.parseStructMaybeUpdateThrottleTimeMetrics(NetworkClient.java:637)
```

The exception was thrown by Kafka clients not KoP, so the only reason is that KoP sent an error response back to client.

Therefore, we need to find a new way to fix the connections not disconnected.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
```diff
@@ -294,11 +303,16 @@ protected void writeAndFlushResponseToClient(Channel channel) {
                         response.getRequest().getHeader());
                 }
 
-                ByteBuf result = responseToByteBuf(response.getResponseFuture().get(), response.getRequest());
+                ByteBuf result = responseToByteBuf(response.getResponseFuture().get(30,
+                        TimeUnit.SECONDS), response.getRequest());
```
BewareMyPower, line 307: Could you explain why use the hard coded 30 seconds here?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
```diff
@@ -57,8 +58,12 @@
     @Getter
     protected AtomicBoolean isActive = new AtomicBoolean(false);
 
-    // Queue to make request get responseFuture in order.
-    private Queue<ResponseAndRequest> requestsQueue;
+    // Queue to make response get responseFuture in order.
+    private Queue<ResponseAndRequest> responseQueue;
+
+    final int maxQueuedRequests = 500;
```
BewareMyPower, line 64: > Use explicit scoping instead of the default package private level

Follow what Codacy check suggested.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
```diff
@@ -294,11 +303,16 @@ protected void writeAndFlushResponseToClient(Channel channel) {
...
                 channel.writeAndFlush(result);
             } catch (Exception e) {
-                // should not comes here.
-                log.error("error to get Response ByteBuf:", e);
+                // get response result may timeout.
+                log.error("request {}: error to get Response ByteBuf: ",
+                        response.getRequest().getHeader().apiKey(), e);
+                // make sure there is response to client.
+                responseQueue.add(response);
+                writeAndFlushWhenInactiveChannelOrTimeout(channel);
```
BewareMyPower, line 315: Catch the specific exceptions to handle the timeout

```java
catch (ExecutionException | TimeoutException e) {
    // ...
} catch (Exception e) {
    // should not come here
```

Also, IMO, for the `TimeoutException`, we needn't to log the exception stack, but for the `ExecutionException`, we need to log `e.getCause()` instead of `e` itself.

See the documentation of `CompletableFuture#get`:

```java
    /**
     * Waits if necessary for at most the given time for this future
     * to complete, and then returns its result, if available.
     *
     * @param timeout the maximum time to wait
     * @param unit the time unit of the timeout argument
     * @return the result value
     * @throws CancellationException if this future was cancelled
     * @throws ExecutionException if this future completed exceptionally
     * @throws InterruptedException if the current thread was interrupted
     * while waiting
     * @throws TimeoutException if the wait timed out
     */
    public T get(long timeout, TimeUnit unit)
        throws InterruptedException, ExecutionException, TimeoutException {
```

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
dockerzhang: reference to `request.timeout.ms` in kafka.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
```diff
@@ -294,11 +306,17 @@ protected void writeAndFlushResponseToClient(Channel channel) {
...
+                        kafkaConfig.getRequestTimeoutMs(),
+                        TimeUnit.MILLISECONDS), response.getRequest());
                 channel.writeAndFlush(result);
-            } catch (Exception e) {
-                // should not comes here.
-                log.error("error to get Response ByteBuf:", e);
+            } catch (ExecutionException | TimeoutException | InterruptedException e) {
+                // get response result may timeout.
+                log.error("request {}: error to get Response ByteBuf: ",
+                        response.getRequest().getHeader().apiKey(), e.getCause());
```
BewareMyPower, line 316: Maybe I didn't express well before. We should use `getCause()` only for `ExecutionException`, because it's thrown  from `CompletableFuture.exceptionally` and `getCause()` is the real cause. For `TimeoutException` and `InterruptedException`, `getCause()` returns null.

BewareMyPower, line 316:
```suggestion
                // get response result may timeout or fail
                if (e instanceof ExecutionException) {
                    log.error("request {}: error to get Response ByteBuf: ",
                            response.getRequest().getHeader(), e.getCause());
                } else {
                    log.error("request {} timed out or was interrupted", response.getRequest().getHeader());
                }
```

Two main changes:
1. Only log detail stacks for `ExecutionException`
2. Log the header instead of the api key, because the header contains correlation id that may be helpful to see which request failed.

BewareMyPower (changes requested):

BewareMyPower:
I've added a suggestion, PTAL again.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
dockerzhang: I got it, thanks for your suggestion.

BewareMyPower (approved):

BewareMyPower:
I think we can ignore the Codacy check here. It recommends to catch multiple exception types instead of `instanceof` check. But here we can use `instanceof` to reduce repeated code.

PR #440, opened 2021-04-12
[WIP] use direct memory instead of heap for kafka entry formatter decode 

hangc0276:
### Motivation
In KafkaEntryFormatter#decode, it will allocate a new ByteBuffer to store all entries' payload. However, the current implementation is use heap memory, which will lead to frequent GC.

### Changes
1. use direct memory instead of heap memory to store the entries' payload


BewareMyPower:
Mark it as WIP first because it may cause direct memory OOM quickly.

image

image

PR #443, opened 2021-04-13
Use setMetadataFromEntryData API to retrieve offset

### LINKED ISSUE #332: [BUG] Offset in send callback is not continuous when batching is enabled

BewareMyPower:
Fixes #332 

The offset in produce callback may be not accurate when the messages are sent in batch, see https://github.com/streamnative/kop/issues/332#issuecomment-762712235 for detail explanation.

Since https://github.com/apache/pulsar/pull/9257 introduced a new API that supports fetching some metadata from the entry data, we can use this API to get the accurate offset.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessagePublishContext.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessagePublishContext.java
@@ -13,15 +13,16 @@
  */
 package io.streamnative.pulsar.handlers.kop;
 
+import io.netty.buffer.ByteBuf;
 import io.netty.util.Recycler;
 import io.netty.util.Recycler.Handle;
-import io.streamnative.pulsar.handlers.kop.utils.MessageIdUtils;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.bookkeeper.mledger.ManagedLedger;
 import org.apache.pulsar.broker.service.Topic;
 import org.apache.pulsar.broker.service.Topic.PublishContext;
+import org.apache.pulsar.common.api.proto.BrokerEntryMetadata;
+import org.apache.pulsar.common.protocol.Commands;
 
 /**
  * Implementation for PublishContext.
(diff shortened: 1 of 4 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessagePublishContext.java
```diff
@@ -33,7 +34,23 @@
...
+
+    @Override
+    public void setMetadataFromEntryData(ByteBuf entryData) {
+        try {
+            final BrokerEntryMetadata brokerEntryMetadata = Commands.peekBrokerEntryMetadataIfExist(entryData);
+            if (brokerEntryMetadata == null) {
+                throw new IllegalStateException(
+                        "There's no BrokerEntryData, check if your broker configured brokerEntryMetadataInterceptors");
+            }
+            baseOffset = brokerEntryMetadata.getIndex() - (numberOfMessages - 1);
```
jiazhai, line 47: There may be cases, that entryMetdataEnabled, but Index not used.

jiazhai (approved):
overall lgtm. left a minor comments

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessagePublishContext.java
jiazhai, line 47: There maybe cases that  brokerEntryMetadata enabled, but index not used, and only contains default value. we should handle this case？

BewareMyPower:
It looks like `SchemaRegistryTest.testAvroProduceAndConsume` always failed but it succeeded in my local env. I'll dig into the reason later.

PR #168, opened 2020-08-03
Added a new MetadataUtils with an implementation of createKafkaMetadataIfMissing…

wmccarley:
…that does not overwrite the cluster metadata. Modified several use cases to support the change.

Logic for the createKafkaMetadataIfMissing implementation is as follows:
 - If the tenant does not exist it will be created
 - If the tenant exists but the allowedClusters list does not include the cluster this method will add the cluster to the allowedClusters list
 - If the namespace does not exist it will be created
 - If the namespace exists but the replicationClusters list does not include the cluster this method will add the cluster to the replicationClusters list
 - If the offset topic does not exist it will be created
 - If the offset topic exists but some partitions are missing, the missing partitions will be created

PR #445, opened 2021-04-18
Create cluster if it doesn't exist during metadata initialisation

### LINKED ISSUE #185: [BUG] Pulsar standalone failed to start without any existed data
**Describe the bug**
The first time to run Pulsar standalone would fail.

**To Reproduce**
- KoP 0.3.0
- Pulsar 2.6.1

1. Download [apache-pulsar-2.6.1.tar.gz](https://www.apache.org/dyn/mirrors/mirrors.cgi?action=download&filename=pulsar/pulsar-2.6.1/apache-pulsar-2.6.1-bin.tar.gz) and uncompress it
2. Modify `conf/standalone.conf`:
    ```properties
    protocolHandlerDirectory=./protocols
    messagingProtocols=kafka
    allowAutoTopicCreationType=partitioned
    listeners=PLAINTEXT://127.0.0.1:9092
    advertisedAddress=127.0.0.1
    ```
3. Run `./bin/pulsar-daemon start standalone`
4. The Pulsar standalone failed and exit with following log:

```
[main] ERROR io.streamnative.pulsar.handlers.kop.KafkaProtocolHandler - initGroupCoordinator failed with
org.apache.pulsar.broker.PulsarServerException$NotFoundException: Configured cluster does not exist
    at io.streamnative.pulsar.handlers.kop.utils.MetadataUtils.createKafkaMetadataIfMissing(MetadataUtils.java:82) ~[?:?]
    at io.streamnative.pulsar.handlers.kop.KafkaProtocolHandler.initGroupCoordinator(KafkaProtocolHandler.java:312) ~[?:?]
```

However, if we start the standalone with the default config, i.e. no messaging protocol was specified. Then stop the standalone, modify the config as step 2 above, start, it worked.

Seems that when the KoP called `initGroupCoordinator`, the admin couldn't found the cluster info:

```java
    public static void createKafkaMetadataIfMissing(PulsarAdmin pulsarAdmin, KafkaServiceConfiguration conf)
        throws PulsarServerException, PulsarAdminException {
        String cluster = conf.getClusterName();
        /* ... */

        try {
            Clusters clusters = pulsarAdmin.clusters();
            if (!clusters.getClusters().contains(cluster)) {
                throw new PulsarServerException.NotFoundException("Configured cluster does not exist");
            } else {
                /* ... */
            }
```

But if the standalone started while the metadata of cluster already existed, the cluster could be found.

**Expected behavior**
Pulsar standalone can start without any existed data

**Additional context**
I also tried AoP, the behavior didn't happen, it might be KoP's problem.

KoP 0.2.0 with Pulsar 2.5.0 also works well.

josephglanville:
This case is usually encountered when attempting to use KoP with
standalone mode. When new behavior was introduced in #168 standalone
mode was broken due to sample namespace setup etc not taking place until
after the broker is initalised (which in turn initialises protocol
handlers).

Fixes #185

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
@@ -60,6 +60,7 @@ import org.apache.pulsar.client.impl.PulsarClientImpl;
 import org.apache.pulsar.common.naming.NamespaceBundle;
 import org.apache.pulsar.common.naming.NamespaceName;
 import org.apache.pulsar.common.naming.TopicName;
+import org.apache.pulsar.common.policies.data.ClusterData;
 import org.apache.pulsar.common.util.FutureUtil;
 
 /**
@@ -370,7 +371,11 @@ public class KafkaProtocolHandler implements ProtocolHandler {
             .build();
 
         PulsarAdmin pulsarAdmin = service.pulsar().getAdminClient();
-        MetadataUtils.createOffsetMetadataIfMissing(pulsarAdmin, kafkaConfig);
+        ClusterData clusterData = new ClusterData(service.getPulsar().getWebServiceAddress(),
+                                                  service.getPulsar().getWebServiceAddressTls(),
+                                                  service.getPulsar().getBrokerServiceUrl(),
+                                                  service.getPulsar().getBrokerServiceUrlTls());
+        MetadataUtils.createOffsetMetadataIfMissing(pulsarAdmin, clusterData, kafkaConfig);
 
 
         this.groupCoordinator = GroupCoordinator.of(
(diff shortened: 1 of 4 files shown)


josephglanville:
I see in numerous places references to TLS not being supported, is that relevant here when constructing the cluster data?

BewareMyPower:
Thanks for your contribution! The standalone problem has existed for a long time and I nearly forgot to look into the issue. BTW, recently schemaRegistryTest and Kafka Streams related tests were not stable in github CI. I think we can merge this PR even if they failed.

> I see in numerous places references to TLS not being supported, is that relevant here when constructing the cluster data?

IMO it's not. The cluster metadata is used for Pulsar not KoP. And if you didn't configure TLS related metadata, it wouldn't be written to Z-node.

BewareMyPower (approved):

PR #447, opened 2021-04-20
[ENHANCEMENT] memory tuning for decode phrase

wuzhanpeng:
## Motivation
Keep the data in the direct memory of nio as much as possible to reduce the damage to performance.

## Modifications
1. modify `ByteBuf` directly instead of `MemoryRecords`
2. delay the release timing of `Entry` until after writing to the network channel

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
@@ -52,6 +52,7 @@ import org.apache.kafka.common.requests.FetchRequest;
 import org.apache.kafka.common.requests.FetchResponse;
 import org.apache.kafka.common.requests.FetchResponse.PartitionData;
 import org.apache.kafka.common.requests.IsolationLevel;
+import org.apache.kafka.common.requests.ResponseCallbackWrapper;
 import org.apache.pulsar.broker.service.Consumer;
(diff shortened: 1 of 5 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
```diff
@@ -398,6 +398,14 @@ private void readMessagesInternal(KafkaHeaderAndRequest fetch,
                                     responseData,
                                     ((Integer) THROTTLE_TIME_MS.defaultValue),
                                     ((FetchRequest) fetch.getRequest()).metadata().sessionId()));
+                    resultFuture.whenComplete((r, e) -> {
+                        // ATTENTION: Please note that the release of entries should be placed after
+                        // future complete operation to ensure that the data has been successfully
+                        // written to the network channel
+                        responseValues.entrySet().forEach(topicPartitionListEntry -> {
+                            topicPartitionListEntry.getValue().forEach(Entry::release);
+                        });
```
BewareMyPower, line 407: I think there's still a race condition. The previous `whenComplete` for response future is

```java
                ctx.channel().eventLoop().execute(() -> {
                    writeAndFlushResponseToClient(channel);
                });
```

We can see `writeAndFlushResponseToClient` may not be executed immediately.

hangc0276 (approved):

BewareMyPower (approved):

PR #451, opened 2021-04-22
Add inner topic protection

hangc0276:
### Motivation
For inner topic, such as, __consumer_offsets, __transaction_stats, and system topic, such as, __change_event, they are expose to all users and can produce / consumer message to / from those topic, which will cause the cluster crash.

### Changes
1. add produce protection for inner topic on KOP side.

TODO
1. for consume protection, it's in testing.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/PendingProduce.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/PendingProduce.java
@@ -126,6 +126,12 @@ public class PendingProduce {
             throw new RuntimeException(e);
         }
 
+        if (persistentTopic.isSystemTopic()) {
+            log.error("Not support produce message to system topic: {}", persistentTopic);
+            responseFuture.complete(new PartitionResponse(Errors.INVALID_TOPIC_EXCEPTION));
+            return;
+        }
+
         if (log.isDebugEnabled()) {
             log.debug("publishMessages for topic partition: {}, records size is {}", partitionName, numMessages);
         }
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/PendingProduce.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/PendingProduce.java
@@ -126,6 +126,12 @@ public class PendingProduce {
             throw new RuntimeException(e);
         }
 
+        if (persistentTopic.isSystemTopic()) {
+            log.error("Not support producing message to system topic: {}", persistentTopic);
+            responseFuture.complete(new PartitionResponse(Errors.INVALID_TOPIC_EXCEPTION));
+            return;
+        }
+
         if (log.isDebugEnabled()) {
             log.debug("publishMessages for topic partition: {}, records size is {}", partitionName, numMessages);
         }
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/PendingProduce.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/PendingProduce.java
@@ -126,12 +126,6 @@ public class PendingProduce {
             throw new RuntimeException(e);
         }
 
-        if (persistentTopic.isSystemTopic()) {
-            log.error("Not support produce message to system topic: {}", persistentTopic);
-            responseFuture.complete(new PartitionResponse(Errors.INVALID_TOPIC_EXCEPTION));
-            return;
-        }
-
         if (log.isDebugEnabled()) {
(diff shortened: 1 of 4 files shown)


BewareMyPower:
Just a doubt about your TODO: do we need to protect consuming? If I didn't remember wrong, Kafka allows consumers to read `__consumer_offsets` and some admin tools are implemented by this way.

hangc0276:
> Just a doubt about your TODO: do we need to protect consuming? If I didn't remember wrong, Kafka allows consumers to read `__consumer_offsets` and some admin tools are implemented by this way.

OK, the consume may expose to users.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/InnerTopicProtectionTest.java
```diff
@@ -0,0 +1,185 @@
...
+            producer.send(new ProducerRecord<>(topic, value), ((metadata, exception) -> {
+                if (assertException) {
+                    Assert.assertNotNull(exception);
+                } else {
+                    Assert.assertNull(exception);
+                }
+            })).get();
+        } catch (Exception e) {
+
+        }
```
BewareMyPower, line 174:
```suggestion
        try {
            producer.send(new ProducerRecord<>(topic, value)).get();
        } catch (Exception e) {
            if (assertException) {
                Assert.assertEquals(e.getCause().getMessage(),
                        "The request attempted to perform an operation on an invalid topic.");
            } else {
                Assert.fail();
            }
        }
```

BewareMyPower (changes requested):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/InnerTopicProtectionTest.java
```diff
@@ -0,0 +1,185 @@
...
+    protected void resetConfig() {
+        offsetsTopicNumPartitions = 16;
+        conf = resetConfig(
+            brokerPort,
+            brokerWebservicePort,
+            kafkaBrokerPort);
+
+        log.info("Ports --  broker: {}, brokerWeb:{}, kafka: {}",
+            brokerPort, brokerWebservicePort, kafkaBrokerPort);
+    }
```
BewareMyPower, line 101: Could you explain why overwrite `resetConfig` here? I tried to remove the `resetConfig` code, and the tests still passed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/PendingProduce.java
```diff
@@ -126,6 +126,12 @@ public void publishMessages() {
             throw new RuntimeException(e);
         }
 
+        if (persistentTopic.isSystemTopic()) {
+            log.error("Not support produce message to system topic: {}", persistentTopic);
+            responseFuture.complete(new PartitionResponse(Errors.INVALID_TOPIC_EXCEPTION));
+            return;
+        }
```
BewareMyPower, line 133: I noticed this case was not protected by test case. If there's something wrong with system topics, you can open a new PR to handle it.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/InnerTopicProtectionTest.java
hangc0276: The test can cover turn on system topic and topic policy with resetConfig. However, KOP can't work when turn on topic policy. In the end, it turn off the topic policy in resetConfig. We can turn on when it fixed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/PendingProduce.java
hangc0276: When turn on system topic, the KOP doesn't work. I turn off it in the test.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/InnerTopicProtectionTest.java
BewareMyPower: So I think these two `resetConfig` could be removed as well.

BewareMyPower (changes requested):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/InnerTopicProtectionTest.java
BewareMyPower: OK

BewareMyPower:
I see it works well with system topic enabled. Is https://github.com/apache/pulsar/pull/10341 necessary?

BewareMyPower:
Sorry I just misunderstood.

BewareMyPower (approved):

PR #465, opened 2021-05-04
Refactor writeAndFlushResponseToClient

BewareMyPower:
### Motivation

https://github.com/streamnative/kop/pull/429 introduced a request timeout to avoid some requests being blocked by the previous requests. However, it changes the behavior of `writeAndFlushResponseToClient`.

Before #429, `writeAndFlushResponseToClient` removed all completed response futures in the head of `responseQueue` and then sent the completed response to client.

After #429, `writeAndFlushResponseToClient` removed all response futures, and use `CompletableFuture#get(long timeout, TimeUnit unit)` to wait until all futures are completed or expired.

We shouldn't wait any `CompletableFuture` in a `CompletableFuture`'s callback.

### Modifications

In `writeAndFlushResponseToClient`, only remove completed **or expired** response futures. Here we combine `peek` and `boolean remove(Object o)` for thread safety.

For responses that are expired or completed exceptionally, just skip them and don't send any response to client because Kafka client has a retry mechanism. When Kafka processes an expired request, it also doesn't send any response to client.

For responses that are completed normally, send the response and release the internal Netty buffers if necessary.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
@@ -26,9 +26,7 @@ import java.net.SocketAddress;
 import java.nio.ByteBuffer;
 import java.util.Queue;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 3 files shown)


BewareMyPower:
@dockerzhang @wuzhanpeng PTAL

wuzhanpeng:
LGTM 👍

jiazhai (approved):

PR #488, opened 2021-05-10
Fix publish throttling and LoadManager API for pulsar upgrade

BewareMyPower:
This PR upgrades pulsar dependency to 2.8.0-rc-202105092228, which has two major API changes.

https://github.com/apache/pulsar/pull/10391 changed `LoadManager` API so that `MetadataCache` is used instead of `ZookeeperCache` in this PR.

https://github.com/apache/pulsar/pull/7406 changed the throttling strategy. However, currently KoP is different from Pulsar that the produce and its callback may be in different threads. KoP calls `PersistentTopic#publishMessages` in a callback of `KafkaTopicManager#getTopic` if the returned future is not completed immediately. Otherwise, it's called just in the I/O thread. Therefore, here we still use a **channel based** publish bytes stats for throttling, while https://github.com/apache/pulsar/pull/7406 uses a **thread based** publish bytes stats.

The other refactors are:
1. Change the throttling related fields from `InternalServerCnx` to `KafkaRequestHandler`.
2. Use `BrokerService#getPausedConnections` to check if the channel's auto read is disabled and modify the tests as well.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/InternalServerCnx.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/InternalServerCnx.java
@@ -13,9 +13,7 @@
  */
 package io.streamnative.pulsar.handlers.kop;
 
-import com.google.common.annotations.VisibleForTesting;
 import java.net.InetSocketAddress;
-import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.pulsar.broker.service.Producer;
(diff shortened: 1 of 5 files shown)


dockerzhang (approved):
+1

BewareMyPower:
@jiazhai PTAL, it looks like the dependency from bintray is not available from now. See another PR's CI: https://github.com/streamnative/kop/pull/495/checks?check_run_id=2568314587

> Failed to execute goal org.apache.maven.plugins:maven-remote-resources-plugin:1.5:process (process-resource-bundles) on project pulsar-protocol-handler-kafka-parent: Failed to resolve dependencies for one or more projects in the reactor. Reason: Unable to get dependency information for org.apache.pulsar:pulsar-broker:jar:2.8.0-rc-202104202206: Failed to retrieve POM for org.apache.pulsar:pulsar-broker:jar:2.8.0-rc-202104202206: Could not transfer artifact org.apache.pulsar:pulsar-broker:pom:2.8.0-rc-202104202206 from/to bintray-streamnative-maven (https://dl.bintray.com/streamnative/maven): authorization failed for https://dl.bintray.com/streamnative/maven/org/apache/pulsar/pulsar-broker/2.8.0-rc-202104202206/pulsar-broker-2.8.0-rc-202104202206.pom, status: 403 Forbidden

jiazhai (approved):

hangc0276 (approved):

PR #490, opened 2021-05-11
add produce/fetch timeout

dockerzhang:
KoP uses `requestTimeoutMs` config to prevent asynchronous `responseFuture` timeout, but `FetchRequest` and `ProduceRequest` have it's own timeout parameters which are brought in from the client, we should replace the default request timeout for `Produce` and `Fetch`.


BewareMyPower (approved):

BewareMyPower:
I approved this PR first.

Maybe in the future we can find a way to add tests for produce/fetch timeout. At this moment, I didn't find the timeout tests in Kafka for reference.

Issue #493, opened 2021-05-12
[BUG] There're still bugs during unload after #404


BewareMyPower:
**Describe the bug**
#404 tried to fix the bundle unload bug by removing topics that are not owned by current broker from static caches in namespace bundle ownership listener. But the unload tests are still easy to fail because somehow the `KafkaTopicConsumerManager` (let's say TCM in short) still cached some null values. I'll give the detail analysis at the last section.

**To Reproduce**
Run unload tests for several times:

```bash
mvn test -pl tests -Dtest='DistributedClusterTest#testMutiBrokerUnloadReload'
```

**Expected behavior**
The test should be stable to pass.

**Additional context**
This section gives the related logs for one failure. The test failed because `kafkaConsumeCommitMessage` timed out.

First, we can grep these logs: https://github.com/streamnative/kop/blob/d8f2bbd8eadcc7bead561b499493ae84caadd0c0/tests/src/test/java/io/streamnative/pulsar/handlers/kop/DistributedClusterTest.java#L228-L232

> 17:42:54.275 [TestNG-method=testMutiBrokerUnloadReload-1:io.streamnative.pulsar.handlers.kop.DistributedClusterTest@230] DEBUG io.streamnative.pulsar.handlers.kop.DistributedClusterTest - kConsumer kopMutiBrokerUnloadReload10consumer-group-1 start poll message: 30
> ...
> 17:43:12.291 [TestNG-method=testMutiBrokerUnloadReload-1:io.streamnative.pulsar.handlers.kop.DistributedClusterTest@230] DEBUG io.streamnative.pulsar.handlers.kop.DistributedClusterTest - kConsumer kopMutiBrokerUnloadReload10consumer-group-1 start poll message: 30

Then from client's logs:

> 17:42:54.175 [TestNG-method=testMutiBrokerUnloadReload-1:org.apache.kafka.clients.consumer.internals.Fetcher$1@227] DEBUG org.apache.kafka.clients.consumer.internals.Fetcher - [Consumer clientId=consumer-1, groupId=consumer-group-1] Fetch READ_UNCOMMITTED at offset 4 for partition kopMutiBrokerUnloadReload10-2 returned fetch data (error=NOT_LEADER_FOR_PARTITION, highWaterMark=-1, lastStableOffset = -1, logStartOffset = -1, abortedTransactions = null, recordsSizeInBytes=0)
[...]

PR #495, opened 2021-05-12
Fix null KafkaTopicConsumerManager not removed

BewareMyPower:
Fixes https://github.com/streamnative/kop/issues/493

This bug was introduced by https://github.com/streamnative/kop/pull/473. In `MessageFetchContext#handleFetch`, when the `KafkaTopicConsumerManager`'s future is completed with null, we should remove the future from `KafkaTopicManager#consumerTopicManagers`.

In addition, this PR adds some refactors for `consumerTopicManagers`:
1. Don't use getter to expose this field, use methods to operate it instead.
2. Check null for completed future before close `KafkaTopicConsumerManager`.


BewareMyPower:
After this PR, I've tested `DistributedClusterTest#testMutiBrokerUnloadReload` for 15 times in a row and it never failed.

BewareMyPower:
@jiazhai PTAL.

dockerzhang (approved):

jiazhai (approved):

PR #500, opened 2021-05-14
Use LinkedBlockingQueue as the container of ResponseAndRequest

BewareMyPower:
### Motivation
There're `requestQueue` and `responseQueue` in each channel. `responseQueue` stores the pending `ResponseAndRequest` to make responses be sent in order. However, `requestQueue` is just to limit the max number of requests and it's a fake queue because it only maintains a request count.

Sometimes an exception may be thrown like

>  [pulsar-io-28-27] WARN  io.netty.util.concurrent.AbstractEventExecutor - A task raised an exception. Task: io.streamnative.pulsar.handlers.kop.KafkaCommandDecoder$$Lambda$1097/0x000000084087f440@60104314
java.lang.IllegalStateException: poll() is called when count is 0

It's because when an empty `requestQueue` called `poll()` method. It may be caused by the `close()` method, which clears the `requestQueue`. However, the pending response future may complete and trigger `writeAndFlushResponseToClient`, which triggers another `poll()` call. In addition, `writeAndFlushResponseToClient` may not remove any responses from `responseQueue` but each time it's called the `requestQueue.poll()` will be called.

### Modifications
- Use `LinkedBlockingQueue` to store `ResponseAndRequest` and limit the max number of requests, use the `requestQueue` as its name to avoid misunderstanding.
- When the channel is closed, before clear the `requestQueue`, cancel all response futures and release the response buffer when it's a FETCH response. If the response future is cancelled, don't call `writeAndFlushResponseToClient` in callback.
- Check if the response future is completed in `DelayedOperation`'s callback because the callback may be triggered twice.
- If a response future is completed, send the response even if it's expired.


BewareMyPower:
`DifferentNamespaceTestBase` still failed, I'll first take a look.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
```diff
@@ -85,9 +81,20 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
...
 
     protected void close() {
+        assert !isActive.get();
+        // Clear the request queue
+        log.info("close channel {} with {} pending responses", ctx.channel(), requestQueue.size());
+        while (true) {
+            final ResponseAndRequest responseAndRequest = requestQueue.poll();
+            if (responseAndRequest != null) {
+                // Trigger writeAndFlushResponseToClient immediately, but it will do nothing because isActive is false
+                responseAndRequest.getResponseFuture().complete(null);
```
hangc0276, line 91: May cause direct memory leak.

BewareMyPower: It's a serious problem. I'll rethink how to resolve it.

BewareMyPower:
@hangc0276 From your comment, I just thought of that this problem also existed in current code because the `responseQueue` will be cleared. Even it was not cleared, because `isActive` was false now, `writeAndFlushResponseToClient` wouldn't remove the response and the callback wouldn't be called. And for the case that the response is expired, we also completed it with null. It might also cause direct memory leak.

I think we need another way to handle this case.

BewareMyPower:
@hangc0276 I've pushed the fix, PTAL again.

BewareMyPower:
It looks like after rebasing to master, some tests failed. I'll take a look.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/DifferentNamespaceTestBase.java
```diff
@@ -99,8 +98,7 @@ void testSimpleProduceAndConsume(String topic) {
                     log.info("Successfully send {} to {}-partition-{}",
                             key, recordMetadata.topic(), recordMetadata.partition());
                 } else {
-                    log.error("Failed to send {}", key);
-                    fail("Failed to send " + key);
+                    log.error("Failed to send {}: {}", key, e.getMessage());
```
hangc0276, line 101: why delete fail check?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
```diff
@@ -85,9 +82,20 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
     }
 
     protected void close() {
+        assert !isActive.get();
```
hangc0276, line 85: if assert failed, it will throw exception?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
```diff
@@ -168,18 +175,25 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
 
             CompletableFuture<AbstractResponse> responseFuture = new CompletableFuture<>();
             responseFuture.whenComplete((response, e) -> {
-                // Do nothing if it's triggered when responseFuture is expired
-                if (response == null) {
+                if (e != null) {
```
hangc0276, line 178: we should check whether response is null, otherwise it will cause NPE in `responseToByteBuf` and `writeAndFlush` stage.

hangc0276 (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
```diff
@@ -67,7 +64,7 @@
     public KafkaCommandDecoder(StatsLogger statsLogger, KafkaServiceConfiguration kafkaConfig) {
         this.requestStats = new RequestStats(statsLogger);
         this.kafkaConfig = kafkaConfig;
-        this.requestQueue = new FakeArrayBlockingQueue(kafkaConfig.getMaxQueuedRequests());
+        this.requestQueue = new LinkedBlockingQueue<>(kafkaConfig.getMaxQueuedRequests());
```
hangc0276, line 67: ArrayBlockingQueue maybe more effective

hangc0276 (changes requested):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/DifferentNamespaceTestBase.java
BewareMyPower: The fail check in callback can't stop the test immediately. It only makes the thread broken and prints unrelated logs. Here we send these messages asynchronously so we also don't want to stop the test if one produce failed. Otherwise here we should use synchronously send. (Whether to use async send here is another issue)

Removing the `fail` check also helps figuring out all messages' send result for debugging.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
BewareMyPower: No. Expect we enabled `-ea` option. `assert` is used for logic assumption. Because we want to make sure `KafkaCommandDecoder#close` is called in the block of `if (isActive.getAndSet(false))`, which means `isActive.get()` must return false expect there's a bug with Java standard library.

But I'll remove this line. Because I used `assert` here to ensure `isActive` is false and when `writeAndFlushResponseToClient` is called by null completion, then the loop will be skipped. However, we use `cancel` now so that `writeAndFlushResponseToClient` won't be called and the assertion becomes useless.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
BewareMyPower: OK

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
BewareMyPower: Kafka also uses `LinkedBlockingDeque` for `responseQueue`, see

https://github.com/apache/kafka/blob/ad91c5edf09af83cf44e2be71f19407ade1896bb/core/src/main/scala/kafka/network/SocketServer.scala#L823

(BTW, I don't understand why it uses deque but not queue. Only `put` and `poll` methods are used)

Here I concern more about the memory usage over performance. `ArrayBlockingQueue` allocates 500 (the default `maxQueuedRequests` value) objects when it's constructed. For N channels, there're 500*N objects that were allocated on heap. Maybe it looks like not very much, but it's still a waste.

BewareMyPower:
The `DifferentNamespacePulsarTest` is flaky, I'll open a new PR to fix it.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
BewareMyPower: I think it won't cause NPE, because the null `response` is never used. I'll just add the null check because the behavior is  unexpected. The response future should not be completed with null expect some error happened.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/DifferentNamespaceTestBase.java
BewareMyPower: No matter if `fail` check exists, if any send failed, the test will eventually fail for timeout because consumer cannot receive `numMessages` messages in test's timeout.

BewareMyPower:
@hangc0276 PTAL again

tests/src/test/java/io/streamnative/pulsar/handlers/kop/DifferentNamespaceTestBase.java
hangc0276: we'd better use try... catch to deal with the produce exception instead of use callback mode.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
```diff
@@ -171,18 +177,29 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
...
+                        }
+                    } else {
+                        log.error("[{}] Request {} is completed with exception",
+                                ctx.channel(), kafkaHeaderAndRequest.getHeader(), e);
+                    }
+                    return;
+                }
                 if (response == null) {
+                    log.error("[{}] Unexpected null completed future for request {}",
+                            ctx.channel(), kafkaHeaderAndRequest.getHeader());
```
hangc0276, line 194: when response is null or cancelled by timeout, whether we should return an exception to client, otherwise, the client doesn't know what happens in the server side, and no error code returned to client.

hangc0276 (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
BewareMyPower: There're two cases that the request is cancelled:
1. Channel is closing. We needn't to send a response because Kafka client will know the channel is closed after server side calls `ctx.close()`.
2. The request is expired. We added a server side `request.timeout.ms` before, so it's triggered in `writeAndFlushResponseToClient` immediately. But I still didn't send back any response when it's expired. Because it's just to avoid blocking the followed response futures. Client side has its own `request.timeout.ms` that should be the **real** timeout.

As for when response is null, actually there's no possibility that response will be null from the code analysis. I check it only for the unexpected case and still rely on the client's own `request.timeout.ms`.

hangc0276: What will happen when client configured `request.timeout.ms` value greater than server side?

BewareMyPower:
@hangc0276 I adjusted the logic of response future's callback. The key point is that when the future is cancelled, skip `writeAndFlushResponseToClient`. 

There're two main reasons:
1. If the response future is cancelled during `close`, the `responseQueue` will be cleared by `close()` in loop. We don't want to trigger `writeAndFlushResponseToClient` in each loop. Even it will do nothing because `isActive` is false.
2. If the response future is cancelled by expired request, we'll print the log and send the error response in `writeAndFlushResponseToClient`.

If the response future is completed exceptionally or completed with null value, it will be handled in `writeAndFlushResponseToClient` later so we don't need to do anything in `whenComplete`.

jiazhai (approved):

hangc0276 (approved):
LGTM

PR #506, opened 2021-05-16
Fix topic creation may not complete after METADATA response is sent

BewareMyPower:
### Motivations

Sometimes `DifferentNamespaceTest` may fail, after checking the logs I found send failed.

> Failed to send 0: This server is not the leader for that topic-partition

Kafka 2.0 client's default `retry` config is 0, so it won't attempt to send message again.

It's because when KoP handled PRODUCE request, the `PersistentTopic` instance of broker was not created. It's because when a topic was automatically created, an asynchronous method was called but METADATA response future didn't wait for it. The followed lookup request doesn't need a topic to be created.

### Modifications

When a METADATA request is handled and a topic is created automatically, wait until the topic creation is completed.


hangc0276 (approved):
LGTM

dockerzhang (approved):
+1

jiazhai (approved):

Issue #513, opened 2021-05-18
[BUG] High CPU usage and connection may be broken when a consumer is connected


BewareMyPower:
**Describe the bug**
When a consumer is connected to a KoP cluster with a single broker, the CPU usage became very high and connection may be broken.

**To Reproduce**
When I run a performance test, with 50 MB/s throughput and 16 partitions, if there's only a single producer, it works well. If I started a single consumer, the CPU usage is too high. Then the broker will be broken and the e2e latency would became too large that is up to 10+ seconds.

**Expected behavior**
The CPU usage should not be too high

**Screenshots**

The CPU usage:

![image](https://user-images.githubusercontent.com/18204803/118624902-25473f80-b7fc-11eb-8b90-db3be5c25bc2.png)

The thread that used the most CPU is `pulsar-client-io` thread. (this is a thread id of a previous test)

![image](https://user-images.githubusercontent.com/18204803/118622891-4d35a380-b7fa-11eb-84c0-ab717eb21255.png)

**Additional context**

The broker's logs:

```
08:15:07.887 [pulsar-io-28-6] INFO  io.streamnative.pulsar.handlers.kop.KafkaRequestHandler - channel active: [id: 0x88e88cb6, L:/10.0.0.53:9092 - R:/10.0.0.63:36180]
08:15:08.327 [pulsar-io-28-4] ERROR io.streamnative.pulsar.handlers.kop.KafkaCommandDecoder - [[id: 0x89671c81, L:/10.0.0.53:9092 ! R:/10.0.0.63:36178]] Failed to write RequestHeader(apiKey=OFFSET_COMMIT, apiVersion=4, clientId=consumer-sub-000-yYl5OxE-1, correlationId=537)
io.netty.channel.unix.Errors$NativeIoException: writevAddresses(..) failed: Broken pipe
```

It appears for 384 times and are all caused by `OFFSET_COMMIT` requests.

```
$ grep "Failed to write RequestHeader" pulsar-broker-ip-10-0-0-53.log| wc -l
     384
$ grep "Failed to write RequestHeader" pulsar-broker-ip-10-0-0-53.log | grep "OFFSET_COMMIT" | wc -l
     384
$ grep "NativeIoException" pulsar-broker-ip-10-0-0-53.log | wc -l
     386
```

[pulsar-broker-ip-10-0-0-53.log](https://github.com/streamnative/kop/files/6499885/pulsar-broker-ip-10-0-0-53.log)

[...]

PR #515, opened 2021-05-18
Use Subscription#resetCursor for cumulative ACK

BewareMyPower:
### Motivation

Currently KoP uses `OffsetAcker` to manage a list of Pulsar consumers for cumulative ACK. However, it wastes a lot of resources. For example, a topic with N partitions will create N consumers across all brokers. Besides, for each client side consumer, broker will create an associated broker side consumer (in package `org.apache.pulsar.broker.service`). In addition, the broker side consumer also updates two fields.

See https://github.com/streamnative/kop/issues/513, it could easily makes a network channel broken when `OFFSET_COMMIT` request are frequent.

What's worse is, when a bundle ownership changes, the consumers in `OffsetAcker` cannot detect the change. Because it needs to call `BrokerService#getTopic` to get the `PersistentTopic`, then find the offset's position by looking up in the `ManagedLedger`. However, even it failed, no error code will be sent to Kafka client.

### Modifications

- Use `Subscription#resetCursor` for cumulative ACK, for the first time when the cursor doesn't exist, create the `PersistentSubscription`.
- Only for those partitions that succeed in `OffsetAcker#ackOffset` will we write the data to `__consumer_offsets`.
- Remove  `CommitOffsetBacklogTest` because it relies on the update of `Subscription`.
- Revert https://github.com/streamnative/kop/pull/263, because it relies on the `OffsetAcker`'s internal consumers and the associated broker side consumers.


BewareMyPower:
@hangc0276 The CPU usage problem may be related to the openmessaging-benchmark's bug, the `OFFSET_COMMIT` requests are very often.

> 17:07:26.295 [bookkeeper-ml-workers-OrderedExecutor-3-0] INFO  org.apache.bookkeeper.mledger.impl.ManagedCursorImpl - [public/default/persistent/test-topic-4jMpA3M-0000-partition-14] Initiate reset position to 685:307 on cursor sub-000-keXE2c4
> 17:07:26.312 [bookkeeper-ml-workers-OrderedExecutor-3-0] INFO  org.apache.bookkeeper.mledger.impl.ManagedCursorImpl - [public/default/persistent/test-topic-4jMpA3M-0000-partition-14] Initiate reset position to 685:308 on cursor sub-000-keXE2c4
> 17:07:26.354 [bookkeeper-ml-workers-OrderedExecutor-3-0] INFO  org.apache.bookkeeper.mledger.impl.ManagedCursorImpl - [public/default/persistent/test-topic-4jMpA3M-0000-partition-14] Initiate reset position to 685:309 on cursor sub-000-keXE2c4
> 17:07:26.372 [bookkeeper-ml-workers-OrderedExecutor-3-0] INFO  org.apache.bookkeeper.mledger.impl.ManagedCursorImpl - [public/default/persistent/test-topic-4jMpA3M-0000-partition-14] Initiate reset position to 685:310 on cursor sub-000-keXE2c4

KoP has reset the cursor for about 33 to 47 times in every second. There may be something wrong with my benchmark client. (openmessaging-benchmark)

However, for the same workloads, after this PR, KoP has a significant better performance. See the following compare.

image

image

Before this PR, broker could be easily broken when lots of OFFSET_COMMIT requests arrived.

BewareMyPower:
@dockerzhang I removed the consumer side stats in this PR. Reusing broker side consumer's stats is convenient, however, KoP will manage N broker consumers and N client consumers for N topics/partitions. When N is too large, the performance could be impacted significantly.

I think we can add these metrics to prometheus metrics later.

wuzhanpeng:
I have a question, have you tried adjusting different commit frequencies to observe performance changes? Or in other words, how to determine that the performance loss is caused by changing the way of commit(ackOffset -> resetCursor) ?

BewareMyPower:
@wuzhanpeng I'm going to do this today because I just found the commit frequency is too last after this change (lots of INFO logs from pulsar side). BTW, for the previous test (https://github.com/streamnative/kop/issues/513), you can see all errors in `writeAndFlush`'s callback were related to `OFFSET_COMMIT`.

BewareMyPower:
There're 2 tests that may fail.

```
Error:    BasicEndToEndKafkaTest.testDeleteClosedTopics:98 null
Error:    KafkaRequestHandlerTest.testOffsetCommitRequestRetentionMs:603 » NoSuchElement
```

I'll try to fix them or remove them.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/OffsetAcker.java
```diff
@@ -13,213 +13,134 @@
...
+            if (ackResultMap.size() == numPartitions) {
+                future.complete(ackResultMap);
+            }
+        };
+
+        partitionAndOffsets.forEach((topicPartition, offsetAndMetadata) -> {
+            final String partitionTopicName = new KopTopic(topicPartition.topic())
+                    .getPartitionName(topicPartition.partition());
+            // 1. find PersistentTopic
+            brokerService.getTopicIfExists(partitionTopicName).whenComplete((optTopic, e) -> {
```
hangc0276, line 70: Why not use `admin.topics().resetCursor` api to commit offset, which can avoid topic ownership change.

hangc0276 (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/OffsetAcker.java
BewareMyPower: Because even if we use `PulsarAdmin` to reset cursor, we still need to `getTopic` in current broker so that we can find the `Position` of the offset.

BewareMyPower: Another reason is, when Kafka handles `OFFSET_COMMIT` request, it will also eventually call `ReplicaManager#appendRecords`. `ReplicaManager#getPartitionAndLeaderReplicaIfLocal` is responsible to check if the current broker is the leader of the partition. If not, a `NotLeaderForPartitionException` will be thrown and converted to the `NOT_LEADER_FOR_PARTITIONE` error code later.

hangc0276: OK

hangc0276 (approved):
LGTM

BewareMyPower:
We're going to remove `OffsetAcker`, so close this PR.

PR #520, opened 2021-05-20
expose consumer stats to prometheus instead of consumer internal stats

hangc0276:
### Motivation
In order to expose consumer stats, KOP create a consumer and update the stats metric to consumer object.  
It's too expensive.

### Modifications
1. expose consumer stats to prometheus instead of consumer internal stats

### TODO
1. groupid still should store in zookeeper, it's hard to remove.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -755,8 +755,6 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
 
     protected void handleProduceRequest(KafkaHeaderAndRequest produceHar,
                                         CompletableFuture<AbstractResponse> resultFuture) {
-        final long startProduceNanos = MathUtils.nowInNano();
-
         checkArgument(produceHar.getRequest() instanceof ProduceRequest);
         ProduceRequest produceRequest = (ProduceRequest) produceHar.getRequest();
 
@@ -781,8 +779,6 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
             if (log.isDebugEnabled()) {
                 log.debug("[{}] Request {}: Complete handle produce.", ctx.channel(), produceHar.toString());
             }
-            requestStats.getHandleProduceRequestStats()
-                    .registerFailedEvent(MathUtils.elapsedNanos(startProduceNanos), TimeUnit.NANOSECONDS);
             resultFuture.complete(new ProduceResponse(responseMap));
         };
         BiConsumer<TopicPartition, PartitionResponse> addPartitionResponse = (topicPartition, response) -> {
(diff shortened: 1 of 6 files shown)


BewareMyPower:
I'm afraid it's hard to reuse the current consumer stats design. There's a point that if we're going to remove `OffsetAcker`, the consumers should be created somewhere else.

Or should we just remove `OffsetAcker#ackOffsets` only? Currently the consumers of `OffsetAcker` are created in

- `GroupCoordinator#handleSyncGroup`
    - `OffsetAcker#addOffsetsTracker`
        - `OffsetAcker#getOrCreateConsumer`

Then we'll still create client side consumers and broker side consumers.

BewareMyPower:
Sorry I missed something. I found `getGroupConsumers` is removed, the there's no problem with this PR.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -1286,17 +1286,6 @@ protected void handleOffsetCommitRequest(KafkaHeaderAndRequest offsetCommit,
...
-        long startFetchingRequestNanos = MathUtils.nowInNano();
-        resultFuture.whenComplete((r, e) -> {
-            if (e != null) {
-                requestStats.getHandleFetchRequestStats().registerFailedEvent(
-                        MathUtils.elapsedNanos(startFetchingRequestNanos), TimeUnit.NANOSECONDS);
-            } else {
-                requestStats.getHandleFetchRequestStats().registerSuccessfulEvent(
-                        MathUtils.elapsedNanos(startFetchingRequestNanos), TimeUnit.NANOSECONDS);
-            }
-        });
```
BewareMyPower, line 1298: It's removed because https://github.com/streamnative/kop/pull/514 has replaced it with `REQUEST_LATENCY`, right? And the `handleFetchRequestStats` field of `RequestStats` should be removed as well. Also `HANDLE_PRODUCE_REQUEST` could be removed too.

In addition, I found the `testMetricsProvider` still passed. I found the metrics string contains `NaNkop_server_HANDLE_FETCH_REQUEST`. Should we improve the tests to detect some unused metrics related fields?

BewareMyPower (changes requested):
I found some metrics are outdated,

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
hangc0276: Yes, #514 has introduced all apis latency metric, so we should remove the above one. Some field should also be removed, it will update the code.

The test still passed is due to the metric has been initialized, and just not update the value.

BewareMyPower (approved):

BewareMyPower:
I found FETCH latency metrics doesn't exist

```
kop_server_REQUEST_LATENCY{success="false",quantile="0.99", request="Fetch"} NaN
```

Could you take a look?

hangc0276:
> I found FETCH latency metrics doesn't exist
> 
> ```
> kop_server_REQUEST_LATENCY{success="false",quantile="0.99", request="Fetch"} NaN
> ```
> 
> Could you take a look?

@BewareMyPower  I have address the problem, just due to metric update delay, PTAL.

Issue #392, opened 2021-03-03
[BUG] Topic not served by this instance Exception when consume topic


hangc0276:
**Describe the bug**
When consume topic message using kafka client, the kop throws `topic not served by this instance` exception. The Exception as follow:
```
14:55:10.789 [ForkJoinPool.commonPool-worker-53] WARN  org.apache.pulsar.broker.service.BrokerService - Namespace bundle for topic (persistent://public/default/chenhang_kop_perf_consume_v2-partition-0) not served by this instance. Please redo the lookup. Request is denied: namespace=public/default
14:55:10.789 [ForkJoinPool.commonPool-worker-53] ERROR io.streamnative.pulsar.handlers.kop.coordinator.group.OffsetAcker - [persistent://public/default/chenhang_kop_perf_consume_v2-partition-0] get topic failed when ack for kop_perf_group_v1.
org.apache.pulsar.broker.service.BrokerServiceException$ServiceUnitNotReadyException: Namespace bundle for topic (persistent://public/default/chenhang_kop_perf_consume_v2-partition-0) not served by this instance. Please redo the lookup. Request is denied: namespace=public/default
        at org.apache.pulsar.broker.service.BrokerService.lambda$checkTopicNsOwnershipAsync$53(BrokerService.java:1515) ~[org.apache.pulsar-pulsar-broker-2.8.0-rc-202101252233.jar:2.8.0-rc-202101252233]
        at java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:859) ~[?:?]
        at java.util.concurrent.CompletableFuture.uniWhenCompleteStage(CompletableFuture.java:883) ~[?:?]
        at java.util.concurrent.CompletableFuture.whenComplete(CompletableFuture.java:2251) ~[?:?]
        at org.apache.pulsar.broker.service.BrokerService.checkTopicNsOwnershipAsync(BrokerService.java:1507) ~[org.apache.pulsar-pulsar-broker-2.8.0-rc-202101252233.jar:2.8.0-rc-202101252233]
        at org.apache.pulsar.broker.service.BrokerService.loadOrCreatePersistentTopic(BrokerService.java:1050) ~[org.apache.pulsar-pulsar-broker-2.8.0-rc-202101252233.jar:2.8.0-rc-202101252233]
[...]

PR #521, opened 2021-05-20
Remove OffsetAcker

BewareMyPower:
Fixes https://github.com/streamnative/kop/issues/392

### Motivation

https://github.com/streamnative/kop/pull/105 introduced the `OffsetAcker` to reuse the backlog stat of Pulsar subscription. However, after https://github.com/streamnative/kop/pull/296 introduced the continuous offset, when the `OffsetAcker` want to find the `MessageId` or `Position`, which contains the ledger id and entry id of a batch, it needs to get the `PersistentTopic` using `BrokerService#getTopic`. If the current broker was not the owner broker of the partition, `getTopic` would fail.

However, Kafka consumers send the `OFFSET_COMMIT` requests to the coordinator broker, which is associated with the specified group name. A coordinator broker may not own the partition after the pulsar's topic ownership changed, so the `getTopic` would fail and couldn't be recovered, which affect the performance significantly and a lot of warnings logs would be printed. 

### Modifications

- Remove the `OffsetAcker` with the related metrics (`CoordinatorStats`) and tests (`CommitOffsetBacklogTest`)
- Remove the reused consumer stats from https://github.com/streamnative/kop/pull/263 with the related tests (`testKafkaConsumerMetrics`)
- Add `testMixedConsumersWithSameSubscription` to cover the case that Kafka consumer and Pulsar consumer subscribes the same topic with same subscription name. Because before this PR, Pulsar consumer were aware of the Kafka consumer's consumed position (committed offset).
- Fix the current tests:
    - `testDeleteTopics`: After this PR, the topic could be deleted even the Kafka consumers were still connected because no broker side Consumer were attached to the topic.
    - `testPulsarProduceKafkaConsume2`: It's a test for Kafka consumer with `enable.auto.commit=true`, however the validation is wrong because even a consumer hasn't committed any offset, the consumer can still consume from the latest offset.

Regarding to the cases that Kafka or Pulsar producer produces and Kafka or Pulsar consumer consumes, they are covered by existed `KafkaRequestTypeTest`. And the offset commit case is covered by `DifferentNamespaceTestBase#testCommitOffset`.


BewareMyPower:
@hangc0276 FYI, this PR retains the ZooKeeper utils and add the TODO  (see `MessageFetchContext#readMessagesInternal`) for https://github.com/streamnative/kop/pull/520.

PTAL and merge this PR ASAP if there's not any problem. Then you can rebase your PR.

BTW, when you merge this PR, you need to copy the PR description manually because the automatically generated commit message only contains the description of all single commits.

hangc0276 (approved):
LGTM

PR #522, opened 2021-05-20
Apply getPartitionedTopicMetadata API with correct semantics

BewareMyPower:
Upgrade pulsar dependency to a newer version (2.8.0-rc-202105182205) that contains https://github.com/apache/pulsar/pull/10601, which corrected the wrong semantics of `getPartitionedTopicMetadataAsync` that returns 0 for a non-existed topic.

### DIFF
--- a/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandlerTest.java
+++ b/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandlerTest.java
@@ -349,8 +349,7 @@ public class KafkaRequestHandlerTest extends KopProtocolHandlerTestBase {
     }
 
 
-    //@Test(timeOut = 10000)
-    @Test
+    @Test(timeOut = 10000)
     public void testCreateAndDeleteTopics() throws Exception {
         Properties props = new Properties();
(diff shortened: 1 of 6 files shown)


tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandlerTest.java
```diff
@@ -164,7 +173,7 @@ protected void setup() throws Exception {
         handler.ctx = mockCtx;
     }
 
-    @AfterMethod
+    @AfterClass
```
hangc0276, line 176: Why change to `AfterClass`

hangc0276 (changes requested):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandlerTest.java
BewareMyPower: To reduce the test time because it took a lot of time in my local env. All methods use different topics and won't affect each other.

hangc0276: It should avoid using the same topic between tests, which is dangerous. The following developers may fall into this trap.

hangc0276 (approved):

dockerzhang (approved):

PR #526, opened 2021-05-25
[docs] Prepare for sync KoP content to Hub

Jennifer88huang:

PR #527, opened 2021-05-26
Update README.md

Huanli-Meng:


BewareMyPower:
Could you give your suggestions in https://github.com/streamnative/kop/pull/526 because it may be conflict with this PR?

Huanli-Meng:
> Could you give your suggestions in #526 because it may be conflict with this PR?

It seems that in PR #526, the most contents in `README.md` file are moved to a `kop.md` file to realize automatic sync contents between docs in KoP repo and Hub doc. Therefore, my suggestion is:
if the modification is acceptable, it should be updated to the `kop.md` file. Or, i can close this PR to ensure that the required KoP readme contents are synched to Hub doc. What do you think?

Huanli-Meng:
as most contents are moved to `kop.md` file in #526, therefore, i will close this PR.

PR #531, opened 2021-05-28
enable offset out-of-range checking

### LINKED ISSUE #530: [BUG] Cannot consume any message after deleting and re-creating a topic with the same name
**Describe the bug**
When I deleted an old topic and re-created it with the same configurations, I found that I can not consume anything after I produced some messages into it.

**To Reproduce**
1. Create topic-A
2. Send 10 messages to topic-A
3. Consume all the messages within topic-A
4. Delete topic-A
5. Re-create topic-A with the same configurations
6. Send 5 messages to topic-A
7. Start to consume from topic-A but get nothing

**Expected behavior**
Handle offset out-of-range exception for client properly.

**Screenshots**
When I described a groupid in this time, I can see:
image

**Additional context**
I will submit a PR for it later.

wuzhanpeng:
Resolved #530 

This PR implements the offset out-of-range checking and handles the exceptions for kafka-client.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
@@ -51,6 +51,7 @@ import org.apache.bookkeeper.mledger.AsyncCallbacks.ReadEntriesCallback;
 import org.apache.bookkeeper.mledger.Entry;
 import org.apache.bookkeeper.mledger.ManagedCursor;
 import org.apache.bookkeeper.mledger.ManagedLedgerException;
+import org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl;
 import org.apache.bookkeeper.mledger.impl.NonDurableCursorImpl;
 import org.apache.bookkeeper.mledger.impl.PositionImpl;
 import org.apache.bookkeeper.stats.OpStatsLogger;
@@ -153,6 +154,37 @@ public final class MessageFetchContext {
 
                             long offset = ((FetchRequest) fetchRequest.getRequest()).fetchData()
                                 .get(topicPartition).fetchOffset;
+
+                            // handle offset out-of-range exception
+                            ManagedLedgerImpl managedLedger = (ManagedLedgerImpl) tcm.getManagedLedger();
+                            try {
+                                long logStartOffset = MessageIdUtils.getLogStartOffset(managedLedger);
+                                long logEndOffset = MessageIdUtils.getLogEndOffset(managedLedger);
+                                if (offset > logEndOffset || offset < logStartOffset) {
+                                    log.error("Received request for offset {} for partition {}, "
+                                                    + "but we only have entries in the range {} to {}.",
+                                            offset, topicPartition, logStartOffset, logEndOffset);
+                                    responseData.put(topicPartition, new FetchResponse.PartitionData<>(
+                                            Errors.OFFSET_OUT_OF_RANGE,
+                                            FetchResponse.INVALID_HIGHWATERMARK,
+                                            FetchResponse.INVALID_LAST_STABLE_OFFSET,
+                                            FetchResponse.INVALID_LOG_START_OFFSET,
+                                            null,
+                                            MemoryRecords.EMPTY));
+                                    return null;
+                                }
+                            } catch (Exception e) {
+                                log.error("[{}] Failed to get log start offset for managed ledger {}",
+                                        topicPartition, managedLedger.getName(), e);
+                                responseData.put(topicPartition, new FetchResponse.PartitionData<>(
+                                        Errors.UNKNOWN_SERVER_ERROR,
+                                        FetchResponse.INVALID_HIGHWATERMARK,
+                                        FetchResponse.INVALID_LAST_STABLE_OFFSET,
+                                        FetchResponse.INVALID_LOG_START_OFFSET,
+                                        null,
+                                        MemoryRecords.EMPTY));
+                                return null;
+                            }
 
                             if (log.isDebugEnabled()) {
                                 log.debug("Fetch for {}: remove tcm to get cursor for fetch offset: {} .",
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MessageIdUtils.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MessageIdUtils.java
@@ -34,12 +34,6 @@ public class MessageIdUtils {
         return ((ManagedLedgerInterceptorImpl) managedLedger.getManagedLedgerInterceptor()).getIndex();
     }
 
-    public static long getLogStartOffset(ManagedLedgerImpl managedLedger) throws Exception {
-        long firstLedgerId = managedLedger.getLedgersInfo().firstKey();
-        return getOffsetOfPosition(managedLedger,
-                new PositionImpl(firstLedgerId, 0), false, -1L).get();
-    }
-
     public static long getHighWatermark(ManagedLedger managedLedger) {
         return getCurrentOffset(managedLedger) + 1;
     }
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MessageIdUtils.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MessageIdUtils.java
@@ -38,6 +38,12 @@ public class MessageIdUtils {
         return ((ManagedLedgerInterceptorImpl) managedLedger.getManagedLedgerInterceptor()).getIndex();
     }
 
+    public static long getLogStartOffset(ManagedLedgerImpl managedLedger) throws Exception {
+        long firstLedgerId = managedLedger.getLedgersInfo().firstKey();
+        return getOffsetOfPosition(managedLedger,
+                new PositionImpl(firstLedgerId, 0), false, -1L).get();
+    }
+
     public static long getHighWatermark(ManagedLedger managedLedger) {
         return getCurrentOffset(managedLedger) + 1;
     }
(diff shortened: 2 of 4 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
```diff
@@ -154,6 +155,37 @@ public void recycle() {
                             long offset = ((FetchRequest) fetchRequest.getRequest()).fetchData()
                                 .get(topicPartition).fetchOffset;
 
+                            // handle offset out-of-range exception
+                            ManagedLedgerImpl managedLedger = (ManagedLedgerImpl) tcm.getManagedLedger();
+                            try {
+                                long logStartOffset = MessageIdUtils.getLogStartOffset(managedLedger);
```
hangc0276, line 161: `getLogStartOffset` will read entry from BookKeeper, which is time consuming, and it will called by each handle fetch request. It is too expensive.  In my opinion, after get exception, we try to call `getLogStartOffset` maybe better.

hangc0276 (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
BewareMyPower: I agree with @hangc0276 . I think we can only check `offset > logEndOffset` here.

And could you give an example to show in which case will the offset be less than `logStartOffset`? If it's hard to simulate this case, could you explain what happens if `offset < logStartOffset`? If it there will be an exception in `ManagedCursor#asyncReadEntries`, I think we can just add the check here.

wuzhanpeng: For example I have a flink job using KoP and I trigger a savepoint `S` at offset `A` and then stop it. After a period of time, the offset becomes to range `[A+N, A+M]` satisfying `N,M>0, M>N` due to topic retention. At this time the user chooses to recover its job from savepoint `S`, which will lead a out-of-range that meets `offset < logStartOffset`. Here is also a reference to the implementation of Kafka.

wuzhanpeng: Maybe we can utilize the short-circuit rules in logic operations to make some optimizations? 🤣 
```java
...
if (offset > logEndOffset || offset < MessageIdUtils.getLogStartOffset(managedLedger)) {...}
...
```

hangc0276: if offset is between LSO and LEO range, it still go into `getLogStartOffset`

BewareMyPower: So could you answer the question that what happens if `offset < logStartOffset`? If it's possible, could you add a test to cover this case?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
```diff
@@ -154,6 +155,37 @@ public void recycle() {
...
+                                long logEndOffset = MessageIdUtils.getLogEndOffset(managedLedger);
+                                if (offset > logEndOffset || offset < logStartOffset) {
+                                    log.error("Received request for offset {} for partition {}, "
+                                                    + "but we only have entries in the range {} to {}.",
+                                            offset, topicPartition, logStartOffset, logEndOffset);
+                                    responseData.put(topicPartition, new FetchResponse.PartitionData<>(
```
aloyszhang, line 167: If just retrurn an error response, how can kafka consumer fetch new messages?  
Assumes that the offset in Fetch request is less than LSO, can the consuemr or broker  move the offset to the LSO? If not, does the consume will block at the invilad offset?

aloyszhang (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
aloyszhang: Agree with that `getLogStartOffset` is an expensive operation since it will goes into bookie.
An optimization for  reducing the cost of `getLogStartOffset`  is  adding a `startOffset` field  for metadata of ManagedLedger in broker side and then we can get the `startOffse` from ManagedLedger directly without reading entry from bookie.

BewareMyPower: @aloyszhang It's okay. But it should be merged into pulsar repo ASAP because 2.8.0 is in releasing process now.

hangc0276: @aloyszhang  We have another method to achieve this goal.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
hangc0276: Kafka consumer will deal with out of range exception.

wuzhanpeng:
I supplemented a testcase for `offset < logStartOffset`. @BewareMyPower

BewareMyPower:
I changed your code to

```java
                            try {
                                long logEndOffset = MessageIdUtils.getLogEndOffset(managedLedger);
                                if (offset > logEndOffset) {
```

`testLessThanStartOffset` still passed.

wuzhanpeng:
> I changed your code to
> 
> ```java
>                             try {
>                                 long logEndOffset = MessageIdUtils.getLogEndOffset(managedLedger);
>                                 if (offset > logEndOffset) {
> ```
> 
> `testLessThanStartOffset` still passed.

This is because when you assign a offset which is less than earliest position to `KafkaTopicConsumerManager#getCursorByOffset`, you will get a earliest position by default.

BewareMyPower:
So is the check for log start offset necessary?

wuzhanpeng:
@BewareMyPower In my opinion, if we set `auto.offset.reset=latest`(also the default value) in kafka-client, when we fetch data with a offset that less than earliest position, the reading position is forced to reset to earliest instead of the latest. This will cause semantic inconsistency.

We may need to weigh which is the most suitable solution.

BewareMyPower:
So could you move the log start offset check to `handleListOffset` to get the accurate offset?

wuzhanpeng:
> So could you move the log start offset check to `handleListOffset` to get the accurate offset?

Why do we need to move the checker to `handleListOffset`? It seems that `handleListOffset` has already handle the cases about `timestamp == EARLIEST_TIMESTAMP` and `timestamp == LATEST_TIMESTAMP`.

BewareMyPower:
I mean if we already ensure the offset that is returned to consumer (in LIST_OFFSET response) is less than log start offset, then we could remove the check in FETCH handler. Because FETCH requests are much more than LIST_OFFSET requests.

wuzhanpeng:
You can see my example disscussed above(https://github.com/streamnative/kop/pull/531#discussion_r641874800). If we recovery the offset from the external system, like a flink `savepoint`, we can not completely avoid this situation for the time being.

hangc0276:
This PR should also be resolved before release. The LSO can't be calculated by `LEO - numberOfEntries`, because LEO just mark the number of messages. 

One possible solution is,
When get position by offset, we should check whether the got position is the start of the manangedLedger, If so, we should double check whether the got position's offset is the request fetching offset. If the double check not pass, we should return `out-of-range` exception to client.

wuzhanpeng:
@BewareMyPower Or maybe we can deal with the case of `offset > LogEndOffset` first and leave `offset < LogStartOffset` as a todo comment to handle later. What do you think?

BewareMyPower:
> @BewareMyPower Or maybe we can deal with the case of `offset > LogEndOffset` first and leave `offset < LogStartOffset` as a todo comment to handle later. What do you think?

Yeah, I agree.

hangc0276:
> > @BewareMyPower Or maybe we can deal with the case of `offset > LogEndOffset` first and leave `offset < LogStartOffset` as a todo comment to handle later. What do you think?
> 
> Yeah, I agree.

We'd better give a non-effective solution just for `offset < Log start offset` case, otherwise, user who use KoP may trigger this bug may cause consumer stuck.

BewareMyPower:
> > > @BewareMyPower Or maybe we can deal with the case of `offset > LogEndOffset` first and leave `offset < LogStartOffset` as a todo comment to handle later. What do you think?
> > 
> > 
> > Yeah, I agree.
> 
> We'd better give a non-effective solution just for `offset < Log start offset` case, otherwise, user who use KoP may trigger this bug may cause consumer stuck.

I think we need to estimate how much it will affect the performance. If not much, that's okay. Otherwise, we should not take the risk to affect the more common cases.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MessageIdUtils.java
```diff
@@ -34,6 +34,12 @@ public static long getCurrentOffset(ManagedLedger managedLedger) {
         return ((ManagedLedgerInterceptorImpl) managedLedger.getManagedLedgerInterceptor()).getIndex();
     }
 
+    public static long getLogStartOffset(ManagedLedgerImpl managedLedger) throws Exception {
+        long firstLedgerId = managedLedger.getLedgersInfo().firstKey();
+        return getOffsetOfPosition(managedLedger,
+                new PositionImpl(firstLedgerId, 0), false, -1L).get();
+    }
+
```
BewareMyPower, line 42: Remove this method because it's never used.

BewareMyPower (changes requested):

BewareMyPower:
@hangc0276 I also noticed the current `getLogStartOffset` method is a synchronous implementation. It's very dangerous that may cause problems like https://github.com/streamnative/kop/issues/544

BewareMyPower (approved):

hangc0276 (approved):
LGTM

hangc0276:
> @hangc0276 I also noticed the current `getLogStartOffset` method is a synchronous implementation. It's very dangerous that may cause problems like #544

@BewareMyPower yes, it's dangerous. let's merge the PR first, and continue to consider how to resolve `offset < log start offset` case

Issue #418, opened 2021-03-28
Add docs for the prometheus metrics


jiazhai:
Regarding how to enable the monitor for prometheus, we can open a new PR to add the docs.

It would also need to add related docs for the meaning each metrics

_Originally posted by @BewareMyPower in https://github.com/streamnative/kop/issues/389#issuecomment-808848839_

PR #536, opened 2021-05-31
add producer stats

hangc0276:
### Motivation
For producer bytesIn and messageIn metric, the current implementation is reuse Pulsar metric, which mix topic name with partition,  and it's hard to add PromQL for grafana query and alert.

### Modification
1. use individual producer metric for kop and devide topic name and partition into two label
2. add `batch count per memory recods` gauge metric to monitor batch count per memory records, due to current KoP version doesn't support multi batches in one memory recods.


dockerzhang (approved):
LGTM

hangc0276:
resolve #418

BewareMyPower (approved):

PR #538, opened 2021-05-31
Refactor FETCH request handler to reduce GC

BewareMyPower:
### Motivation

When a consumer sends FETCH requests frequently, for example, polling in a tight loop,

```java
while (true) {
    ConsumerRecords<String, byte[]> records = consumer.poll(Duration.ofMillis(100));
    // process records quickly
}
```
the heap memory increasing quickly and GC will happen frequently about once per minute. The frequency is not related to the throughput but the count of FETCH requests.

It's because each time a FETCH request arrived, KoP will create a `MessageFetchContext`. Though it's allocated in a Netty recycler, there're a lot of small objects that are still allocated from heap memory. There're a lot of wait all operations, for example, after all `PersistentTopic`s of each partition were get, it collected all those `PersistentTopic`s to a list and retrieve partition's associated data from a `Map`.

We should process each partition in parallel and avoid collecting the middle result to a list and map.

### Modifications

Refactor the `MessageFetchContext`, only collect the results to a list after the final `PartitionData` of each partition was done.

Besides, this PR removes some implementation details.
1. Checks for `minBytes` and `maxBytes`. Because the existed check is wrong. The default `minBytes` is 1 so `allSize > minBytes || allSize > maxBytes` is always true.  The `maxBytes` check is meaningless.
2. This PR only tries to read `maxReadEntriesNum` entries for only once. It's the same reason as previous.
3. Remove the `TOTAL_MESSAGE_READ` metrics because we don't wait all partitions' messages read anymore.
4. Remove the handle of FETCH timeout process, because it was still wrong before that it didn't count for `asyncReadEntries`, which is more likely to be blocked.

For the read bytes limit like `max.partition.fetch.bytes` and `fetch.max.bytes`, we should open a new PR to implement it correctly.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -1305,8 +1305,7 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
             });
         }
 
-        MessageFetchContext fetchContext = MessageFetchContext.get(this, requestStats);
-        fetchContext.handleFetch(resultFuture, fetch, transactionCoordinator, fetchPurgatory);
+        MessageFetchContext.get(this, fetch, resultFuture).handleFetch();
     }
 
     protected void handleJoinGroupRequest(KafkaHeaderAndRequest joinGroup,
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopServerStats.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopServerStats.java
@@ -58,7 +58,6 @@ public interface KopServerStats {
      * </p>
      */
     String PREPARE_METADATA = "PREPARE_METADATA";
-    String TOTAL_MESSAGE_READ = "TOTAL_MESSAGE_READ";
     String MESSAGE_READ = "MESSAGE_READ";
     String FETCH_DECODE = "FETCH_DECODE";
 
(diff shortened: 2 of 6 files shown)


BewareMyPower:
Here's the test compare, after this PR, the heap memory increase speed is slower than before.

Before PR 538:
Before PR 538

GC logs:
```
15:25:32 [187.363s][info][gc] GC(6) Pause Young (Normal) (G1 Evacuation Pause) 4114M->113M(8192M) 34.331ms
15:27:03 [278.498s][info][gc] GC(7) Pause Young (Normal) (G1 Evacuation Pause) 4113M->112M(8192M) 33.489ms
15:28:34 [369.634s][info][gc] GC(8) Pause Young (Normal) (G1 Evacuation Pause) 4112M->113M(8192M) 35.152ms
```

After PR 538:
After PR 538

GC logs:
```
15:36:23 [179.284s][info][gc] GC(6) Pause Young (Normal) (G1 Evacuation Pause) 4113M->112M(8192M) 34.538ms
15:38:30 [306.100s][info][gc] GC(7) Pause Young (Normal) (G1 Evacuation Pause) 4112M->112M(8192M) 35.586ms
```

> NOTE: the GC metrics seems not accurate, so I pasted the GC logs for each graph.

We can see the GC interval was longer than before, about 2 seconds vs 1.5 seconds.

Since even after this PR, for each FETCH request (and PRODUCE request), there're still some small objects that are allocated from heap like

```
private Map<TopicPartition, PartitionData<MemoryRecords>> responseData;
```

In future, we should also use Netty recycler to allocate these little objects.

jiazhai (approved):

wuzhanpeng:
Nice job. It seems that the allocation and release time of a cycle has changed from 90s to 120s. Can you simply share your test method? I want to see how high frequency of a fetch request will cause such situation.

BewareMyPower:
@wuzhanpeng I used openmessaging-benchmark in my own branch https://github.com/BewareMyPower/openmessaging-benchmark/commits/bewaremypower/deploy-kop

But it's simple to simulate the case. Just run a producer and a consumer in different machines.

dockerzhang (approved):
great improvement

PR #554, opened 2021-06-07
Handle group name was failed to be written to ZK gracefully

BewareMyPower:
### Motivation

Currently KoP writes group name and the associated client's host and id to ZK so that consumer stats could be updated while FETCH request contains no group info. However, there is something wrong with ZooKeeperUtils.

First, when the `groupIdStoredPath` already existed, the write to ZK may fail with `NodeExists` because of race condition.

Second, if the data was failed to be written to ZK, every time a FETCH request comes, a ZK read operation will be performed. It's because `ZooKeeperUtils#getData` returns null if the node doesn't exist and a null value cannot be put into a `Map`. In this case, `ConcurrentHashMap#computeIfAbsent` will always execute the second argument to create a value.

### Modifications

- Create groupIdZooKeeperPath only when KoP starts, and only log the error message if ZK write failed.
- When `ZooKeeperUtils#getData` failed, returns an empty string instead of null so that if the group name was failed to be written to ZK, the empty returned value would be cached.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/ZooKeeperUtils.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/ZooKeeperUtils.java
@@ -29,9 +29,12 @@ import org.apache.zookeeper.data.Stat;
 @Slf4j
 public class ZooKeeperUtils {
 
-    private static boolean tryCreatePath(ZooKeeper zooKeeper, String path, byte[] data) {
+    public static boolean tryCreatePath(ZooKeeper zooKeeper, String path, byte[] data) {
         try {
             zooKeeper.create(path, data, ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.PERSISTENT);
+            if (log.isDebugEnabled()) {
+                log.debug("Created ZK path: {} data: {}", path, new String(data, StandardCharsets.UTF_8));
+            }
         } catch (KeeperException e) {
             if (!e.code().equals(KeeperException.Code.NODEEXISTS)) {
                 log.error("Failed to create ZooKeeper node {}: {}", path, e.getMessage());
@@ -42,16 +45,6 @@ public class ZooKeeperUtils {
             return false;
         }
         return true;
-    }
-
-    public static void createPath(ZooKeeper zooKeeper, String zkPath, String subPath, byte[] data) {
-        tryCreatePath(zooKeeper, zkPath, new byte[0]);
-        final String addSubPath = zkPath + subPath;
-        final boolean result = tryCreatePath(zooKeeper, addSubPath, data);
-        if (result && log.isDebugEnabled()) {
-            log.debug("create zk path, addSubPath:{} data:{}.",
-                    addSubPath, new String(data, StandardCharsets.UTF_8));
-        }
     }
 
     public static @NonNull String getData(ZooKeeper zooKeeper, String zkPath, String subPath) {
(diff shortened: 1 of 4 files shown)


BewareMyPower:
Here're two logs screenshot from my test environment.

image

image

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/ZooKeeperUtils.java
```diff
@@ -26,38 +28,43 @@
...
             }
+        } catch (InterruptedException e) {
+            log.error("Failed to create ZooKeeper node {}: {}", path, e.getMessage());
+            return false;
+        }
+        return true;
+    }
+
+    public static void createPath(ZooKeeper zooKeeper, String zkPath, String subPath, byte[] data) {
+        tryCreatePath(zooKeeper, zkPath, new byte[0]);
```
aloyszhang, line 48: I think we can create znode for `groupIdStoredPath` at the time of KoP started instead of called every time.

aloyszhang (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/ZooKeeperUtils.java
BewareMyPower: Done

aloyszhang:
LGTM

aloyszhang (approved):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/ZooKeeperUtils.java
```diff
@@ -26,42 +28,43 @@
...
+            if (log.isDebugEnabled()) {
+                log.debug("Created ZK path: {} data: {}", path, new String(data, StandardCharsets.UTF_8));
             }
-            String addSubPath = zkPath + subPath;
-            if (zooKeeper.exists(addSubPath, false) == null) {
-                zooKeeper.create(addSubPath,
-                        data, ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.PERSISTENT);
-            } else {
-                zooKeeper.setData(addSubPath, data, -1);
+        } catch (KeeperException e) {
```
hangc0276, line 38: use 
```
 catch (KeeperException | InterruptedException e ) {
     // ...
 }
```
may better

hangc0276 (changes requested):

dockerzhang (approved):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/ZooKeeperUtils.java
BewareMyPower: How to check the error code if these two exceptions are merged?

```java
        } catch (KeeperException | InterruptedException e) {
            if (e instanceof KeeperException && ((KeeperException) e).code().equals(KeeperException.Code.NODEEXISTS)) {
                return true;
            } else {
                log.error("Failed to create ZooKeeper node {}: {}", path, e.getMessage());
                return false;
            }
        }
```

IMO, the code logic will be more complicated.

BewareMyPower: OK, just now I found the returned value is ignored. So I won't check the code of KeeperException and merge these exceptions.

BewareMyPower: Done.

BewareMyPower:
@hangc0276 PTAL again

hangc0276 (approved):

Issue #544, opened 2021-06-02
[BUG] KoP cannot support a FETCH request with many partitions


BewareMyPower:
**Describe the bug**
When I run performance tests on a topic with 100 partitions, the topic will be not available quickly. What's worse is the whole KoP cluster will be broken because in this occasion, no bookie will be available. So even a new topic was created after that, it would still be unavailable because `Not enough non-faulty bookies available`.

> 15:07:19.134 [BookKeeperClientWorker-OrderedExecutor-0-0] WARN  org.apache.bookkeeper.client.PendingAddOp - Failed to write entry (177, 15): Bookie operation timeout
> 15:07:19.143 [BookKeeperClientWorker-OrderedExecutor-0-0] WARN  org.apache.bookkeeper.client.RackawareEnsemblePlacementPolicyImpl - Failed to find 1 bookies : excludeBookies [<Bookie:10.0.0.220:3181>, <Bookie:10.0.0.198:3181>, <Bookie:10.0.0.228:3181>], allBookies [<Bookie:10.0.0.228:3181>, <Bookie:10.0.0.198:3181>, <Bookie:10.0.0.220:3181>].
> 15:07:19.143 [BookKeeperClientWorker-OrderedExecutor-0-0] WARN  org.apache.bookkeeper.client.RackawareEnsemblePlacementPolicyImpl - Failed to choose a bookie: excluded [<Bookie:10.0.0.220:3181>, <Bookie:10.0.0.198:3181>, <Bookie:10.0.0.228:3181>], fallback to choose bookie randomly from the cluster.
> <repeated logs>...
> 15:07:19.143 [BookKeeperClientWorker-OrderedExecutor-0-0] ERROR org.apache.bookkeeper.client.MetadataUpdateLoop - UpdateLoop(ledgerId=177,loopId=4ef321e8) Exception updating
  1828 org.apache.bookkeeper.client.BKException$BKNotEnoughBookiesException: Not enough non-faulty bookies available

The root cause is `Bookie operation timeout`. It seems to lead to the result that all bookies are excluded.

After that, producer might still be enable to write some messages, but there could be a possibility that corrupted message could be written to BK, see https://github.com/streamnative/kop/issues/442 for reference. And then a direct memory leak will happen as well. https://github.com/streamnative/kop/pull/542 is to handle this corrupted messages.

When there's no connected consumer, this case won't happen.

**To Reproduce**
Deploy a KoP cluster, create 
[...]

PR #556, opened 2021-06-08
Fix deadlock caused by synchronous asyncFindPosition

BewareMyPower:
Fixes https://github.com/streamnative/kop/issues/544

### Motivation

After the debugging, when a FETCH request contains many partitions, for each partition, `KafkaTopicConsumerManager#remove` will be executed in the same `pulsar-io-<suffix>` thread, which will eventually call `MessageIdUtils#getPositionForOffset` to get `PositionImpl` of the offset.

However, `getPositionForOffset` is a synchronous method that waits until a `CompletableFuture` is done. The future is returned by `ManagedLedgerImpl#asyncFindPosition` that calls `ManagedLedgerImpl#asyncReadEntry` internally. If the producer is publishing messages at the same time, the `asyncAddEntry` method could be called for the same `ManagedLedgerImpl` object. It will somehow cause the deadlock.
 
### Modifications
- Just use `ManagedLedgerImpl#asyncFindPosition` instead of the synchronous call in `MessageIdUtils#getPositionForOffset`.
- Caches the future of `Pair<ManagedCursor, Long>` in KafkaTopicConsumerManager instead of the `Pair<ManagedCursor, Long>`. Then in `MessageFetchContext`, use the `ManagedCursor` in future's callback .
- Add a unit test to ensure after this change, the non-durable cursor will be still created once for each consumer. This condition is easily to break in a refactor like this PR and it may cause performance problem because cursor might be created each time a FETCH request arrived. So here I added a test to avoid the case.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicConsumerManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicConsumerManager.java
@@ -15,9 +15,10 @@ package io.streamnative.pulsar.handlers.kop;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
-import io.streamnative.pulsar.handlers.kop.utils.MessageIdUtils;
+import io.streamnative.pulsar.handlers.kop.utils.OffsetSearchPredicate;
 import java.io.Closeable;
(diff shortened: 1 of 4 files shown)


jiazhai (approved):

hangc0276 (approved):
good job

PR #569, opened 2021-06-14
[Docs] Add more details for SSL keys

BewareMyPower:
Currently the docs related to SSL keys is not friendly to users that are not familiar with `keytool` and `openssl` commands. For some commands we need to specify the password, while for some other commands we need to input the password we have specified before.

In addition, this PR removes the redundant blanks of the line tail.


docs/security.md
```diff
@@ -276,33 +276,46 @@ The following example shows how to connect KoP through SSL.
...
     keytool -keystore server.keystore.jks -alias CARoot -import -file ca-cert
+    # You must input the password of server.keystore.jks: "server-keystore"
     keytool -keystore server.keystore.jks -alias localhost -import -file cert-signed
     ```
 
+    In above example, we have input four passwords:
+
+    - `server-keystore` for `server.keystore.jks` file
+    - `server` for `ca-cert` and `ca-key` files
+    - `server-truststore` for `client.truststore.jks` file
```
gaoran10, line 302:
```suggestion
    - `server-truststore` for `server.truststore.jks` file
```

jiazhai (approved):

PR #585, opened 2021-06-18
unify the topic name in __consumer_offsets

### LINKED ISSUE #584: [BUG] Topic name is not unified in __consumer_offsets
**Describe the bug**
I find that using different topic name(but acutully they are the same, such as `persistent://public/default/test` and `test`) to commit/reset offset may cause unexpected results.

**To Reproduce**
1. produce some messages and consume to the end in a topic using full name(`persistent://public/default/test`)
2. reset the corresponding groupid to `earliest` using kafka name(`test`)
3. you will find that it does not work and returns no error

**Expected behavior**
Correctly committing offsets.

**Screenshots**
I printed out the contents inside `__consumer_offsets` and I found that `persistent://public/default/test` and `test` as the key of the message exist independently. Just like:
```log
# commit with topic: persistent://public/default/test
[my-groupid,persistent://public/default/test,0]
[my-groupid,persistent://public/default/test,1]
[my-groupid,persistent://public/default/test,2]

# commit with topic: test
[my-groupid,test,0]
[my-groupid,test,1]
[my-groupid,test,2]
```

**Additional context**
I think we better unify the topic name in `__consumer_offsets`. I will submit a PR later.

wuzhanpeng:
Resolves #584

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -61,6 +61,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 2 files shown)


BewareMyPower:
Please fix the spotbugs check and codacy check first.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -947,19 +948,67 @@ protected void handleFindCoordinatorRequest(KafkaHeaderAndRequest findCoordinato
...
+                    request.partitions() == null ? Optional.ofNullable(request.partitions()) : Optional.of(
+                            request.partitions()
+                            .stream()
+                            .map(tp -> {
+                                try {
+                                    TopicPartition newTopicPartition = new TopicPartition(
+                                            new KopTopic(tp.topic()).getFullName(), tp.partition());
+                                    replacingIndex.put(newTopicPartition, tp);
+                                    return newTopicPartition;
+                                } catch (Exception e) {
```
BewareMyPower, line 987: Catch `RuntimeException` here. Though I think a better fix is to just throw `IllegalArgumentException` in `KopTopic`'s constructor and catch `IllegalArgumentException` here.

BewareMyPower (changes requested):

BewareMyPower (approved):

PR #597, opened 2021-06-28
Update download link to synchronizing content to StreamNative hub

BewareMyPower:
When KoP is auto released by streamnative-bot, the `{{protocol:version}}` will be replaced to the actual version so that https://hub.streamnative.io/protocol-handlers/kop/0.2.0 can grab the content.


timmyyuan:
LGTM

PR #605, opened 2021-07-06
Add test framework for Kafka clients of different versions

BewareMyPower:
### Motivation

Currently KoP verifies other versions of Kafka client only by running a simple e2e test (`KafkaIntegrationTest`) that Kafka client is running inside the docker container.

This test is simple and has some flaws:
1. If we want to change the test code, we need to update the docker image.
2. We can't make use of IDE like Intellij Idea to debug the client side code.
3. It's too simple and doesn't cover the basic cases like whether the key, headers can be processed well and whether the Kafka producers/consumers of mixed versions.

And we can't test other versions just by changing the version of `kafka-clients` dependency. Because KoP server side depends on some common classes from `org.apache.kafka.common` package, which is also included in `kafka-clients` dependency. If we modified the `kafka-clients` dependency, the KoP server would be affected and might not be compiled successfully.

### Modifications
This PR uses maven-shade-plugin to import `kafka-clients` with multiple versions. Normally if we imported an artifact with multiple versions in a Maven project, it would only import one version of them eventually. However, maven-shade-plugin can include the dependencies to the JAR of a module and relocate the class path to avoid conflict.

Based on the maven-shade-plugin,  this PR adds following modifications:
1. Add a `kafka-client-api` module that includes interfaces for Kafka producer, consumer and the wrapper classes of other data structures without any `kafka-clients` dependency. Since Java doesn't support duck typing, it uses a lot of reflection operations to avoid much repeated code in the implementation of these interfaces.
2. Add a `kafka-1-0` module, which is the implementation for `kafka-client-api` with Kafka 1.0.0 dependency. It's a shaded JAR that is built by maven-shade-plugin. It relocated `org.apache.kafka.common` package to `org.apache.kafka.common-1-0-0` as well as `clients` package to avoid conflict.
3. Add a `kafka-client-factory` module as the factory class of `kafka-client-api`. It's a dependent module to avoid cyclic dependency.
4. Refactor the `pom.xml` of root directory because it imported many unused dependencies that might be not necessary for modules. Also after this change, the size of KoP NAR file reduced from 13 MB to 8 MB.
5. Add a `compatibility` package that contains tests for Kafka clients of different versions to cover following cases:
  - Different Kafka producers produce messages, different Kafka consumers consume all these messages.
  - (entryFormat=pulsar) Different Kafka producers produce messages, a Pulsar consumer consumes these messages.
  - (entryFormat=pulsar) A Pulsar producer produces a message, different Kafka consumers consume the message.

Since this PR only adds Kafka client 1.0.0, the tests only contain two Kafka versions (1.0.0 and default 2.0.0). We can add other versions in future because this PR already contains many changes.

### DIFF
--- a/kafka-1-0/pom.xml
+++ b/kafka-1-0/pom.xml
@@ -1,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>pulsar-protocol-handler-kafka-parent</artifactId>
+    <groupId>io.streamnative.pulsar.handlers</groupId>
+    <version>2.9.0-SNAPSHOT</version>
+  </parent>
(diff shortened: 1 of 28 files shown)


BewareMyPower:
To keep this PR simple, there're some additional jobs that are not included by this PR, I'll open new issues to track. Such as
1. `commitOffset` is not verified.
2. Modules for other Kafka client versions like 1.1.0 and 2.1.0 to 2.8.0 are not added.

BewareMyPower:
Some dependency changes affected the Transaction, I'll fix it soon.

BewareMyPower:
Now the tests passed, PTAL @jiazhai

jiazhai (approved):

PR #610, opened 2021-07-09
Make build pass on JDK11

eolivelli:
This patch allows you to build KOP on JDK11.

Without this patch the build fails with this error:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project kafka-client-factory: Compilation failure: Compilation failure: 
[ERROR] /Users/enrico.olivelli/dev/kop/kafka-client-factory/src/main/java/io/streamnative/kafka/client/api/KafkaClientFactoryImpl.java:[33,20] cannot access org.apache.kafka-1-0-0.clients.producer.KafkaProducer
[ERROR]   class file for org.apache.kafka-1-0-0.clients.producer.KafkaProducer not found
[ERROR] /Users/enrico.olivelli/dev/kop/kafka-client-factory/src/main/java/io/streamnative/kafka/client/api/KafkaClientFactoryImpl.java:[41,20] cannot access org.apache.kafka-1-0-0.clients.consumer.KafkaConsumer
[ERROR]   class file for org.apache.kafka-1-0-0.clients.consumer.KafkaConsumer not found
[ERROR] -> [Help 1]
[ERROR] 
```

**Modifications**
do not use '-' in the package name

As I do not have signed an ICLA with this repo, I would like to state explicitly that I am sending this patch according to the terms Apache License v2.

### DIFF
--- a/kafka-1-0/pom.xml
+++ b/kafka-1-0/pom.xml
@@ -57,11 +57,11 @@
               <relocations>
                 <relocation>
                   <pattern>org.apache.kafka.common</pattern>
-                  <shadedPattern>org.apache.kafka-1-0-0.common</shadedPattern>
+                  <shadedPattern>org.apache.kafka100.common</shadedPattern>
                 </relocation>
                 <relocation>
                   <pattern>org.apache.kafka.clients</pattern>
-                  <shadedPattern>org.apache.kafka-1-0-0.clients</shadedPattern>
+                  <shadedPattern>org.apache.kafka100.clients</shadedPattern>
                 </relocation>
               </relocations>
             </configuration>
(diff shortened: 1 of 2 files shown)


BewareMyPower:
Thanks for your contribution. It makes sense to me.

But I have a concern that some version number may have two digits like Kafka 0.10.0.0. If we supported these versions in future, what's the best way to name the package? `kafka01000`?

BewareMyPower (changes requested):
BTW, this change could affect the tests. See https://github.com/streamnative/kop/blob/master/kafka-client-api/src/main/java/io/streamnative/kafka/client/api/KafkaVersion.java

BewareMyPower:
The `getStringSerializer` and `getStringDeserializer` should also be modified, for example,

```java
    public String getStringSerializer() {
        /* ... */
        return String.format("org.apache.kafka-%s.common.serialization.StringSerializer", name);
    }
```

should be modified to

```java
    public String getStringSerializer() {
        /* ... */
        return String.format("org.apache.kafka%s.common.serialization.StringSerializer", name);
    }
```

It also applies to `getStringDeserializer()` because they will use the relocated class path to load classes of different Kafka versions.

BewareMyPower:
Hi @eolivelli , could you apply my comments to fix the broken tests?

eolivelli:
sure @BewareMyPower 
doing it now

BewareMyPower (approved):

eolivelli:
thank you !

PR #632, opened 2021-08-03
Support Kafka consumer for Pulsar format messages when entryFormat is kafka

BewareMyPower:
### Motivation

There's a limit when `entryFormat=kafka` is configured that Kafka client can only interact with Kafka client but not Pulsar client. This PR intends to support Kafka consumer interacts with Kafka producer.

BTW, to support Kafka producer interacts with Kafka consumer, we need a feature from Pulsar's side.

### Modifications

Add a method `decodePulsarEntryToKafkaRecords` to convert Pulsar's metadata and payload to Kafka's `MemoryRecords`. Since it's basically migrated from the current `PulsarEntryFormatter`'s implementation, both `KafkaEntryFormatter` and `PulsarEntryFormatter` will reuse this method.

In `KafkaEntryFormatter#decode`, for each entry, first parse the `entry.format` property that is set in `KafkaEntryFormatter#encode`, if it's not `kafka`, treat it as a message of Pulsar format and use `decodePulsarEntryToKafkaRecords` to decode it.

In addition, this PR also fixes some bugs of the original Pulsar format decode implementation:
1. Use `isPartitionKeyB64Encoded()` instead of `hasPartitionKeyB64Encoded()` to check if the key is Base64 encoded. Because currently Pulsar always set this field false so that `hasPartitionKeyB64Encoded()` returns true but the key is not Base64 encoded.
2. Don't return the `ByteBuffer` passed to `MemoryRecordsBuilder`'s constructor as the returned value. Because if `MemoryRecordsBuilder`'s internal `ByteBuffer` doesn't have enough space to contain more bytes, it will increased until the max bytes limit, then the outer `ByteBuffer` won't be the actual `ByteBuffer`. This bug was not exposed yet just because the initial buffer size is 1MB, which is great enough usually.

Finally, this PR adds two tests to verify it. One is to verify Pulsar producer and Kafka consumer for messages with or without keys or headers. The other is to verify mixed (Pulsar & Kafka) producers and Kafka consumer for messages with keys and headers.

### DIFF
--- a/kafka-client-api/src/main/java/io/streamnative/kafka/client/api/Header.java
+++ b/kafka-client-api/src/main/java/io/streamnative/kafka/client/api/Header.java
@@ -33,6 +33,9 @@ public class Header {
     private final String value;
 
     private static Header fromHeader(final Object originalHeader) {
+        if (originalHeader == null) {
+            return null;
+        }
         final Class<?> clazz = originalHeader.getClass();
         return new Header(
                 (String) ReflectionUtils.invoke(clazz, "key", originalHeader),
@@ -69,7 +72,9 @@ public class Header {
             return null;
         }
         return headers.stream()
-                .map(header -> constructor.apply(header.getKey(), header.getValue().getBytes(StandardCharsets.UTF_8)))
+                .map(header -> (header != null)
+                        ? constructor.apply(header.getKey(), header.getValue().getBytes(StandardCharsets.UTF_8))
+                        : null)
                 .collect(Collectors.toList());
     }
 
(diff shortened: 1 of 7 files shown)


BewareMyPower:
When I implement this logic, I also found some points to be improved.
1. The initial size of `MemoryRecords` is too great (1 MB). It means each time an entry is decoded, 1 MB heap memory will be allocated. If the entry's size is little, there will be much waste. Kafka's implementation just compute the sum of the record size as the initial size.
2. There's no way to reuse Netty's pool to allocate `ByteBuf` in this case. As I've said before, each time an entry is decoded, some heap memory will be allocated. But it might be a huge work to migrate a `ByteBuf` based `MemoryRecords` and related classes from Kafka.

BewareMyPower:
@jiazhai PTAL
@hangc0276 Could you also take a look if you're free?

jiazhai (approved):

Issue #385, opened 2021-02-25
[BUG] there are topics that do not belong to the KoP namespace if use kafka-topics.sh to create list-request


dockerzhang:
**Describe the bug**
there are topics that do not belong to the KoP namespace if use kafka-topics.sh to create list request.

**To Reproduce**
1. create topics in other pulsar namespaces.
2. list topic use `kafka-topics.sh`
3. See an error

**Expected behavior**
 `kafka-topics.sh` list-request only contains topics that belong to the KoP namespace.

**Screenshots**
only `10pkafka` and `10pkafka1` are KoP topics.
![image](https://user-images.githubusercontent.com/18047329/109117733-57a54b80-777d-11eb-891e-609b0fa1df41.png)

BewareMyPower:
To support multi-tenancy, it needs to find all topics since Kafka has no way to specify namespace. Currently it only filters the `__consumer_offset` topic. I have two suggestions to solve this problem.

1. Add configurations like `kopAllowedNamespaces` to specify a list of namespaces that topics can be accessed. But it would still list all topics across those namespaces.
2. Change the behavior of `handleTopicMetadataRequest`. If users wanted to list topics under a specific namespace, use wildcards symbol, like `public/default/*`.

BewareMyPower:
I'll fix this issue soon.

PR #206, opened 2020-10-27
Support tenant and namespace

BewareMyPower:
Currently KoP only support produce or consume topics in the configured namespace by `kafkaTenant` and `kafkaNamespace`. If we want to make use of Pulsar's multi-tenancy, we need to enable SASL authentication and encode the namespace info into the `sasl.jaas.config` property.

This PR is to use Pulsar topic name as Kafka topic name, the topic mapping from topic in Kafka client to the real Pulsar topic is:

- my-topic => `persistent://<kafkaTenant>/<kafkaNamespace>/my-topic`
- my-tenant/my-ns/my-topic => `persistent://my-tenant/my-ns/my-topic`

This work is done by the new added `KopTopic`.

Besides, if client want to fetch the metadata of all topics, before this PR, only the topics of `<kafkaTenant>/<kafkaNamespace>` would be returned, now the topics of all namespaces would be returned.

Following unit tests are added:

- `KopTopicTest`: test the basic functions of `KopTopic`;
- `DifferentNamespaceTest`: test the newly supported topic name version to specify a custom namespace

PR #636, opened 2021-08-06
Only allow topics under specific namespaces to be listed

BewareMyPower:
Fixes https://github.com/streamnative/kop/issues/385 

### Motivation
Currently if Kafka client want to list all topics, KoP will list topics under all namespaces. If users don't use all namespaces for Kafka clients, the other namespaces should not be accessed.

### Modifications
Add a config `kopAllowedNamespaces` to specify the allowed namespaces to list topics. If it's not set, only the default namespace is allowed.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -35,6 +35,7 @@ Pulsar is a multi-tenant system that requires users to specify the [tenant and n
 | kafkaNamespace         | The default namespace of Kafka topics          | default |
 | kafkaMetadataTenant    | The tenant used for storing Kafka metadata topics    | public  |
 | kafkaMetadataNamespace | The namespace used for storing Kafka metadata topics | __kafka |
+| kopAllowedNamespaces   | The allowed namespace to list topics with a comma separator.<br>For example, "public/default,public/kafka".<br>If it's not set or empty, the allowed namespaces will be "<kafkaTenant>/<kafkaNamespace>". | |
 
 ## Performance
 
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -199,7 +199,7 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
     public final int maxReadEntriesNum;
     private final String offsetsTopicName;
     private final String txnTopicName;
-    private final List<String> allowedNamespaces;
+    private final Set<String> allowedNamespaces;
     // store the group name for current connected client.
     private final ConcurrentHashMap<String, String> currentConnectedGroup;
     private final String groupIdStoredPath;
@@ -265,7 +265,7 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
                 kafkaConfig.getKafkaMetadataNamespace(),
                 TRANSACTION_STATE_TOPIC_NAME)
         ).getFullName();
-        this.allowedNamespaces = kafkaConfig.getAllowedNamespaces();
+        this.allowedNamespaces = kafkaConfig.getKopAllowedNamespaces();
         this.entryFormatter = EntryFormatterFactory.create(kafkaConfig.getEntryFormat());
         this.currentConnectedGroup = new ConcurrentHashMap<>();
         this.groupIdStoredPath = kafkaConfig.getGroupIdZooKeeperPath();
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -197,6 +197,9 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
     private final String advertisedListeners;
     private final int defaultNumPartitions;
     public final int maxReadEntriesNum;
+    private final String offsetsTopicName;
+    private final String txnTopicName;
+    private final List<String> allowedNamespaces;
     // store the group name for current connected client.
     private final ConcurrentHashMap<String, String> currentConnectedGroup;
     private final String groupIdStoredPath;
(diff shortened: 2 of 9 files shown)


BewareMyPower:
The `listTopics()` method of Kafka consumer needs to fix.

docs/configuration.md
```diff
@@ -35,6 +35,7 @@ Pulsar is a multi-tenant system that requires users to specify the [tenant and n
 | kafkaNamespace         | The default namespace of Kafka topics          | default |
 | kafkaMetadataTenant    | The tenant used for storing Kafka metadata topics    | public  |
 | kafkaMetadataNamespace | The namespace used for storing Kafka metadata topics | __kafka |
+| kopAllowedNamespaces   | The allowed namespace to list topics with a comma separator.<br>For example, "public/default,public/kafka".<br>If it's not set or empty, the allowed namespaces will be "<kafkaTenant>/<kafkaNamespace>". | |
```
sijie, line 38: > If it's not set or empty, the allowed namespaces will be "<kafkaTenant>/<kafkaNamespace>"

Doesn't this change the default behavior? I thought if it is not set or empty, the allowed namespaces will be all namespaces, no?

BewareMyPower: It changes the default behavior. It it's not set or empty, the allowed namespaces will only be the default namespace (`public/default`).

BewareMyPower: The current semantic before this PR could go back to https://github.com/streamnative/kop/pull/206, which allows users to specific a topic of any namespace without any SASL authentication.
- Before #206, `AdminClient#listTopics()` returns the topic list of `<kafkaTenant>/<kafkaNamespace>`.
- Before this PR, `AdminClient#listTopics()` returns the topic list of all namespaces. However, this semantic is dangerous and may cause heavy load when KoP is enabled on an existing cluster with a lot of topics even if authentication is enabled.
- After this PR, `AdminClient#listTopics()` returns the topic list of `<kafkaTenant>/<kafkaNamespace>`. What's different is, we can configure the `kopAllowedNamespaces`. Then `AdminClient#listTopics()` can return the topic list of specific namespaces.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/MetricsProviderTest.java
```diff
@@ -166,8 +166,9 @@ public void testMetricsProvider() throws Exception {
         Assert.assertTrue(sb.toString().contains("request=\"ApiVersions\""));
         Assert.assertTrue(sb.toString().contains("request=\"ListOffsets\""));
         Assert.assertTrue(sb.toString().contains("request=\"Fetch\""));
-        Assert.assertTrue(sb.toString().contains("kop_server_REQUEST_LATENCY{success=\"true\",quantile=\"0.99\", "
-                + "request=\"Produce\"}"));
+        // The followed check may fail in CI environment, comment it first
+        //Assert.assertTrue(sb.toString().contains("kop_server_REQUEST_LATENCY{success=\"true\",quantile=\"0.99\", "
+        //        + "request=\"Produce\"}"));
```
BewareMyPower, line 171: I commented this line because it's very likely to fail in CI environment. It looks like this check is too strict. Do you have any idea? @hangc0276 

Here is the full metric string: [metrics-string.log](https://github.com/streamnative/kop/files/6947504/metrics-string.log)

docs/configuration.md
BewareMyPower: @sijie Is there any more question?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
```diff
@@ -340,6 +343,14 @@
     )
     private int kopPrometheusStatsLatencyRolloverSeconds = 60;
 
+    @FieldContext(
+            category = CATEGORY_KOP,
+            doc = "The allowed namespaces to list topics with a comma separator.\n"
+                    + " For example, \"public/default,public/kafka\".\n"
+                    + "If it's not set or empty, the allowed namespaces will be \"<kafkaTenant>/<kafkaNamespace>\"."
+    )
+    private String kopAllowedNamespaces;
```
codelipenghui, line 352: We should check if the provided namespace is valid to void any invalid namespace names, In Pulsar `NamespaceName.validateNamespaceName()` can be used to check.

codelipenghui: And for the configuration class, I think we can define as Set<String> directly. Link to https://github.com/apache/pulsar/blob/6704f12104219611164aa2bb5bbdfc929613f1bf/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java#L914

tests/src/test/java/io/streamnative/pulsar/handlers/kop/MetricsProviderTest.java
codelipenghui: @BewareMyPower Could you please raise up an issue to track the metrics fix? So that we can unblock this PR.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -393,63 +407,45 @@ protected void handleInactive(KafkaHeaderAndRequest kafkaHeaderAndRequest,
...
-                return null;
-            }
-            Map<String, List<TopicName>> topicMap = Maps.newConcurrentMap();
+        final Map<String, List<TopicName>> topicMap = new ConcurrentHashMap<>();
+        final AtomicInteger pendingNamespacesCount = new AtomicInteger(allowedNamespaces.size());
+
+        for (String namespace : allowedNamespaces) {
+            pulsarService.getNamespaceService().getListOfPersistentTopics(NamespaceName.get(namespace))
+                    .whenComplete((topics, e) -> {
+                        if (e != null) {
```
codelipenghui, line 427: We might get namespaces not exist here? If yes, we should not failed the request here.

BewareMyPower: If the namespace doesn't exist, in the callback, `topics` will be empty and `e` is still null. I'll add the tests.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/MetricsProviderTest.java
BewareMyPower: https://github.com/streamnative/kop/issues/643

docs/configuration.md
sijie: Okay. Make sense.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
codelipenghui: I mean one namespace does not exist because here is check all the namespaces. Such as we have 10 namespaces but when list topics for a namespace of the 10 namespaces, we got a namespace not exist exception here, we should return the topic list of the other 9 namespaces.

codelipenghui (approved):

Issue #633, opened 2021-08-05
[BUG] Lookup can not find broker address in k8s


wangjialing218:
**Describe the bug**
Deploy broker with kop in k8s, expose ports with nodeport.
when `advertisedAddress` is not same with the first address in `advertisedListeners`, lookup can not find broker address in k8s

**To Reproduce**
I use nodeip:nodeport to expose the address of brokers in k8s.
Here is the service stats of broker.
![kop](https://user-images.githubusercontent.com/65590138/128325574-5e77823f-1b9f-42e8-bcf1-702a3de4f173.PNG)

configuration of broker.conf:
`advertisedAddress: {worknode_ip}:30873`
`advertisedListeners: internal:{headless service domain}:6650,external:pulsar://{worknode_ip}:31964`
`kafkaListeners=PLAINTEXT://0.0.0.0:9092`
`kafkaAdvertisedListeners=PLAINTEXT://{worknode_ip}:31553`

Pulsar client can connect to pulsar outside k8s with `pulsar://{worknode_ip}:31964` by setting `listenerName` with `external`
But when use kafka client outside k8s connect to pulsar with `{worknode_ip}:31553`, it will fail to get topic metadata due to broker lookup failed.

**Additional context**
Current lookup logic of kop:
1. use `pulsarService.getClient().getLookup().getBroker()` to find the broker of topic, result is `InetSocketAddress`
2. list all broker address in zk under `/loadbalance/brokers/`
3. find match brokers in step 2 which host name is same with the `InetSocketAddress` in step 1
4. find the broker which pulsarServiceUrl and port or WebServiceUrl and port is same with `InetSocketAddress`
5. return `kafkaAdvertisedListeners` of that broker in step 4

In my condition, in step 1 the lookup result is the first address in `advertisedListeners`. which is `{headless service domain}:6650`, and in step 2 the broker address under `/loadbalance/brokers/` is `{worknode_ip}:30873`. because the broker address here will be used in admin request redirection, I need to set the address here with nodeport in order to handle admin request outside.
Then because `{headless service domain}` is not same with `{worknode_ip}`, step 3 will fail with no match broker.
![kop2](https://user-images.githubusercontent.com/65590138/128326346-be2ffe0d-1760-4926-994f-2d679e696a11.PNG)

Current workaround:
In step 1,  I create a pulsar client for topic lookup with listenerName `extenrnal` instead of use the pulsar client of `pulsarService.getClient()`, so the result could be `{worknode_ip}:31964`

[...]

Issue #574, opened 2021-06-15
[FEATURE] Support multiple listeners of one protocol


BewareMyPower:
**Is your feature request related to a problem? Please describe.**

Currently for KoP we can configure `kafkaListeners` as what interfaces KoP binds to and `kafkaAdvertisedListeners` as how clients connect. However, currently KoP can only bind to one interface for each protocol.

For example, if KoP was configured with

```properties
kafkaListeners=PLAINTEXT://127.0.0.1:9092,SSL://127.0.0.1:9093
```

it would work well.

But if KoP was configured with

```properties
kafkaListeners=PLAINTEXT://127.0.0.1:9092,SSL://127.0.0.1:9093,SSL://127.0.0.1:9094
```

KoP would fail to start with following logs.

```
15:56:29.408 [main] ERROR io.streamnative.pulsar.handlers.kop.KafkaProtocolHandler - KafkaProtocolHandler newChannelInitializers failed with 
java.lang.IllegalStateException: PLAINTEXT://127.0.0.1:9092,SSL://127.0.0.1:9093,SSL://127.0.0.1:9094 has multiple listeners whose protocol is SSL
	at io.streamnative.pulsar.handlers.kop.EndPoint.parseListeners(EndPoint.java:77) ~[pulsar-protocol-handler-kafka-2.8.0-rc-202106091215.nar-unpacked/:?]
	at io.streamnative.pulsar.handlers.kop.KafkaProtocolHandler.newChannelInitializers(KafkaProtocolHandler.java:313) [pulsar-protocol-handler-kafka-2.8.0-rc-202106091215.nar-unpacked/:?]
	at org.apache.pulsar.broker.protocol.ProtocolHandlerWithClassLoader.newChannelInitializers(ProtocolHandlerWithClassLoader.java:71) [io.streamnative-pulsar-broker-2.8.0-rc-202106071430.jar:2.8.0-rc-202106071430]
	at org.apache.pulsar.broker.protocol.ProtocolHandlers.newChannelInitializers(ProtocolHandlers.java:124) [io.streamnative-pulsar-broker-2.8.0-rc-202106071430.jar:2.8.0-rc-202106071430]
	at org.apache.pulsar.broker.PulsarService.start(PulsarService.java:787) [io.streamnative-pulsar-broker-2.8.0-rc-202106071430.jar:2.8.0-rc-202106071430]
[...]

PR #646, opened 2021-08-11
Allow setting listener name for topic lookup and optimize the lookup

### LINKED ISSUE #644: [Enhancement] Use BrokerService instead of PulsarClient for topic lookup
Currently KoP use Pulsar broker's builtin `PulsarClient` for topic lookup, see https://github.com/streamnative/kop/blob/9630e5c0de041119ec2027502f655a3ed814f12b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java#L211-L212

There's a limit for this way. For example, all connections share the same `PulsarClient` instance. However, there's a pending requests limit on `PulsarClient`, i.e. if there're many connections, the load on a `PulsarClient` will be high and `TooManyRequestsException` could be thrown. There's another case that if a namespace has many topics, when a Kafka client tries to list all topics, the `TooManyRequestsException` could also be thrown because all topics need to be looked up in METADATA request handler.

Use `BrokerService` instead of `PulsarClient` for topic lookup.

BewareMyPower:
Fixes #644 

### Motivation

Using `PulsarClient` for topic lookup might encounter `TooManyRequestException` easily because all client connections share the same `PulsarClient` for topic lookup. It's also a waste of resource.

### Modifications

Add a common method `KafkaProtocolHandler#getTopic` for all places that need topic lookup in KoP. In this method, first use `NamespaceService.getBrokerServiceUrlAsync` to get broker's Internet address.

However, when the namespace bundle is not owned by any broker, which usually happens during a bundle unload, the `getBrokerServiceUrlAsync` might return a `LookupResult` instance whose `redirect` field is true. Pulsar client can handle this field and resend the lookup request to another broker, but Kafka client cannot do it. In this case, this PR use `PulsarClient` for topic lookup as a fallback.

In addition, this PR adds a `kafkaListenerName` config and `getPulsarClientImpl()` method to create a builtin `PulsarClient` with listener name configured.

Since the `DistributedClusterTest` might fail if its three tests were run in a whole, this PR removes the lookup logic in `OffsetAndTopicListener#onLoad`.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
@@ -45,11 +45,13 @@ import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.configuration.Configuration;
 import org.apache.commons.configuration.PropertiesConfiguration;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.kafka.common.internals.Topic;
 import org.apache.kafka.common.record.CompressionType;
 import org.apache.kafka.common.security.auth.SecurityProtocol;
 import org.apache.kafka.common.utils.Time;
 import org.apache.pulsar.broker.PulsarServerException;
+import org.apache.pulsar.broker.PulsarService;
 import org.apache.pulsar.broker.ServiceConfiguration;
 import org.apache.pulsar.broker.namespace.NamespaceBundleOwnershipListener;
 import org.apache.pulsar.broker.protocol.ProtocolHandler;
@@ -132,23 +134,10 @@ public class KafkaProtocolHandler implements ProtocolHandler {
                             KafkaTopicManager.removeTopicManagerCache(name.toString());
                             KopBrokerLookupManager.removeTopicManagerCache(name.toString());
                             // update lookup cache when onload
-                            try {
-                                CompletableFuture<InetSocketAddress> retFuture = new CompletableFuture<>();
-                                ((PulsarClientImpl) service.pulsar().getClient()).getLookup()
-                                        .getBroker(TopicName.get(topic))
-                                        .whenComplete((pair, throwable) -> {
-                                            if (throwable != null) {
-                                                log.warn("cloud not get broker", throwable);
-                                                retFuture.complete(null);
-                                            }
-                                            checkState(pair.getLeft().equals(pair.getRight()));
-                                            retFuture.complete(pair.getLeft());
-                                        });
-                                KafkaTopicManager.LOOKUP_CACHE.put(topic, retFuture);
-                                KopBrokerLookupManager.updateTopicManagerCache(topic, retFuture);
-                            } catch (PulsarServerException e) {
-                                log.error("onLoad PulsarServerException ", e);
-                            }
+                            final CompletableFuture<InetSocketAddress> retFuture =
+                                    KafkaProtocolHandler.getBroker(service.pulsar(), TopicName.get(topic));
+                            KafkaTopicManager.LOOKUP_CACHE.put(topic, retFuture);
+                            KopBrokerLookupManager.updateTopicManagerCache(topic, retFuture);
                         }
                     } else {
                         log.error("Failed to get owned topic list for "
@@ -462,6 +451,19 @@ public class KafkaProtocolHandler implements ProtocolHandler {
             FutureUtil.waitForAll(lists).get();
         } else {
             log.info("Current broker: {} does not own any of the txn log topic partitions", currentBroker);
+        }
+    }
+
+    public static CompletableFuture<InetSocketAddress> getBroker(final PulsarService pulsarService,
+                                                                 final TopicName topicName) {
+        try {
+            return ((PulsarClientImpl) pulsarService.getClient()).getLookup()
+                    .getBroker(topicName)
+                    .thenApply(Pair::getLeft);
+        } catch (PulsarServerException e) {
+            final CompletableFuture<InetSocketAddress> addressFuture = new CompletableFuture<>();
+            addressFuture.completeExceptionally(e);
+            return CompletableFuture.completedFuture(null);
         }
     }
 }
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
@@ -280,7 +280,9 @@ public class KafkaProtocolHandler implements ProtocolHandler {
             throw new IllegalStateException(e);
         }
 
-        final PulsarClient pulsarClient = getPulsarClientImpl(brokerService.getPulsar());
+        // Create PulsarClient for topic lookup, the listenerName will be set if kafkaListenerName is configured.
+        // After it's created successfully, this method won't throw any exception.
+        getPulsarClientImpl(brokerService.getPulsar());
         final ClusterData clusterData = ClusterData.builder()
                 .serviceUrl(brokerService.getPulsar().getWebServiceAddress())
                 .serviceUrlTls(brokerService.getPulsar().getWebServiceAddressTls())
@@ -288,6 +290,14 @@ public class KafkaProtocolHandler implements ProtocolHandler {
                 .brokerServiceUrlTls(brokerService.getPulsar().getBrokerServiceUrlTls())
                 .build();
 
+        // Use the builtin PulsarClient for creating producers and readers in group coordinator
+        PulsarClient pulsarClient;
+        try {
+            pulsarClient = brokerService.getPulsar().getClient();
+        } catch (PulsarServerException e) {
+            log.error("Failed to create builtin PulsarClient", e);
+            throw new IllegalStateException(e);
+        }
         try {
             MetadataUtils.createOffsetMetadataIfMissing(pulsarAdmin, clusterData, kafkaConfig);
         } catch (PulsarAdminException e) {
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
@@ -545,9 +545,10 @@ public class KafkaProtocolHandler implements ProtocolHandler {
             return getFailedAddressFuture(new IllegalStateException("NamespaceService is not available"));
         }
 
+        final PulsarClientImpl pulsarClient = getPulsarClientImpl(pulsarService);
         final LookupOptions options = LookupOptions.builder()
                 .authoritative(false)
-                .advertisedListenerName(null)
+                .advertisedListenerName(pulsarClient.getConfiguration().getListenerName())
                 .loadTopicsInBundle(true)
                 .build();
         final CompletableFuture<InetSocketAddress> future =
@@ -564,7 +565,7 @@ public class KafkaProtocolHandler implements ProtocolHandler {
             final LookupResult lookupResult = optLookupResult.get();
             if (lookupResult.isRedirect()) {
                 // Kafka client can't process redirect field, so here we fallback to PulsarClient's topic lookup
-                return getPulsarClientImpl(pulsarService).getLookup().getBroker(topicName).thenApply(Pair::getLeft);
+                return pulsarClient.getLookup().getBroker(topicName).thenApply(Pair::getLeft);
             } else {
                 return getAddressFutureFromBrokerUrl(lookupResult.getLookupData().getBrokerUrl());
             }
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
@@ -16,7 +16,6 @@ package io.streamnative.pulsar.handlers.kop;
 import static com.google.common.base.Preconditions.checkState;
 import static io.streamnative.pulsar.handlers.kop.KopServerStats.SERVER_SCOPE;
 import static io.streamnative.pulsar.handlers.kop.utils.TopicNameUtils.getKafkaTopicNameFromPulsarTopicname;
-import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static org.apache.pulsar.common.naming.TopicName.PARTITIONED_TOPIC_SUFFIX;
 
 import com.google.common.collect.ImmutableMap;
@@ -41,9 +40,9 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import lombok.Getter;
@@ -51,7 +50,6 @@ import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.configuration.Configuration;
 import org.apache.commons.configuration.PropertiesConfiguration;
-import org.apache.commons.lang3.tuple.Pair;
 import org.apache.kafka.common.internals.Topic;
 import org.apache.kafka.common.record.CompressionType;
 import org.apache.kafka.common.security.auth.SecurityProtocol;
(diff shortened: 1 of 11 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
```diff
@@ -464,4 +459,66 @@ private void loadTxnLogTopics(TransactionCoordinator txnCoordinator) throws Exce
...
+                log.debug("[{}] Lookup result {}", topicName.toString(), optLookupResult);
+            }
+            if (!optLookupResult.isPresent()) {
+                return getFailedAddressFuture(ClientCnx.getPulsarClientException(
+                        ServerError.ServiceNotReady,
+                        "No broker was available to own " + topicName));
+            }
+
+            final LookupResult lookupResult = optLookupResult.get();
+            if (lookupResult.isRedirect()) {
```
BewareMyPower, line 504: @codelipenghui In this case, it seems there's no good way to handle it. I've tried two ways before.
1. Complete it with the broker url of lookup data. However, then Kafka client will send a subsequent PRODUCE request to a non-owner broker.
2. Complete it exceptionally just like the case `optLookupResult` is not present. Then Kafka client still failed.
They can both be reproduced in KoP's `DistributedClusterTest`, which run two mocked broker.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
```diff
@@ -464,4 +459,66 @@ private void loadTxnLogTopics(TransactionCoordinator txnCoordinator) throws Exce
...
+    public static CompletableFuture<InetSocketAddress> getBroker(final PulsarService pulsarService,
+                                                                 final TopicName topicName) {
+        final NamespaceService namespaceService = pulsarService.getNamespaceService();
+        if (namespaceService == null) {
+            return getFailedAddressFuture(new IllegalStateException("NamespaceService is not available"));
+        }
+
+        final LookupOptions options = LookupOptions.builder()
+                .authoritative(false)
+                .advertisedListenerName(null)
```
BewareMyPower, line 489: @wangjialing218 I think we can set advertised listener name here like https://github.com/streamnative/kop/issues/644#issuecomment-895770747. But it's not clear that how to get the listener name?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
BewareMyPower: The reason might be that if we completed the future immediately, the actual owner broker's ownership cache could not be updated without an extra request from client side.

wangjialing218: the candidate broker is already in `lookupResult` here, can we create a pulsar client with the candidate broker url to send the extra request with `authoritative = true`?
another way we could return the candidate broker url as lookup result to Kafka client, Kafka client will send next request to the candidate broker, we can update the ownership cache there in the next request handler.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
wangjialing218: we could add configration in broker.conf like this in k8s:

> kafkaListenerName=kafkaListener
> kafkaListeners=PLAINTEXT://0.0.0.0:9092
> kafkaAdvertisedListeners=PLAINTEXT://{k8s node ip}:{nodeport of 9092}
> advertisedListeners=internal:pulsar://{pod clusterIP}:6650,external:pulsar://{k8s node ip}:{nodeport of 6650},kafkaListener:pulsar://{k8s node ip}:{nodeport of 9092}

and we set advertised listener here with the value of `kafkaListenerName`

We could get the lookup result for kafka advertised address `{k8s node ip}:{nodeport of 9092}` by set listener name. No neet to convert the broker address to kafka advertised address through `KopBrokerLookupManager.getProtocolDataToAdvertise`, which may cause the issue in #633

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
BewareMyPower: > can we create a pulsar client with the candidate broker url to send the extra request with authoritative = true

Sorry I don't understand well. Here I used `PulsarAdmin` to do it. Can you explain the difference between `PulsarClient` and `PulsarAdmin` for topic lookup? And what does `authoritative = true` means?

> another way we could return the candidate broker url as lookup result to Kafka client

This way doesn't work. If we return the candicate broker url to Kafka client, the Kafka client won't send another round of METADATA request. Instead, it will send PRODUCE request to that broker directly, then `NOT_LEADER_FOR_PARTITION` happens.

wangjialing218: 1. `admin.lookups().lookupTopicAsync()` currently does not support listener name, so we can not set listener name here to do topic lookup. Listener name is supported by HttpLookupService by https://github.com/apache/pulsar/pull/10319 but admin api miss to add the parameter.
2. Http lookup will call `TopicLookupBase.internalLookupTopicAsync`，which set `loadTopicsInBundle=false`, but we set `.loadTopicsInBundle(true)` before, I'm not sure if this will cause any difference in future operation. 
3. `authoritative = true` means this is a redirect request, the broker which handle the request is the candidate broker to own the topic. otherwise loadmanager will select a candidate broker again.

BewareMyPower: Thanks for you detail explanation very much! I'll apply your suggestions soon.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
BewareMyPower: If I understand correctly, set `advertisedListenerName` to the configured `kafkaListenerName`, right?

> No neet to convert the broker address to kafka advertised address through KopBrokerLookupManager.getProtocolDataToAdvertise

Regarding to this problem, I'll read the related code and confirm it soon.

wangjialing218: > If I understand correctly, set `advertisedListenerName` to the configured `kafkaListenerName`, right?
Yes. Current we support only one kafka advertised listener, we could set the listener name in configuration.
But kafka support multi advertised listeners, when we start the work for #574, we may need to consider how to handle multi advertised listener case.

BewareMyPower:
I'll add a test for `kafkaListenerName` soon.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
```diff
@@ -157,10 +153,9 @@ public static void clearTopicManagerCache() {
         KOP_ADDRESS_CACHE.clear();
     }
 
-    // exception throw for pulsar.getClient();
-    private Producer registerInPersistentTopic(PersistentTopic persistentTopic) throws Exception {
+    private Producer registerInPersistentTopic(PersistentTopic persistentTopic) {
         Producer producer = new InternalProducer(persistentTopic, internalServerCnx,
-            ((PulsarClientImpl) (pulsarService.getClient())).newRequestId(),
+            KafkaProtocolHandler.getPulsarClientImpl(pulsarService).newRequestId(),
```
wangjialing218, line 158: If you create the pulsar client with listener name `kafkaListenerName`, this client should be only used to do topic lookup for kafka client, because the lookup result is kafka advertised address.
Here you create a pulsar producer from this client to send pulsar msg, that will casue pulsar producer connect to kafka advertised address and fail to send msg.

wangjialing218 (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
wangjialing218: I missunderstanded the code here, the `InternalProducer` will not do topic lookup. But I found `GroupCoordinator` seems to create a pulsar producer from this client, please also have a check.

BewareMyPower: Fix it.

BewareMyPower:
Except the known tests, there're still two tests that failed in CI but succeeded in my local env. I'll take a look.

```
Error:    DistributedClusterTest.testMutiBrokerUnloadReload:424->kafkaPublishMessage:197 » Execution
Error:    KafkaListenerNameTest.testListenerName » ThreadTimeout Method io.streamnative....
```

BewareMyPower:
Now all tests passed. But the `DistributedClusterTest` is still unstable. Here're the test results in my local env. Each test case was run for 10 times, `O` represents success, `X` represents failure.

1. Run the whole tests

```bash
mvn test -pl tests -Dtest='DistributedClusterTest'
```

OOXXO XXOOO

All failure reasons are the same:

```
java.util.concurrent.ExecutionException: org.apache.kafka.common.errors.TimeoutException: Expiring 1 record(s) for kopMutiBrokerUnloadReload10-0: 10004 ms has passed since batch creation plus linger time
	at org.apache.kafka.clients.producer.internals.FutureRecordMetadata.valueOrError(FutureRecordMetadata.java:94)
	at org.apache.kafka.clients.producer.internals.FutureRecordMetadata.get(FutureRecordMetadata.java:64)
	at org.apache.kafka.clients.producer.internals.FutureRecordMetadata.get(FutureRecordMetadata.java:29)
	at io.streamnative.pulsar.handlers.kop.DistributedClusterTest.kafkaPublishMessage(DistributedClusterTest.java:197)
```

The broker logs are like

```
23:58:14.251 [bookkeeper-ml-scheduler-OrderedScheduler-4-0:org.apache.pulsar.broker.PulsarService@1011] INFO  org.apache.pulsar.broker.PulsarService - Loaded 4 topics on public/default/0x00000000_0x80000000 -- time taken: 0.003 seconds
23:58:17.145 [pulsar-load-manager-96-1:org.apache.pulsar.broker.loadbalance.impl.ModularLoadManagerImpl@437] INFO  org.apache.pulsar.broker.loadbalance.impl.ModularLoadManagerImpl - Writing local data to metadata store because maximum change 41.955193877220154% exceeded threshold 10%; time since last report written is 5.01 seconds
23:58:18.467 [pulsar-load-manager-137-1:org.apache.pulsar.broker.loadbalance.impl.ModularLoadManagerImpl@437] INFO  org.apache.pulsar.broker.loadbalance.impl.ModularLoadManagerImpl - Writing local data to metadata store because maximum change 38.13229501247406% exceeded threshold 10%; time since last report written is 5.01 seconds
23:58:24.243 [kafka-producer-network-thread | DemoKafkaOnPulsarProducer:org.apache.kafka.clients.NetworkClient@698] DEBUG org.apache.kafka.clients.NetworkClient - [Producer clientId=DemoKafkaOnPulsarProducer] Disconnecting from node 36016841 due to request timeout.
23:58:24.243 [pulsar-io-140-3:io.streamnative.pulsar.handlers.kop.KafkaRequestHandler@298] INFO  io.streamnative.pulsar.handlers.kop.KafkaRequestHandler - channel inactive [id: 0x695fd4c4, L:/127.0.0.1:15011 ! R:/127.0.0.1:51123]
23:58:24.244 [pulsar-io-140-3:io.streamnative.pulsar.handlers.kop.KafkaCommandDecoder@88] INFO  io.streamnative.pulsar.handlers.kop.KafkaCommandDecoder - close channel [id: 0x695fd4c4, L:/127.0.0.1:15011 ! R:/127.0.0.1:51123] with 1 pending responses
```

There's 1 pending response (`METADATA`) that has been pending for over 10 seconds, then Kafka client's send failed and the connection was closed. 

2. Only run the unload test

```bash
mvn test -pl tests -Dtest='DistributedClusterTest#testMutiBrokerUnloadReload'
```

OOOOO OOOOO

Never failed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -2044,7 +2044,7 @@ protected boolean isTransactionTopic(String topic) {
         CompletableFuture<PartitionMetadata> returnFuture = new CompletableFuture<>();
 
         topicManager.getTopicBroker(topic.toString())
-            .thenCompose(pair -> getProtocolDataToAdvertise(pair, topic))
+            .thenCompose(address -> getProtocolDataToAdvertise(address, topic))
```
wangjialing218, line 2048: can we add a judgement in `getProtocolDataToAdvertise`? if `kafkaListenerName` is used, do not convert broker url to kafka advertised address in `getProtocolDataToAdvertise`, just use the `address` here combine with `SecurityProtocol`.

wangjialing218:
> java.util.concurrent.ExecutionException: org.apache.kafka.common.errors.TimeoutException: Expiring 1 record(s) for kopMutiBrokerUnloadReload10-0: 10004 ms has passed since batch creation plus linger time
> 	at org.apache.kafka.clients.producer.internals.FutureRecordMetadata.valueOrError(FutureRecordMetadata.java:94)
> 	at org.apache.kafka.clients.producer.internals.FutureRecordMetadata.get(FutureRecordMetadata.java:64)
> 	at org.apache.kafka.clients.producer.internals.FutureRecordMetadata.get(FutureRecordMetadata.java:29)
> 	at io.streamnative.pulsar.handlers.kop.DistributedClusterTest.kafkaPublishMessage(DistributedClusterTest.java:197)
> ```

I also got this error sometimes in my local env.
According to debug log of kop, there is some lookup operation blocked, `FindBroker for topic kopMutiBrokerUnloadReload10, partitions found/all:` only up to 9/10, and the bloked lookup operation is `namespaceService.getBrokerServiceUrlAsync`
I'm not sure if this problem is related with https://github.com/apache/pulsar/pull/11310, do we base on the broker version with this pr?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
BewareMyPower: Could you explain why not use `getProtocolDataToAdvertise` if `kafkaListenerName` is used? For example, in `KafkaListenerNameTest`, the `address` here is `localhost:15002` and `getProtocolDataToAdvertise(address, topic)` will be completed with `PLAINTEXT://localhost:15003,SSL://localhost:15004`.

If we don't use `getProtocolDataToAdvertise`, which `SecurityProtocol` should we use? Could you give an example implementation and tell the reason?

BewareMyPower:
> I'm not sure if this problem is related with apache/pulsar#11310, do we base on the broker version with this pr?

I'm afraid not. I upgraded the pulsar dependency to 2.8.0.12 that contains this PR and the test is still flaky. I suspected there's something wrong with the clean up phase. If we run unload test first (rename to `test1MutiBrokerUnloadReload`) it would never fail.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
wangjialing218: I think the configuration in `KafkaListenerNameTest` is not correct. We should not set `InternalListenerName` same with `KafkaListenerName`. Broker will use the address in `InternalListenerName` as advertised address and regist to zk path `/loadbalance/brokers`.
If `InternalListenerName` is not set, default the first listener `internal` will be regist to zk path and `getProtocolDataToAdvertise` will failed with "No node for broker under zk://loadbalance"

BewareMyPower: Could you give an example config in test environment? Since there's no proxy, broker can only listen on `localhost:<brokerPort>`, other listeners are unused.

BewareMyPower: Do you mean given following config

```
internalListenerName=internal
advertisedListeners=internal:pulsar://localhost:6650,external:PLAINTEXT://localhost:9092
kafkaListenerName=external
```

The returned address (`stringOptional` is following code) should be `localhost:9092`? 

```java
        topicManager.getTopicBroker(topic.toString())
            .thenCompose(address -> getProtocolDataToAdvertise(address, topic))
            .whenComplete((stringOptional, throwable) -> {
````

Comment (BewareMyPower) [reply kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java:2048]: @wangjialing218 To correct my previous comment, the `advertisedListeners` cannot accept protocols other than `pulsar` or `pulsar+ssl`. Therefore, the `internal:pulsar://localhost:6650,external:PLAINTEXT://localhost:9092` is invalid.

If I didn't understand wrong, maybe you want to configure

```bash
advertisedListeners=pulsar:pulsar://localhost:6650,kop:pulsar://localhost:9092
```

and return the `PLAINTEXT://localhost:9092` as the listeners that will be passed to `EndPoint.getSslEndPoint` or `EndPoint.getPlainTextEndPoint`?

If so, I think you can open a new PR that is based on this PR. Because it introduced a new mapping rule from Pulsar's `TopicDomain` to Kafka's `SecurityProtocol`:
- `pulsar` -> `PLAINTEXT`
- `pulsar+ssl` -> `SSL`
- `SASL_PLAINTEXT` and `SASL_SSL` are not supported in KoP because we use `saslAllowedMechanisms` config to differ whether the SASL is enabled.

And it should be documented as well. I think this PR's original purpose is to provide a dependent method for topic lookup.

wangjialing218: I want to configure like this:
 `advertisedListeners=pulsar:pulsar://192.168.0.1:6650,kop:pulsar://10.254.1.1:9092`
192.168.0.1 is private network ip and 10.254.1.1 is public network ip. 
In order to deploy pulsar cluster that can be accessed through both public network private network.
https://github.com/apache/pulsar/wiki/PIP-61:-Advertised-multiple-addresses

Current If I configure like this, `getProtocolDataToAdvertise` will failed.
I will open a new PR after this PR merged.

BewareMyPower: Got it.

Demogorgon314 (approved):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
BewareMyPower: @wangjialing218 It's merged now, you can start the work.

PR #648, opened 2021-08-12
Support pulsar to obtain the correct unconfigured hostname listeners of KafkaProtocolHandler.

### LINKED ISSUE #647: [BUG] Unbale to find OwnedBundle for topic error unable to recover automatically
**Describe the bug**
broker.conf:
`messagingProtocols=kafka
protocolHandlerDirectory=./protocols
allowAutoTopicCreationType=partitioned

kafkaListeners=PLAINTEXT://:9996
brokerEntryMetadataInterceptors=org.apache.pulsar.common.intercept.AppendIndexMetadataInterceptor
brokerDeleteInactiveTopicsEnabled=false`

we can see znode value for /loadbalance/brokers/10.xx.xx.58:8089:
"protocols": {
        "kafka": "PLAINTEXT://:9996"
    }

When kop findBroker(topic), it will get the pulsar node metadata. The protocol about kafka is the information of the above znode. When constructing the Endpoint, the current local hostname will be used to construct the node. If the current node is not the topic's onwer broker, the partition leader in TopicMetadata obtained by the client is always the wrong leader node.

image

The current leader node of persistent://public/default/test_v1_01-partition-0 is 57 node, but when PartitionMeata is constructed it is 58 node.

**To Reproduce**
1. Configuration in broker.conf: kafkaListeners=PLAINTEXT://:9996 
2. Start multiple pulsar brokers
3. The kafka client uses a different broker address to access kop

wenbingshen:
Fixes #647 

### Motivation
Pulsar will access the getProtocolDataToAdvertise of the protocol handler when constructing LocalBrokerData. If kafkaListeners is not configured with a host name, the protocols of Pulsar LocalBrokerData will not have the host name information of the kop protocol, which will cause kop to return the wrong partition leader.

### Modifications
`getKafkaAdvertisedListeners` will check whether the listeners contain the host name, if not, use `InetAddress.getLocalHost().getCanonicalHostName()` instead

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
@@ -13,6 +13,8 @@
  */
 package io.streamnative.pulsar.handlers.kop;
 
+import static com.google.common.base.Preconditions.checkState;
+
 import com.google.common.collect.Sets;
 import io.streamnative.pulsar.handlers.kop.coordinator.group.OffsetConfig;
 import java.io.FileInputStream;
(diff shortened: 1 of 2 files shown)


BewareMyPower (changes requested):
Please fix the code style to make `mvn checkstyle:check` pass

wenbingshen:
@BewareMyPower Thanks for your review. Address your comment. PTAL :)

BewareMyPower (approved):
LGTM

BewareMyPower:
Currently there's a flaky test. I'll try to rerun for at most 2 times. If it still didn't pass, I'll merge this PR first.

PR #654, opened 2021-08-14
Fix KoP will cause Kafka Errors.REQUEST_TIMED_OUT when consume multi TopicPartition in one consumer request

### LINKED ISSUE #604: [BUG] KoP will cause Kafka Errors.REQUEST_TIMED_OUT when consume multi TopicPartition in one comsume request
**Describe the bug**
When consume multi TopicPartition in one request, in MessageFetchContext.handleFetch(), tryComplete() may enter race condition

When one topicPartition removed from responseData, maybe removed again, which will cause Errors.REQUEST_TIMED_OUT

wenbingshen:
Fixes #604 

### Motivation
When consume multi TopicPartition in one request, in MessageFetchContext.handleFetch(), tryComplete() may enter race condition
When one topicPartition removed from responseData, maybe removed again, which will cause Errors.REQUEST_TIMED_OUT

ReadEntries and CompletableFuture.complete operations for each partition are all performed by BookKeeperClientWorker- Different threads in the OrderedExecutor thread pool are executed. When the partition can read data, because the read data and decode operations will take uncertain time, the competition in this case is relatively weak; and when the partition has no data to write, and the consumer After all the data has been consumed, I have been making empty fetch requests, which can be reproduced stably at this time.
Stable steps to reproduce:

A single broker has two partition leaders for one topic;
The topic is not writing data, and consumers have consumed the old data;
At this time, the consumer client continues to send Fetch requests to broker;
Basically, you will soon see that the server returns error_code=7, and the client will down。
1、One fetch request, two partitions, and two threads. The data obtained is an empty set without any protocol conversion operation.
2、When the BookKeeperClientWorker-OrderedExecutor-25-0 thread adds test_kop_222-1 to the responseData, BookKeeperClientWorker-OrderedExecutor-23- 0 thread adds test_kop_222-3 to responseData,
3、at this time responseData.size() >= fetchRequest.fetchData().size(), because tryComplete has no synchronization operation, two threads enter at the same time,
4、fetchRequest.fetchData().keySet() .forEach two threads traverse at the same time, resulting in the same partition multiple times responseData.remove(topicPartition), partitionData is null and cause the REQUEST_TIMED_OUT error.

image

### Modifications
`MessageFetchContext.tryComplete` add synchronization lock

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
@@ -110,6 +110,24 @@ public final class MessageFetchContext {
         return context;
     }
 
+    //only used for unit test
+    public static MessageFetchContext getForTest(FetchRequest fetchRequest,
+                                          CompletableFuture<AbstractResponse> resultFuture) {
+        MessageFetchContext context = RECYCLER.get();
+        context.responseData = new ConcurrentHashMap<>();
+        context.decodeResults = new ConcurrentLinkedQueue<>();
+        context.requestHandler = null;
+        context.maxReadEntriesNum = 0;
+        context.topicManager = null;
+        context.statsLogger = null;
+        context.tc = null;
+        context.clientHost = null;
+        context.fetchRequest = fetchRequest;
+        context.header = null;
+        context.resultFuture = resultFuture;
+        return context;
+    }
+
     private MessageFetchContext(Handle<MessageFetchContext> recyclerHandle) {
         this.recyclerHandle = recyclerHandle;
     }
@@ -128,6 +146,18 @@ public final class MessageFetchContext {
         header = null;
         resultFuture = null;
         recyclerHandle.recycle(this);
+    }
+
+    //only used for unit test
+    public void addErrorPartitionResponseForTest(TopicPartition topicPartition, Errors errors) {
+        responseData.put(topicPartition, new PartitionData<>(
+                errors,
+                FetchResponse.INVALID_HIGHWATERMARK,
+                FetchResponse.INVALID_LAST_STABLE_OFFSET,
+                FetchResponse.INVALID_LOG_START_OFFSET,
+                null,
+                MemoryRecords.EMPTY));
+        tryComplete();
     }
 
     private void addErrorPartitionResponse(TopicPartition topicPartition, Errors errors) {
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
@@ -39,6 +39,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.bookkeeper.common.util.MathUtils;
@@ -90,6 +91,7 @@ public final class MessageFetchContext {
     private FetchRequest fetchRequest;
     private RequestHeader header;
     private volatile CompletableFuture<AbstractResponse> resultFuture;
+    private AtomicBoolean hasComplete;
 
     // recycler and get for this object
     public static MessageFetchContext get(KafkaRequestHandler requestHandler,
@@ -107,6 +109,7 @@ public final class MessageFetchContext {
         context.fetchRequest = (FetchRequest) kafkaHeaderAndRequest.getRequest();
         context.header = kafkaHeaderAndRequest.getHeader();
         context.resultFuture = resultFuture;
+        context.hasComplete = new AtomicBoolean(false);
         return context;
     }
 
@@ -125,6 +128,7 @@ public final class MessageFetchContext {
         context.fetchRequest = fetchRequest;
         context.header = null;
         context.resultFuture = resultFuture;
+        context.hasComplete = new AtomicBoolean(false);
         return context;
     }
 
@@ -145,6 +149,7 @@ public final class MessageFetchContext {
         fetchRequest = null;
         header = null;
         resultFuture = null;
+        hasComplete = null;
         recyclerHandle.recycle(this);
     }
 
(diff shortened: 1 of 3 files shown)


wenbingshen:
@BewareMyPower Hi, Yunze brother, how about your think of this issue? PTAL. :)

tests/src/test/java/io/streamnative/pulsar/handlers/kop/MessageFetchContextTest.java
```diff
@@ -0,0 +1,214 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.streamnative.pulsar.handlers.kop;
+
+import static org.apache.kafka.common.protocol.CommonFields.THROTTLE_TIME_MS;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
```
BewareMyPower, line 18: We use TestNG in KoP.

BewareMyPower (changes requested):

BewareMyPower (changes requested):
Great find. Just leave some suggestions.
1. We use TestNG in KoP, please replace JUnit with TestNG.
2. `MessageFetchContextTest` doesn't rely on the `KopProtocolHandlerTestBase`, so it should be put under `kafka-impl/tests` instead of the `tests/` directory.
3. `MessageFetchContextTest` doesn't test `MessageFetchContext` directly, it looks like a simulation. i.e, if `MessageFetchContext` was modified in future, this test would still pass so it couldn't protect the code.

wenbingshen:
@BewareMyPower Thanks for your review very much. I will modify it to TestNG and move it to the correct directory, I will consider modifying the test cases to protect the code. Thanks. :)

wenbingshen:
@BewareMyPower Thanks for your review very much. I will modify it to TestNG and move it to the correct directory, I will consider modifying the test cases to protect the code. Thanks. :)

wenbingshen:
@BewareMyPower Hi, Yunze brother, I have addressed your comment. PTAL :)

BewareMyPower:
It looks like your test cannot pass, see `
kop mvn build check and kafka-impl test / build (pull_request)`.

```
[INFO] Running io.streamnative.pulsar.handlers.kop.MessageFetchContextTest
Error: The operation was canceled.
```

BewareMyPower:
If it's not convenient to add the test, it will be acceptable to not add the test.

The race condition happens when `complete()` method is called by two threads. It's because `responseData.size()` and `responseData.add()` are not an atomic operation.

wenbingshen:
@BewareMyPower 
Take two partitions as an example, after tryComplete adds the synchronization lock,
The first situation:
If after the first thread responseData.put, the second thread immediately follows responseData.put, and the first thread just enters tryComplete, the judgment (responseData.size() >= fetchRequest.fetchData().size()) condition is satisfied , The first thread enters complete, and finally recycles many object resources, which will set responseData to null, and the second thread enters tryComplete will cause a null pointer exception;
The second case:
If after the first thread responseData.put, the first thread enters tryComplete first, and the judgment (responseData.size() >= fetchRequest.fetchData().size()) condition is not met, the first thread will not enter complete , And at this time the second thread responseData.put, (responseData.size() >= fetchRequest.fetchData().size()) conditions are met, and complete is executed, and everything is executed normally at this time.

Therefore, we need to make a null pointer judgment on responseData in tryComplete.

image

image

BewareMyPower:
You're right, I think

```
        if (resultFuture == null) {
            // the context has been recycled
            return;
        }
```

in `complete()` method should be moved to `tryComplete()`.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
```diff
@@ -171,9 +176,11 @@ private void addErrorPartitionResponse(TopicPartition topicPartition, Errors err
         tryComplete();
     }
 
-    private synchronized void tryComplete() {
-        if (responseData.size() >= fetchRequest.fetchData().size()) {
-            complete();
+    private void tryComplete() {
+        if (responseData != null && responseData.size() >= fetchRequest.fetchData().size()) {
```
BewareMyPower, line 180: The null check for `responseData` here might not be thread safe without `synchronized`. Should we add `volatile` keyword to `responseData`?

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
wenbingshen: @BewareMyPower good, I replace responseData with resultFuture to judge the null pointer exception, because resultFuture with volatile

wenbingshen:
@BewareMyPower for Codacy Static Code Analysis error, what should I do, please teach me. :)

wenbingshen:
@BewareMyPower All checks have passed, PTAL :)

BewareMyPower:
The test can be simplified. `MessageFetchContextTest` compares `MessageFetchContext` and `MessageFetchContextTest` itself. When `isSafe` is false, `addErrorPartitionResponse` only operates on the fields of `MessageFetchContextTest`, this behavior is not related to any code.

IMO, only `testHandleFetchSafe` is required because if `MessageFetchContext` is not thread safe, this test will fail. We don't need to add a contrast implementation to show what the wrong implementation is.

wenbingshen:
> The test can be simplified. `MessageFetchContextTest` compares `MessageFetchContext` and `MessageFetchContextTest` itself. When `isSafe` is false, `addErrorPartitionResponse` only operates on the fields of `MessageFetchContextTest`, this behavior is not related to any code.
> 
> IMO, only `testHandleFetchSafe` is required because if `MessageFetchContext` is not thread safe, this test will fail. We don't need to add a contrast implementation to show what the wrong implementation is.

@BewareMyPower Gotcha! I have addressed your comment. PTAL

BewareMyPower:
I still have questions about the test. When I modified the `tryComplete()` back to the original implementation:

```java
    private void tryComplete() {
        if (responseData.size() >= fetchRequest.fetchData().size()) {
            complete();
        }
    }
```

The test can still pass in my local env.

wenbingshen:
@BewareMyPower I set invocationCount = 5000 locally and copied REQUEST_TIMED_OUT once and this.responseData is empty three times. I didn’t set invocationCount = 5000 in pr because I think if the code has concurrency problems, it will definitely become a flaky test in future code submissions. , What do you think should be done?

image

image

image

BewareMyPower:
Okay, it makes sense.

BewareMyPower (approved):

Issue #236, opened 2020-11-14
[FEATURE] Separate authorization from authentication


BewareMyPower:
**Is your feature request related to a problem? Please describe.**
KoP checks permissions only during authentication. Some tests of `SaslPlainTest` relies on the behavior. But it means once a client passed the authentication, it would have all permissions no matter what the specific permissions real it have.

**Describe the solution you'd like**
Create an `Authorizer` in `KafkaRequestHandler` which is constructed from info of authentication. Then do the authorization before each request is processed.

PR #657, opened 2021-08-16
[FEATURE] Add Authorizer used for authorization.

Demogorgon314:
## Motivation
#236 Add Authorizer used for authorization.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -42,9 +42,7 @@ import io.streamnative.pulsar.handlers.kop.format.EntryFormatterFactory;
 import io.streamnative.pulsar.handlers.kop.offset.OffsetAndMetadata;
 import io.streamnative.pulsar.handlers.kop.offset.OffsetMetadata;
 import io.streamnative.pulsar.handlers.kop.security.SaslAuthenticator;
-import io.streamnative.pulsar.handlers.kop.security.Session;
 import io.streamnative.pulsar.handlers.kop.security.auth.Authorizer;
-import io.streamnative.pulsar.handlers.kop.security.auth.Resource;
 import io.streamnative.pulsar.handlers.kop.security.auth.SimpleAclAuthorizer;
 import io.streamnative.pulsar.handlers.kop.stats.StatsLogger;
 import io.streamnative.pulsar.handlers.kop.utils.CoreUtils;
@@ -91,7 +89,6 @@ import org.apache.bookkeeper.mledger.impl.PositionImpl;
 import org.apache.commons.lang3.NotImplementedException;
 import org.apache.kafka.common.Node;
 import org.apache.kafka.common.TopicPartition;
-import org.apache.kafka.common.acl.AclOperation;
 import org.apache.kafka.common.errors.CorruptRecordException;
 import org.apache.kafka.common.errors.LeaderNotAvailableException;
 import org.apache.kafka.common.protocol.ApiKeys;
@@ -186,7 +183,6 @@ import org.apache.pulsar.policies.data.loadbalancer.ServiceLookupData;
 public class KafkaRequestHandler extends KafkaCommandDecoder {
     public static final long DEFAULT_TIMESTAMP = 0L;
 
-    private final Authorizer authorizer;
     private final PulsarService pulsarService;
     private final KafkaTopicManager topicManager;
     private final GroupCoordinator groupCoordinator;
@@ -196,6 +192,7 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
     private final ScheduledExecutorService executor;
     private final PulsarAdmin admin;
     private final SaslAuthenticator authenticator;
+    private final Authorizer authorizer;
     private final AdminManager adminManager;
 
     private final Boolean tlsEnabled;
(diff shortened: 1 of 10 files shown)


BewareMyPower:
Please also make Codacy analysis pass.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/Session.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package io.streamnative.pulsar.handlers.kop.security;
+
+
+import lombok.Data;
+
+/**
+ * client auth session.
+ */
+@Data
```
BewareMyPower, line 22: We can use `@AllArgsConstructor` here to avoid defining the constructor explicitly.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/KafkaPrincipal.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public KafkaPrincipal(String principalType, String name) {
+        this.principalType = requireNonNull(principalType, "Principal type cannot be null");
+        this.name = requireNonNull(name, "Principal name cannot be null");
+    }
+
+    @Override
+    public String getName() {
+        return name;
+    }
```
BewareMyPower, line 49: Use `@Getter` and `@AllArgsConstructor` since we prefer to use lombok to reduce code.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/auth/Resource.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+package io.streamnative.pulsar.handlers.kop.security.auth;
+
+
+import lombok.Getter;
+
+/**
+ * The Authorization resource.
+ */
+public class Resource {
```
BewareMyPower, line 22: We can add `@Getter` as class annotation to avoid adding `@Getter` for each field. In addition, use `@AllArgsConstructor` here. If you prefer to use factory method `Resource#of` than the constructor, please make the constructor private.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/auth/SimpleAclAuthorizer.java
```diff
@@ -0,0 +1,147 @@
...
+                                    }
+                                }
+                                permissionFuture.complete(false);
+                            }).exceptionally(ex -> {
+                                log.warn("Client with Principal - {} failed to get permissions for resource - {}. {}",
+                                        principal, topicName, ex.getMessage());
+                                permissionFuture.completeExceptionally(ex);
+                                return null;
+                            });
+                } catch (Exception e) {
```
BewareMyPower, line 95: Don't catch `Exception`, catch specific exception types, which are usually checked exceptions, but sometimes catching unchecked exceptions is also acceptable.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -181,6 +183,7 @@
 public class KafkaRequestHandler extends KafkaCommandDecoder {
     public static final long DEFAULT_TIMESTAMP = 0L;
 
+    private final Authorizer authorizer;
```
BewareMyPower, line 186: If's better to keep the same variable order with the order that variables are initialized in the constructor, i.e. put `authorizer` after `authenticator`.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/auth/SimpleAclAuthorizer.java
```diff
@@ -0,0 +1,147 @@
...
+        return CompletableFuture.completedFuture(role != null && superUserRoles.contains(role));
+    }
+
+    @Override
+    public CompletableFuture<Boolean> canLookupAsync(KafkaPrincipal principal, Resource resource) {
+        CompletableFuture<Boolean> canLookupFuture = new CompletableFuture<>();
+        authorize(principal, AuthAction.produce, resource)
+                .whenComplete((hasProducePermission, ex) -> {
+                    if (ex != null) {
+                        log.warn("Check produce permission error:{}", ex.getMessage());
```
BewareMyPower, line 117: I think we should change the log level to debug, because each time a consumer without produce permission connected to KoP, it would always reach here and there could be a lot of warnings.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/auth/SimpleAclAuthorizer.java
BewareMyPower: Sorry I'm wrong that it won't reach here because `ex` will be null and `hasProducePermission` will be false. Resolve this conversation.

BewareMyPower:
I've left some comments, PTAL.

Demogorgon314:
> I've left some comments, PTAL.

Thx, I've fixed, PTAL.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/KafkaPrincipal.java
```diff
@@ -0,0 +1,41 @@
...
+    public static final String USER_TYPE = "User";
+
+    @Getter
+    private final String principalType;
+
+    /**
+     * Pulsar role.
+     */
+    @Getter
+    private final String name;
```
BewareMyPower, line 39: Add @Getter to `KafkaPrincipal` instead of each field.

```java
@Getter
public class A {
    private final String field1;
    private final String field2;
    // ...
}
```

is better than

```java
public class A {
    @Getter
    private final String field1;
    @Getter
    private final String field2;
    // ...
}
```

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/auth/ResourceType.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * Parse the given string as an ACL resource type.
+     *
+     * @param str    The string to parse.
+     *
+     * @return       The ResourceType, or UNKNOWN if the string could not be matched.
+     */
+    public static ResourceType fromString(String str) throws IllegalArgumentException {
```
BewareMyPower, line 57: `IllegalArgumentException` is caught in this method but it's still marked as the checked exception, please remove the checked exception.

BewareMyPower (changes requested):

BewareMyPower:
Overall LGTM, left a comment, PTAL.

BewareMyPower (approved):

PR #660, opened 2021-08-18
[FEATURE] Add fetch down converted when entryFormat=kafka

### LINKED ISSUE #656: [BUG] Consumer report: Record batch for partition topic-3 at offset xxx is invalid, cause: Record is corrupt
**Describe the bug**
broker.conf:
entryFormat=kafka

Multiple producers use different magic versions to write data to the same topic at the same time.

Consumers using higher or lower versions of magic will report errors: 
Record batch for partition topic-3 at offset xxx is invalid, cause: Record is corrupt

In the case of entryFormat=kafka, kop will not process whether the actual magic of the message in the bookkeeper matches the magic supported by the client, which causes the client to be unable to resolve the version of magic that it does not support.

This bug belongs to the compatibility issue of the old version of kafka with kop.

wenbingshen:
Fixes #656 

### Motivation
When entryFormat=kafka

If SASL/PLAIN authentication is not turned on, so kop does not limit the version of the producer client. When a lower version producer writes a message of magic=0 or magic=1, it will cause the higher version consumer to check the record error and the consumer client will be down.

### Modifications
When entryFormat=kafka, increase KafkaEntryFormatter.decode to check batch.magic and client magic of kafka records. When batch.magic is higher than client magic, perform down conversion.

This pr is part of the support for the lower version of Kafka less than 0.11.x. Since the FETCH version is still 4, it is also a bugfix for the higher version consumer to read the lower version of the magic record verification error.


wenbingshen:
@BewareMyPower For the unit test, I reused BasicEndToEndTestBase, because after adding the kafka-0-10 module, the lower version producer will write a message of magic=1. If the error mentioned in the pr is not fixed, the higher version consumer will report an error and the unit test will fail. For down-conversion, due to some known reasons, the FETCH version is still limited to 4, so after I removed 4, the test lower version consumers can successfully read the higher version data, but I did not release the restriction on the FETCH version=4 to this pr. PTAL.

wenbingshen:
@BewareMyPower Flaky test seems to be related to the new code citations I introduced after merging the master branch. I will investigate, and please review this pr first.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatter.java
```diff
@@ -61,15 +65,40 @@ public DecodeResult decode(List<Entry> entries, byte magic) {
...
+                    // batch magic greater than the magic corresponding to the version requested by the client
+                    // need down converted
+                    if (batchMagic > magic) {
+                        MemoryRecords memoryRecords = MemoryRecords.readableRecords(ByteBufUtils.getNioBuffer(byteBuf));
+                        //down converted, batch magic will be set to client magic
+                        convertedRecords = memoryRecords.downConvert(magic, startOffset, time);
+                        log.trace("[{}:{}] downConvert record, start offset {}, entry magic: {}, client magic: {}"
+                                , entry.getLedgerId(), entry.getEntryId(), startOffset, batchMagic, magic);
+                    }
+                    if (convertedRecords != null) {
```
BewareMyPower, line 88: Should we remove this check and move the block of `convertedRecords != null` into `batchMagic > magic` block?

wenbingshen: @BewareMyPower Thanks for your review, You are right, I have addressed your comment, PTAL

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatter.java
```diff
@@ -61,15 +65,42 @@ public DecodeResult decode(List<Entry> entries, byte magic) {
...
+                            if (!optionalByteBufs.isPresent()) {
+                                optionalByteBufs = Optional.of(new ArrayList<>());
+                            }
+                            optionalByteBufs.ifPresent(byteBufs -> byteBufs.add(byteBuf));
+                            optionalByteBufs.ifPresent(byteBufs -> byteBufs.add(kafkaBuffer));
+                            log.trace("[{}:{}] down convertedRecords not null {}, {}, {}"
+                                    , entry.getLedgerId(), entry.getEntryId(), startOffset, batchMagic, magic);
+                        } else {
+                            throw new IOException("DownConvert failed, convertedRecords is null");
+                        }
```
BewareMyPower, line 99: AFAIK, `downConvert` never returns null. Even if it could return null, we should skip this entry because it means this entry is corrupted. If an exception was thrown for a corrupted entry, the exception would eventually be caught outside. However, the `entries` wouldn't be released in this case. See how we process the `KoPMessageMetadataNotFoundException`:

```java
            } catch (KoPMessageMetadataNotFoundException | IOException e) { // skip failed decode entry
                log.error("[{}:{}] Failed to decode entry. ", entry.getLedgerId(), entry.getEntryId(), e);
                entry.release();
            }
```

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatter.java
```diff
@@ -111,7 +110,11 @@ public DecodeResult decode(List<Entry> entries, byte magic) {
                     }
                     optionalByteBufs.ifPresent(byteBufs -> byteBufs.add(kafkaBuffer));
                 }
-            } catch (KoPMessageMetadataNotFoundException | IOException e) { // skip failed decode entry
+                // Almost all exceptions in Kafka inherit from KafkaException and will be captured and processed in KafkaApis.
+                // Here, whether it is down-conversion or the IOException in builder.appendWithOffset in
+                // decodePulsarEntryToKafkaRecords will be caught by Kafka and the KafkaException will be thrown.
+                // So we need to catch KafkaException here.
+            } catch (KoPMessageMetadataNotFoundException | IOException | KafkaException e) { // skip failed decode entry
```
wenbingshen, line 117: Here, whether it is down-conversion or the IOException in builder.appendWithOffset in  decodePulsarEntryToKafkaRecords will be caught by Kafka and the KafkaException will be thrown.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/format/KafkaEntryFormatter.java
wenbingshen: @BewareMyPower Great find. Almost all exceptions in Kafka inherit from KafkaException and will be captured and processed in KafkaApis. We can catch KafkaException here. I updated the code, PTAL

BewareMyPower (approved):

Issue #571, opened 2021-06-15
[FEATURE] Super roles don't need to be granted permissions


BewareMyPower:
**Is your feature request related to a problem? Please describe.**
When KoP is configured with SASL/PLAIN authentication mechanism, it actually uses Pulsar's token authentication. However, for the roles of `superUsersRoles` configuration, they still need to be [grant permissions](http://pulsar.apache.org/docs/en/security-jwt/#authorization) currently.

**Describe the solution you'd like**
Reuse Pulsar's authorization implementation in KoP.

**Describe alternatives you've considered**

**Additional context**

PR #662, opened 2021-08-18
[FEATURE] Add authorization to handle topic metadata request

### LINKED ISSUE #415: [BUG] Change of behavior between KoP 0.3.0 (2.6.2) and 2.7+ using SASL
**Describe the bug**
We noticed a change in behavior between KoP 0.3.0 (2.6.2) and 2.7+ using SASL with token authentication (#206?) 

- In 0.3.0, when multiple clients connected using different SASL credentials (username=tenant/namespace) there was no cross-talk between them: Client1 with SASL1 on TopicA did not see data from Client2 with SASL2 on TopicA. 
- With 2.7.x this has changed, and Client1 and Client2 can freely exchange data.

**To Reproduce**
1. Configure Pulsar and KoP to use SASL w/ token authentication.
2. Run Kafka producers/consumers using different SASL credentials. The Kafka clients will cross-talk (Pulsar clients are not affected by this).

**Expected behavior**
The desired behavior is to have SASL credentials prevent cross-talk between Kafka clients.

**Additional context**
We understand that Kafka is not normally multi-tenant, but we think the behavior seen in 0.3.0 is very useful. We are interested to know what the intentions/goals are here. Is the behavior we rely on with 0.3.0 the intended behavior, a side-effect of the original implementation, etc?

Demogorgon314:
#236 Add authorization to handleTopicMetadataRequest.

Fix #415 and #571 

## Motivation
When client fetch metadata need check topic permission, so we need add authorization in handleTopicMetadataRequest, and do not perform role verification in authentication.

## Modifications
Add a common method in `KafkaRequestHandler#authorize` , this method use `authorizer` to authorization.
Modify the authentication behavior, and do not verify the role during authentication, verify the role in fetch metadata(#571  )

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -434,7 +434,9 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
                             topicMapFuture.completeExceptionally(e);
                             return;
                         }
-
+                        if (topicMapFuture.isCompletedExceptionally()) {
+                            return;
+                        }
                         for (String topic : topics) {
                             final TopicName topicName = TopicName.get(topic);
                             final String key = topicName.getPartitionedTopicName();
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -42,7 +42,10 @@ import io.streamnative.pulsar.handlers.kop.format.EntryFormatterFactory;
 import io.streamnative.pulsar.handlers.kop.offset.OffsetAndMetadata;
 import io.streamnative.pulsar.handlers.kop.offset.OffsetMetadata;
 import io.streamnative.pulsar.handlers.kop.security.SaslAuthenticator;
+import io.streamnative.pulsar.handlers.kop.security.Session;
 import io.streamnative.pulsar.handlers.kop.security.auth.Authorizer;
+import io.streamnative.pulsar.handlers.kop.security.auth.Resource;
+import io.streamnative.pulsar.handlers.kop.security.auth.ResourceType;
 import io.streamnative.pulsar.handlers.kop.security.auth.SimpleAclAuthorizer;
 import io.streamnative.pulsar.handlers.kop.stats.StatsLogger;
 import io.streamnative.pulsar.handlers.kop.utils.CoreUtils;
@@ -89,6 +92,7 @@ import org.apache.bookkeeper.mledger.impl.PositionImpl;
 import org.apache.commons.lang3.NotImplementedException;
 import org.apache.kafka.common.Node;
 import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.acl.AclOperation;
 import org.apache.kafka.common.errors.CorruptRecordException;
 import org.apache.kafka.common.errors.LeaderNotAvailableException;
 import org.apache.kafka.common.protocol.ApiKeys;
@@ -417,7 +421,8 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
     //     persistent://public/default/my-topic-partition-0
     //     persistent://public/default/my-topic-partition-1
     //     persistent://public/default/my-topic-partition-2
-    private void getAllTopicsAsync(CompletableFuture<Map<String, List<TopicName>>> topicMapFuture) {
+    private CompletableFuture<Map<String, List<TopicName>>> getAllTopicsAsync() {
+        CompletableFuture<Map<String, List<TopicName>>> topicMapFuture = new CompletableFuture<>();
         final Map<String, List<TopicName>> topicMap = new ConcurrentHashMap<>();
         final AtomicInteger pendingNamespacesCount = new AtomicInteger(allowedNamespaces.size());
 
@@ -427,9 +432,6 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
                         if (e != null) {
                             log.error("Failed to getListOfPersistentTopic of {}", namespace, e);
                             topicMapFuture.completeExceptionally(e);
-                            return;
-                        }
-                        if (topicMapFuture.isDone()) {
                             return;
                         }
 
@@ -446,6 +448,7 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
                         }
                     });
         }
+        return topicMapFuture;
     }
 
     protected void handleTopicMetadataRequest(KafkaHeaderAndRequest metadataHar,
@@ -470,14 +473,38 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
         //      e.g. <topic1, {persistent://public/default/topic1-partition-0,...}>
         //   1. no topics provided, get all topics from namespace;
         //   2. topics provided, get provided topics.
-        CompletableFuture<Map<String, List<TopicName>>> pulsarTopicsFuture = new CompletableFuture<>();
+        CompletableFuture<Map<String, List<TopicName>>> pulsarTopicsFuture;
 
         if (topics == null || topics.isEmpty()) {
             // clean all cache when get all metadata for librdkafka(<1.0.0).
             KafkaTopicManager.clearTopicManagerCache();
-            // get all topics
-            getAllTopicsAsync(pulsarTopicsFuture);
+            // get all topics, filter by permissions.
+            pulsarTopicsFuture = getAllTopicsAsync().thenApply((allTopicMap)->{
+                final Map<String, List<TopicName>> topicMap = new ConcurrentHashMap<>();
+                allTopicMap.forEach((topic, list) -> {
+                   list.forEach((topicName -> {
+                       authorize(AclOperation.DESCRIBE, Resource.of(ResourceType.TOPIC, topicName.toString()))
+                               .whenComplete((authorized, ex) -> {
+                                   if (ex != null || !authorized) {
+                                       allTopicMetadata.add(new TopicMetadata(
+                                               Errors.TOPIC_AUTHORIZATION_FAILED,
+                                               topic,
+                                               isInternalTopic(topicName.toString()),
+                                               Collections.emptyList()));
+                                       return;
+                                   }
+                                   topicMap.computeIfAbsent(
+                                           topic,
+                                           ignored -> Collections.synchronizedList(new ArrayList<>())
+                                   ).add(topicName);
+                               });
+                   }));
+                });
+
+                return topicMap;
+            });
         } else {
+            pulsarTopicsFuture = new CompletableFuture<>();
             // get only the provided topics
             final Map<String, List<TopicName>> pulsarTopics = new ConcurrentHashMap<>();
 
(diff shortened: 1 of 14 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -429,9 +434,6 @@ private void getAllTopicsAsync(CompletableFuture<Map<String, List<TopicName>>> t
                             topicMapFuture.completeExceptionally(e);
                             return;
                         }
-                        if (topicMapFuture.isDone()) {
-                            return;
-                        }
```
BewareMyPower, line 434: Why remove this check? This check intended to skip the followed logic if the `topicMapFuture` was completed exceptionally.

Demogorgon314: Maybe we should use "isCompletedExceptionally" to make it clearer?

> Why remove this check? This check intended to skip the followed logic if the `topicMapFuture` was completed exceptionally.

Maybe we should use "isCompletedExceptionally" to make it clearer?

BewareMyPower: Agree.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -504,76 +530,116 @@ protected void handleTopicMetadataRequest(KafkaHeaderAndRequest metadataHar,
...
+                        if (ex != null) {
+                            // Authentication failed
+                            completeOneAuthFailedTopic.accept(topic, fullTopicName);
+                            return;
+                        }
+                        if (!authorized) {
+                            // Permission denied
+                            completeOneAuthFailedTopic.accept(topic, fullTopicName);
+                            return;
+                        }
```
BewareMyPower, line 556: Add logs here to print the error message.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/auth/SimpleAclAuthorizer.java
```diff
@@ -138,12 +241,25 @@ protected PulsarService getPulsarService() {
...
 
+    @Override
+    public CompletableFuture<Boolean> canCreateTopicAsync(KafkaPrincipal principal, Resource resource) {
+        checkArgument(resource.getResourceType() == ResourceType.TOPIC
+                || resource.getResourceType() == ResourceType.NAMESPACE,
+                String.format("Expected resource type is TOPIC or NAMESPACE, but have [%s]",
+                        resource.getResourceType()));
+        return authorize(principal, AuthAction.packages, resource);
+    }
+
```
BewareMyPower, line 264: `AuthAction.packages` is not for this case. I think we should not authorize for the `CreateTopic` event. In Pulsar, the topic auto creation is only determined by the config or namespace policy.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
Demogorgon314: In `SimpleAclAuthorizer#128` already have error message, do we need add error message here ?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -470,14 +473,37 @@ protected void handleTopicMetadataRequest(KafkaHeaderAndRequest metadataHar,
...
             KafkaTopicManager.clearTopicManagerCache();
-            // get all topics
-            getAllTopicsAsync(pulsarTopicsFuture);
+            // get all topics, filter by permissions.
+            pulsarTopicsFuture = getAllTopicsAsync().thenApply((allTopicMap) -> {
+                final Map<String, List<TopicName>> topicMap = new ConcurrentHashMap<>();
+                allTopicMap.forEach((topic, list) -> {
+                   list.forEach((topicName ->
+                           authorize(AclOperation.DESCRIBE, Resource.of(ResourceType.TOPIC, topicName.toString()))
+                           .whenComplete((authorized, ex) -> {
```
BewareMyPower, line 487: It's same as my previous comment. There's no need to authorize `DESCRIBE` operation here.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
BewareMyPower: Yes. And I think we should not add error message to the API that returns a future. Though some existing code doesn't obey this rule.

When we design a API, it should be like:

```java
CompletableFuture<T> getSomething(String key) {
    return someService.findAsync(key);
}
```

The caller is responsible to determine which logs should be printed:

```java
String key = "xxx";
getSomething(key).whenComplete(/* ... */); // or exceptionally
```

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
BewareMyPower: Sorry it's my fault, `DESCRIBE` should be authorized. Solve this comment.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/auth/ResourceType.java
```diff
@@ -34,6 +34,13 @@
      * A Pulsar topic.
      */
     TOPIC((byte) 2),
+
+    /**
+     * A Pulsar namespace.
+     */
+    NAMESPACE((byte) 3),
```
BewareMyPower, line 41: It seems that in this PR there's no place that a `Resource` object's type is `NAMESPACE`? If so, we should remove this change and add it in the followed PR.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
BewareMyPower: A exception case is, if we want to swallow the exception and complete the future with `null` instead of the exception. We must print the exception in the API implementation.

BewareMyPower:
My review is completed, PTAL. The main problem is that we should not authorize the `CreateTopic` operation. 

I also add the `doc-required` label because we may update the `docs/security.md` later.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/auth/SimpleAclAuthorizer.java
Demogorgon314: In kafka have authorized create topic event, don't we need this feature ?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
Demogorgon314: Ok, I will change `SimpleAclAuthorizer#128` logs level to debug, and add log when authorize failed instead.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/auth/SimpleAclAuthorizer.java
BewareMyPower: KoP uses Pulsar's authorization provider to grant permissions, so we should still use Pulsar's way. And Pulsar doesn't perform authorization for create topic event.

This permission is mainly for Kafka `AdminClient#createTopics` method, which is similar to Pulsar's admin operation (`PulsarAdmin#createPartitionedTopic`). In this case, we should only allow [roles with admin permission](https://pulsar.apache.org/docs/en/security-authorization/#administer-tenants.) to create topics.

Regarding to the auto topic creation case for Kafka client (not admin), we should still use configuration to control whether it's allowed, just like what Pulsar does.

Demogorgon314: Agree, I will use Pulsar's way.

Demogorgon314:
> My review is completed, PTAL. The main problem is that we should not authorize the `CreateTopic` operation.
> 
> I also add the `doc-required` label because we may update the `docs/security.md` later.

Thanks for review, I will update docs later.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/CustomOAuthBearerCallbackHandlerTest.java
```diff
@@ -78,6 +80,11 @@ protected void setup() throws Exception {
         conf.setKopOauth2AuthenticateCallbackHandler(MockedAuthenticateCallbackHandler.class.getName());
 
         super.internalSetup();
+        admin.namespaces().grantPermissionOnNamespace(
+                tenant + "/" + namespace,
+                USER,
+                Sets.newHashSet(AuthAction.consume, AuthAction.produce, AuthAction.packages)
```
BewareMyPower, line 86:
```suggestion
                Sets.newHashSet(AuthAction.consume, AuthAction.produce)
```

BewareMyPower (changes requested):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandlerWithAuthorizationTest.java
```diff
@@ -0,0 +1,245 @@
...
+                .authorize(eq(AclOperation.DESCRIBE), eq(Resource.of(ResourceType.TOPIC, topic)));
+        final RequestHeader header = new RequestHeader(ApiKeys.METADATA, version, "client", 0);
+        final MetadataRequest request =
+                new MetadataRequest(Collections.singletonList(topic), true, version);
+        final CompletableFuture<AbstractResponse> responseFuture = new CompletableFuture<>();
+        spyHandler.handleTopicMetadataRequest(
+                new KafkaCommandDecoder.KafkaHeaderAndRequest(
+                        header, request, PulsarByteBufAllocator.DEFAULT.heapBuffer(), null),
+                responseFuture);
+         final MetadataResponse response = (MetadataResponse) responseFuture.get();
```
BewareMyPower, line 201: This line is not aligned, but I don't know why checkstyle plugin doesn't point out this.

BewareMyPower (changes requested):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandlerWithAuthorizationTest.java
Demogorgon314: Yes, it's strange.

BewareMyPower (approved):

eolivelli:
This implementation is probably wrong.

I am working on the KOP Proxy and this is the test that fails using the Proxy

https://github.com/datastax/kop/blob/29be05f851c8c6b506703b524fb28d3adecea693/test[…]treamnative/pulsar/handlers/kop/KafkaAuthorizationTestBase.java

In the test we grant AuthAction.consume, AuthAction.produce and the the topic is "visible" using `listTopics` to ANOTHER_USER.

When you use the KOP proxy we use `PulsarAdmin` to implement listTopics and the user is not able to "see" the topic, because with Pulsar REST API you need tenant admin role to get the partitions for a topic.

I have filed this issue on Pulsar
https://github.com/apache/pulsar/issues/11945

we have to have a consistent behaviour in KOP

Demogorgon314:
@eolivelli You're right, the behaviour is not consistent. Current pulsar don't support Topic Level permissions to listTopics, IMO, if we have the topics permissions, we should be able to list the topic?

Issue #669, opened 2021-08-20
[FEATURE] Support multiple listeners by using advertisedListeners


wangjialing218:
**Is your feature request related to a problem? Please describe.**
Here provide a design for #574, could also be a solution for #633 

Suppose we deploy kop with pulsar on a host which have two NIC, such as a cloud ECS. 192.168.1.10 is the address for private network inside the cloud and 172.16.1.10 is the address for public network.
We start pulsar endpoints at 6650, 6651 on both NIC for pulsar, and start endpoints at 9092, 9093, 19002, 19003 for kafka.

We would config like this for pulsar:
> internalListenerName=pulsar_private
> advertisedListeners=pulsar_private:pulsar://192.168.1.10:6650,pulsar_private_ssl:pulsar+ssl://192.168.1.10:6651,pulsar_public:pulsar://172.16.1.10:6650,pulsar_pulic_ssl:pulsar+ssl://172.16.1.10:6651

We would config like this for kafka:

> listeners=PRIVATE://0.0.0.0:9092,PRIVATE_SSL://0.0.0.0:9093,PUBLIC://0.0.0.0:19002,PUBLIC_SSL:0.0.0.0:19003
> advertised.listeners=PRIVATE://192.168.1.10:9092,PRIVATE_SSL://192.168.1.10:9093,PUBLIC://172.16.1.10:19002,PUBLIC_SSL:172.16.1.10:19003
> listener,security.protocal.map=PRIVATE:PLAINTEXT,PRIVATE_SSL:SSL,PUBLIC:PLAINTEXT,PUBLIC_SSL:SSL

Kop current use `kafkaAdvertisedListeners` to config advertised address for kafka, but only support one address of one protocol.

**Describe alternatives you've considered**
Deprecated the setting `kafkaAdvertisedListeners`.
remove the function `KafkaRequestHandler.getProtocolDataToAdvertise()`
and config kop like this:

> kafkaListeners=PRIVATE://0.0.0.0:9092,PRIVATE_SSL://0.0.0.0:9093,PUBLIC://0.0.0.0:19002,PUBLIC_SSL:0.0.0.0:19003
> kafkaProtocalMap=PRIVATE:PLAINTEXT,PRIVATE_SSL:SSL,PUBLIC:PLAINTEXT,PUBLIC_SSL:SSL
[...]

PR #670, opened 2021-08-20
skip to get kafka advertised addres from zk when lookup with listenerName

wangjialing218:
This is the first step for #669, only support one listener which configed in `kafkaListenerName`, and only support PLAINTEXT


BewareMyPower:
Could you fix the test?

wangjialing218:
This PR could resolve #633 if only PLAINTEXT is used.

BewareMyPower:
It looks like KafkaListenerNameTest.testListenerName still failed, could you take a look?

wangjialing218:
> It looks like KafkaListenerNameTest.testListenerName still failed, could you take a look?

Lookup has no problem, but kafka client can't connect to `localAddress` through `kafkaBrokerPort`, I changed `localAddress` to "localhost" and the test will be ok.
Maybe we do not need to use `localAddress` in this test.

BewareMyPower:
Let's see what will happen in CI. I used to use `localhost` but it failed in CI so I changed it to `ServiceConfigurationUtils.getDefaultOrConfiguredAddress(null)` and it works.

PR #690, opened 2021-08-30
[FEATURE] Support non-partitioned Topic

### LINKED ISSUE #674: [FEATURE] Support non-partitioned topics.
I believe KoP should support non-partitioned topics. While I understand why only partitioned topics are supported due to offsets, why not treat non partitioned topics as a single partition topic on KoP side?

When attempting to produce/consume/administrate non-partitioned topics, treat them as a single partitioned topic using an abstraction layer.

**Describe alternatives you've considered**
I have not considered any alternatives worth mentioning.

Demogorgon314:
Fix #674

## Motivation
Some users want to use exists non-partitioned topic in KoP, so we need support the feature.

## Modifications
**Note**: Current don't support transaction.

When KoP handle `METADATA` request, if the topic is non-partitioned topic, then treat as a single partitioned topic. And when we `KafkaTopicManager#getTopic`, add a fallback to get non-partitioned topic. Also when namespace `load` or `unload`, we need clear the cache.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
@@ -219,12 +219,12 @@ public class KafkaTopicManager {
         brokerService.getTopicIfExists(topicName).whenComplete((t2, throwable) -> {
             TopicName topicNameObject = TopicName.get(topicName);
             if (throwable != null) {
+                // Failed to getTopic from current broker, remove cache, which added in getTopicBroker.
+                removeTopicManagerCache(topicName);
                 if (topicNameObject.getPartitionIndex() == 0) {
                     log.warn("Get partition-0 error [{}].", throwable.getMessage());
                 } else {
-                    handleGetTopicEx(topicName, topicCompletableFuture, throwable);
-                    // Failed to getTopic from current broker, remove cache, which added in getTopicBroker.
-                    removeTopicManagerCache(topicName);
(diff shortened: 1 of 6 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
```diff
@@ -211,35 +213,73 @@ private Producer registerInPersistentTopic(PersistentTopic persistentTopic) {
...
         brokerService.getTopicIfExists(topicName).whenComplete((t2, throwable) -> {
+            TopicName topicNameObject = TopicName.get(topicName);
             if (throwable != null) {
-                // The ServiceUnitNotReadyException is retriable so we should print a warning log instead of error log
-                if (throwable instanceof BrokerServiceException.ServiceUnitNotReadyException) {
-                    log.warn("[{}] Failed to getTopic {}: {}",
-                            requestHandler.ctx.channel(), topicName, throwable.getMessage());
-                    topicCompletableFuture.complete(Optional.empty());
+                if (topicNameObject.getPartitionIndex() == 0) {
+                    log.warn("Get partition-0 error [{}].", throwable.getMessage());
```
BewareMyPower, line 219: Why not call `removeTopicManagerCache` in this case?

Demogorgon314: It should be call removeTopicManagerCache in this case, fixed.

BewareMyPower (approved):

PR #695, opened 2021-08-31
Fix many non-durable cursors created on a topic with multiple groups

BewareMyPower:
### Motivation
KoP's `KafkaTopicConsumerManager` (aka TCM) maintains some non-durable cursors and the associated offsets for a consumer on a specific topic. If the fetch offset doesn't exist in TCM, TCM will create a non-durable cursor whose position is associated with the offset. Each time a message is consumed, the offset and cursor pair will be removed.

Currently there's a global map `consumerTopicManagers` whose key is topic name and value is the future of TCM. However, for a topic with multiple consumer groups (subscriptions), all consumers share the same TCM. There's a great possibility that different consumers fetch different offsets concurrently from the same TCM. In this case, a lot of non-durable cursors could be created.

### Modifications
Add a singleton class `KafkaTopicConsumerManagerCache` to manage TCMs. The internal cache has two keys. The first key is the topic name, the second key is the remote address to identify different consumers on the same topic.

To ensure only one non-durable cursor is created for a TCM if no reconnection happened, this PR adds a field `numCreatedCursors` to record the total count of non-durable cursor creations. `testCursorCountForMultiGroups` was added to verify this behavior, it creates 5 consumers to consume the same topic with different group id in parallel. After consuming completed, check all TCMs to ensure each TCM's `numCreatedCursors` is 1.


hangc0276 (approved):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaTopicConsumerManagerTest.java
```diff
@@ -344,10 +356,72 @@ public void testOnlyOneCursorCreated() throws Exception {
...
+                .mapToObj(i -> {
+                    final Properties props = newKafkaConsumerProperties();
+                    props.setProperty(ConsumerConfig.GROUP_ID_CONFIG, "group-" + i);
+                    final KafkaConsumer<String, String> consumer = new KafkaConsumer<>(props);
+                    consumer.subscribe(Collections.singleton(topic));
+                    return consumer;
+                }).collect(Collectors.toList());
+
+        final ExecutorService executor = Executors.newFixedThreadPool(numConsumers);
+        final List<Future<?>> futures = new ArrayList<>();
```
Demogorgon314, line 391: We can use `CountDownLatch` to replace `List<Future<?>>`?

Demogorgon314 (approved):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaTopicConsumerManagerTest.java
BewareMyPower: Yeah, it makes code simpler. Addressed.

PR #697, opened 2021-08-31
[BUGFIX] Fix go sarama client with v0 handshake can successful authenticated but will failed fresh metadata

### LINKED ISSUE #696: [BUG] go sarama client with v0 handshake can successful authenticated but will failed fresh metadata
**Describe the bug**
go sarama client config:

config.Net.SASL.Enable = true
config.Net.SASL.User = "public/default"
config.Net.SASL.Password = "xxx"
config.Net.SASL.Version = 0
config.Metadata.Retry.Max = 0

Use this configuration to access the kop cluster with sasl/plain authentication enabled, we will find that the authentication can pass normally, but the metadata update fails:
`
broker.go:932: Successful SASL handshake. Available mechanisms: [PLAIN]
broker.go:1004: SASL authentication successful with broker 10.39.0.58:9996:4 - [0 0 0 4]
broker.go:214: Connected to broker at 10.39.0.58:9996 (unregistered)
client.go:818: client/metadata got error from broker -1 while fetching metadata: kafka: error decoding packet: message of length 0 too large or too small

`
`kafka: error decoding packet: message of length 0 too large or too small`
The log of this error indicates that there is an error parsing the metadata response, and the length of the read message is 0.

**1. Why can the java client pass v0 authentication and correctly parse the metadata request, but sarama cannot pass v0 to complete the authentication?
2. Why is there no problem with kafka broker returning v0 response, but the v0 response returned by kop makes sarama error?**

I investigated related phenomena and came to some conclusions:

Investigation steps:
### For the first question:
**1. I got the v0 handle sasl token response that kop returned to the go sarama client and java client through packet capture, The results all show that 8 bytes of data are sent, the first 4 bytes indicate the length of the message, and the last 4 bytes themselves are the actual data content.**
Screenshot of the wrong packet:
image
image

**The java client can correctly parse the metadata response because the client processes the authenticated response and the metadata protocol request using different channels, even if it receives 8 bytes of data, no matter if the java client parses 4 bytes or all bytes, the sasl handshake authentication can be completed normally, and the subsequent protocol request will be completed through another port established with kop, so there is no residual data in the channel, which does not affect the parse of the normal metadata response.**
image

**But for the go sarama client, whether it is an authentication request or a metadata request after the authentication is completed, the same channel is used. Sarama processes the v0 authentication response and only parses the first 4 bytes of the response data, which results in 4 remaining Dirty bytes of data will interfere with the subsequent parsing of the metadata response, resulting in the parsing metadata response data length being 0.**
image

Screenshot of the correct packet:
image
image

### For the second question:
2. One of the differences between kop and kafka in processing responses is that kop uses netty for network transmission. Kop is configured with LengthFieldPrepender, and 4 bytes are added to indicate the length of the response data before sending the response data, and this is the reason why the Screenshot of the wrong packet 4 more bytes of data.

**To Reproduce**
The current kop master branch, and use sarama and the above configuration to access the sasl/plain kop cluster.

wenbingshen:
fixes #696 

### Motivation
#696  describes a detailed investigation of the problem and root cause of the Go Sarama client's inability to process the metadata response after passing V0 authentication caused by LengthFieldPrepender
### Modifications
After the go sarama client passes the v0 version authenticate, it can process other data requests normally.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
@@ -324,7 +324,7 @@ public class SaslAuthenticator {
         });
     }
 
-    private static ByteBuf sizePrefixed(ByteBuffer buffer) {
+    public static ByteBuf sizePrefixed(ByteBuffer buffer) {
         ByteBuffer sizeBuffer = ByteBuffer.allocate(4);
         sizeBuffer.putInt(0, buffer.remaining());
         ByteBuf byteBuf = Unpooled.buffer(sizeBuffer.capacity() + buffer.remaining());
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
@@ -16,6 +16,7 @@ package io.streamnative.pulsar.handlers.kop.security;
 import static com.google.common.base.Preconditions.checkArgument;
 import static org.apache.kafka.common.protocol.ApiKeys.API_VERSIONS;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.ChannelHandlerContext;
@@ -324,6 +325,7 @@ public class SaslAuthenticator {
         });
     }
 
+    @VisibleForTesting
     public static ByteBuf sizePrefixed(ByteBuffer buffer) {
         ByteBuffer sizeBuffer = ByteBuffer.allocate(4);
         sizeBuffer.putInt(0, buffer.remaining());
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
@@ -328,8 +328,11 @@ public class SaslAuthenticator {
         ByteBuffer sizeBuffer = ByteBuffer.allocate(4);
         sizeBuffer.putInt(0, buffer.remaining());
         ByteBuf byteBuf = Unpooled.buffer(sizeBuffer.capacity() + buffer.remaining());
+        // why we reset writer index? see https://github.com/streamnative/kop/issues/696
+        byteBuf.markWriterIndex();
         byteBuf.writeBytes(sizeBuffer);
         byteBuf.writeBytes(buffer);
+        byteBuf.resetWriterIndex();
         return byteBuf;
     }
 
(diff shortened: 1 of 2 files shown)


wenbingshen:
@BewareMyPower Please check the survey results in #696 . Since we do not have dedicated certification tests for other non-java clients, after the code was repaired, I completed the go sarama and pykafka tests locally. PTAL. Thanks :)

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
```diff
@@ -328,8 +328,11 @@ private static ByteBuf sizePrefixed(ByteBuffer buffer) {
         ByteBuffer sizeBuffer = ByteBuffer.allocate(4);
         sizeBuffer.putInt(0, buffer.remaining());
         ByteBuf byteBuf = Unpooled.buffer(sizeBuffer.capacity() + buffer.remaining());
+        // why we reset writer index? see https://github.com/streamnative/kop/issues/696
+        byteBuf.markWriterIndex();
         byteBuf.writeBytes(sizeBuffer);
         byteBuf.writeBytes(buffer);
+        byteBuf.resetWriterIndex();
```
BewareMyPower, line 335: We can make `sizePrefixed` public and add a related test. For example, pass a `ByteBuffer` with

```java
        final ByteBuffer byteBuffer = ByteBuffer.allocate(8);
        for (int i = 0; i < byteBuffer.capacity(); i++) {
            byteBuffer.put((byte) i);
        }
        byteBuffer.position(4);
```

Then you can verify the reader index, writer index, capacity.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
wenbingshen: @BewareMyPower Thanks for your review. Good suggestion. I have added a test. PTAL :)

BewareMyPower (approved):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
```diff
@@ -324,12 +324,15 @@ private static void sendKafkaResponse(ChannelHandlerContext ctx,
         });
     }
 
-    private static ByteBuf sizePrefixed(ByteBuffer buffer) {
+    public static ByteBuf sizePrefixed(ByteBuffer buffer) {
```
Demogorgon314, line 327: We can add annotation `@VisibleForTesting`, because this method should be private, change to public just for unit test.

wenbingshen: @Demogorgon314 Good suggestion. I have addressed your comment. Thanks.

Demogorgon314 (approved):

PR #708, opened 2021-09-05
fetch.max.bytes and max.partition.fetch.bytes should be support!

lordcheng10:


BewareMyPower:
Please fix the checkstyle and add the tests.

lordcheng10:
> Please fix the checkstyle and add the tests.

I added a unit test and adjusted the code format. @BewareMyPower

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaApisTest.java
```diff
@@ -319,6 +322,43 @@ public void testConsumerListOffsetLatest() throws Exception {
...
+        List<TopicPartition> topicPartitions = new ArrayList<>();
+        TopicPartition tp1 = new TopicPartition(topicName, 0);
+        TopicPartition tp2 = new TopicPartition(topicName, 1);
+        topicPartitions.add(tp1);
+        topicPartitions.add(tp2);
+
+        // producing data and then consuming.
+        KafkaProducer<String, String> kProducer = createKafkaProducer();
+        produceData(kProducer, topicPartitions, 20);
+        KafkaConsumer<String, String> consumer = createKafkaConsumer(5000, 1, maxBytes, maxPartitionBytes);
```
BewareMyPower, line 342: I changed this line to

```java
KafkaConsumer<String, String> consumer = createKafkaConsumer(5000, 1);
```

and run this test in my local env. The test still passed. It might be related to `maxReadEntriesNum` config. Please add a contrast test to prove that with default max bytes, the fetched bytes will be larger than the `maxBytes` and `maxPartitionBytes`.

BewareMyPower (changes requested):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaApisTest.java
lordcheng10: Because the maxBytes and maxPartitionBytes settings are too large, causing maxReadEntriesNum to take effect, so
I reduced maxBytes and maxPartitionBytes to achieve the desired effect： @BewareMyPower 

        int maxBytes = 500;
        int maxPartitionBytes = 100;

BewareMyPower (approved):

Issue #57, opened 2019-12-27
[FEATURE] Proxy support for KoP


jiazhai:
Proxy support for KoP.
- #101 Create sub module in current code structure.

jiazhai:
Currently this is still in our backlog. and not start working on it in our side. If anyone is interested to contribute to this PR welcome to comments on this issue. And Pierre already did some investigation on this, Thanks for @PierreZ's help on this.

eolivelli:
@jiazhai @PierreZ 
is the Envoy proxy able to act like the Pulsar proxy on k8s ?

I mean, in Pulsar we have the "Pulsar proxy" that hides the complexity of the cluster inside k8s.

From the documentation I see that I should setup one Envoy proxy per each Pulsar broker, this is very awkward if you have a large Pulsar cluster, especially in case that the cluster is dynamically scaling up and down.

https://github.com/streamnative/kop/blob/master/docs/envoy-proxy.md

"For each broker with KoP enabled, a dependent Envoy proxy is required. Assuming that you have N brokers whose internal hostname is pulsar-broker-<i>, where i is the broker ID that varies from 0 to N-1."

BewareMyPower:
@eolivelli Yes, not like Pulsar proxy, Envoy acts more like a sidecar. Currently we have no proxy for KoP like Pulsar proxy.

eolivelli:
@BewareMyPower @merlimat @jiazhai @PierreZ 
I have starting working on a Proxy for KOP.
I have also sent a PIP to dev@pulsar.apache.org in order to add protocol handlers in the Pulsar Proxy.

I will be happy to share my prototype and share my work upstream to this repository.

BewareMyPower:
@eolivelli It would be great to have protocol handlers support in the Pulsar Proxy.

eolivelli:
@jiazhai @BewareMyPower @PierreZ @sijie 
I have created this design document for the proxy that I am implementing.
https://github.com/streamnative/kop/issues/717

I have already implemented a working prototype, the performances with the proxy are the same as while using regular Pulsar proxy.
I have also already implemented support for Token Auth.

I will be happy to contribute the prototype to this repository and continue the development directly here and not in my repo.
This way it will be easier to maintain it and also to have more contributions.

PR #710, opened 2021-09-06
Refactor APIs to allow access from the Proxy module

eolivelli:
This patch introduces a little refactor in order to support the implementation of a separate KOP Proxy Module:
- allow to access internal ByteBuf of KafkaHeaderAndRequest, in order to allow zero-copy request magement in the Proxy
- extract some static methods in order to use the same algorithm for mapping coordinators

Relates to #57


BewareMyPower (approved):

eolivelli:
@BewareMyPower all check passed

PR #711, opened 2021-09-06
Add support for passing OriginalPrincipal from Proxy

eolivelli:
- If the authenticated role is a proxyRole use the username as OriginalPrincipal
- remove dependency on "BrokerService" in SaslAuthenticator


BewareMyPower:
Could you fix the checkstyle and codacy check first?

eolivelli:
@BewareMyPower 
The Codacy errors are about a piece of code that I moved.
I agree with Codacy that this is not a safe assignment. But if we want to change it we have to refactor a bit the code

image

eolivelli:
@BewareMyPower all check passed

BewareMyPower (approved):

PR #712, opened 2021-09-06
[FEATURE] Add KopEventManager to repair metadata consistency of cluster and consistency of consumer group status

### LINKED ISSUE #617: [BUG] Describing group whose topic is deleted could cause TimeoutException
**Describe the bug**
If a consumer subscribed a topic with a specific group, after the topic is deleted, describing the group could cause `TimeoutException`.

**To Reproduce**
1. Run a KoP standalone.
2. Describe a group that doesn't exist, the result is expected
```bash
$ ./bin/kafka-consumer-groups.sh --bootstrap-server localhost:9092 --group my-group --describe

Error: Consumer group 'my-group' does not exist.
```
3. Create a topic with a message produced and subscribe it
```bash
$ ./bin/kafka-console-producer.sh --bootstrap-server localhost:9092 --topic my-topic
>1
>
$ ./bin/kafka-console-consumer.sh --bootstrap-server localhost:9092 --topic my-topic --group my-group --from-beginning
1
^CProcessed a total of 1 messages
```
4. Delete the topic
```bash
./bin/kafka-topics.sh --bootstrap-server localhost:9092 --topic my-topic --delete
```
5. Describe the group again, `TimeoutException` happened
```bash
$ ./bin/kafka-consumer-groups.sh --bootstrap-server localhost:9092 --group my-group --describe

Error: Executing consumer group command failed due to org.apache.kafka.common.errors.TimeoutException: Call(callName=metadata, deadlineMs=1626879452007, tries=48, nextAllowedTryMs=1626879452108) timed out at 1626879452008 after 48 attempt(s)
java.util.concurrent.ExecutionException: org.apache.kafka.common.errors.TimeoutException: Call(callName=metadata, deadlineMs=1626879452007, tries=48, nextAllowedTryMs=1626879452108) timed out at 1626879452008 after 48 attempt(s)
	at org.apache.kafka.common.internals.KafkaFutureImpl.wrapAndThrow(KafkaFutureImpl.java:45)
	at org.apache.kafka.common.internals.KafkaFutureImpl.access$000(KafkaFutureImpl.java:32)
	at org.apache.kafka.common.internals.KafkaFutureImpl$SingleWaiter.await(KafkaFutureImpl.java:89)
	at org.apache.kafka.common.internals.KafkaFutureImpl.get(KafkaFutureImpl.java:260)
	at kafka.admin.ConsumerGroupCommand$ConsumerGroupService.getLogEndOffsets(ConsumerGroupCommand.scala:648)
	at kafka.admin.ConsumerGroupCommand$ConsumerGroupService.collectConsumerAssignment(ConsumerGroupCommand.scala:413)
	at kafka.admin.ConsumerGroupCommand$ConsumerGroupService.$anonfun$collectGroupsOffsets$9(ConsumerGroupCommand.scala:595)
	at scala.collection.TraversableLike.$anonfun$flatMap$1(TraversableLike.scala:292)
	at scala.collection.Iterator.foreach(Iterator.scala:943)
	at scala.collection.Iterator.foreach$(Iterator.scala:943)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
	at scala.collection.IterableLike.foreach(IterableLike.scala:74)
	at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
	at scala.collection.AbstractIterable.foreach(Iterable.scala:56)
	at scala.collection.TraversableLike.flatMap(TraversableLike.scala:292)
	at scala.collection.TraversableLike.flatMap$(TraversableLike.scala:289)
	at scala.collection.AbstractTraversable.flatMap(Traversable.scala:108)
	at kafka.admin.ConsumerGroupCommand$ConsumerGroupService.$anonfun$collectGroupsOffsets$2(ConsumerGroupCommand.scala:587)
	at scala.collection.TraversableLike$WithFilter.$anonfun$map$2(TraversableLike.scala:874)
	at scala.collection.mutable.HashMap.$anonfun$foreach$1(HashMap.scala:149)
	at scala.collection.mutable.HashTable.foreachEntry(HashTable.scala:237)
	at scala.collection.mutable.HashTable.foreachEntry$(HashTable.scala:230)
	at scala.collection.mutable.HashMap.foreachEntry(HashMap.scala:44)
	at scala.collection.mutable.HashMap.foreach(HashMap.scala:149)
	at scala.collection.TraversableLike$WithFilter.map(TraversableLike.scala:873)
	at kafka.admin.ConsumerGroupCommand$ConsumerGroupService.collectGroupsOffsets(ConsumerGroupCommand.scala:569)
	at kafka.admin.ConsumerGroupCommand$ConsumerGroupService.describeGroups(ConsumerGroupCommand.scala:369)
	at kafka.admin.ConsumerGroupCommand$.run(ConsumerGroupCommand.scala:76)
	at kafka.admin.ConsumerGroupCommand$.main(ConsumerGroupCommand.scala:63)
	at kafka.admin.ConsumerGroupCommand.main(ConsumerGroupCommand.scala)
Caused by: org.apache.kafka.common.errors.TimeoutException: Call(callName=metadata, deadlineMs=1626879452007, tries=48, nextAllowedTryMs=1626879452108) timed out at 1626879452008 after 48 attempt(s)
Caused by: org.apache.kafka.common.errors.TimeoutException: Timed out waiting for a node assignment. Call: metadata
```

**Expected behavior**
The result of step 5 should be `Error: Consumer group 'my-group' does not exist.`.

wenbingshen:
Fixes #617 

### Motivation
When investigating #617, I found the root cause of TimeOut exception and incorrect consumer group status.

**The first question is a new problem I found while investigating the cause**
**1. For example, when the cluster contains two kop nodes, broker1 is the owner broker of all topics in the current cluster, and broker2 is not the leader of any partition.**

- 1.1. When the client describesGroup, it needs to connect to the configured brokerList first.
- 1.2. Suppose it connects to broker2 to obtain metadata. For the current server-side metadata processing logic, if the leader nodes of all partitions are not broker2, they are all broker1. Since metadata acquisition proceeds normally, the metadata returned by the brokers node only contains broker1.
- 1.3. So next, the client will send FindCoordinatorRequest to broker1. Broker1 calculates based on the consumer group hash and finds that the coordinator is on broker2.
- 1.4. Because the cluster attribute in the client metadata only has the broker1 node, the client can no longer send DescribeGroupRequest to the broker2, and keep sending the update metadata request to broker1 until it times out.

**The second question is described in #617** 
**2. The current GroupCoordinator will not remove the deleted partition from the GroupMetadata after the topic is deleted**

### Modifications
Add a single-threaded event processing queue on the kop server. It will obtain the current cluster node metadata once at startup and place it locally, and will also register to monitor the `/loadbalance/broker` subnode and `/kop/delete_topics` subnode, `/loadbalance /broker` is used to update the metadata of the kop node list, and `/kop/delete_topics` is used to update the GroupMetadata which had these deleted topics, and try to change the state of the consumer group.


wenbingshen:
1. When a new node started, the kop brokers node metadata update is triggered.
image

wenbingshen:
2. Remove the deleted partition from the GroupMetadata and try to update the consumer group status.

image

wenbingshen:
@BewareMyPower There are still some bugs in the code that need to be fixed and related tests added. Before that, can you first take a look at the solution described by this pr? thanks.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
```diff
@@ -0,0 +1,296 @@
...
+            } catch (Exception e) {
+                log.error("Get broker {} ZNode data failed which have an error {}", broker, e.getMessage());
+                e.printStackTrace();
+            }
+        });
+        Collection<? extends Node> oldKopBrokers = adminManager.getBrokers();
+        adminManager.addBrokers(kopBrokers);
+        log.info("Refresh kop brokers new cache {}, old brokers cache {}",
+                adminManager.getBrokers(), oldKopBrokers);
+    }
```
wenbingshen, line 167: Get the address of the kop node under the protocol attribute.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
```diff
@@ -0,0 +1,296 @@
...
+
+                log.info("GroupMetadata delete topics {}, no matching topics {}",
+                        deletedTopics, Sets.difference(topicsFullNameDeletionsSets, deletedTopics));
+
+            } catch (Exception e) {
+                log.error("DeleteTopicsEvent process have an error {}", e.getMessage());
+                e.printStackTrace();
+            }
+        }
+    }
```
wenbingshen, line 282: Remove partitions in consumer groups for partitions which have been deleted.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
```diff
@@ -0,0 +1,296 @@
...
+        @Override
+        public void process() {
+            try {
+                getBrokers(kopZkClient.getBrokers());
+            } catch (Exception e) {
+                log.error("BrokersChangeEvent process have an error {}", e.getMessage());
+                e.printStackTrace();
+            }
+        }
+    }
```
wenbingshen, line 294: /loadbalance/broker broker change event.

wenbingshen (commented):
Processing cluster brokers metadata updates and processing deleted topics for consumer groups

wenbingshen:
@BewareMyPower My PR and I are ready. Zk client has been replaced with MetadataStore.  PTAL :)

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
```diff
@@ -0,0 +1,291 @@
...
+        try {
+            putLock.lock();
+            queue.put(event);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            log.error("Error put event {} to coordinator event queue {}", event, e);
+        } finally {
+            putLock.unlock();
+        }
+    }
```
BewareMyPower, line 100: We prefer the following lock idiom.

```java
lock.lock();
try {
    // ...
} finally {
    lock.unlock();
}
```

Because if `lock.lock()` failed with exceptions, `.unlock()` would still be called.
See examples in https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReentrantReadWriteLock.html.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
```diff
@@ -0,0 +1,291 @@
...
+
+                    if (element != null) {
+                        String kopBrokerStr = element.getAsString();
+                        Node kopNode = getNode(kopBrokerStr);
+                        kopBrokers.add(kopNode);
+                    }
+                }
+            } catch (Exception e) {
+                log.error("Get broker {} path data failed which have an error {}", broker, e.getMessage());
+                e.printStackTrace();
```
BewareMyPower, line 169: If you want to print the stack trace, use

```java
log.error("Get broker {} path data failed which have an error", broker, e);
```

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
BewareMyPower: The same applies to `clearAndPut`.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
```diff
@@ -0,0 +1,291 @@
...
+            this.deletionTopicsHandler.handleChildChange();
+        }
+    }
+
+    private void getBrokers(List<String> pulsarBrokers) {
+        HashSet<Node> kopBrokers = Sets.newHashSet();
+        pulsarBrokers.forEach(broker -> {
+            try {
+                Optional<GetResult> brokerData = metadataStore.get(
+                        getBrokersChangePath() + "/" + broker).join();
```
BewareMyPower, line 153: `join()` throws unchecked exception, which means you don't want to catch this exception. If you want to handle the exceptional case, use `get()` and don't catch `Exception`, catch the checked exceptions.

But I also concerns the synchronous call here. Because I found `getBrokers` could also be called in `BrokersChangeEvent#process`, which is a callback. Can it be changed to asynchronous call?

BewareMyPower (changes requested):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KopEventManagerTest.java
```diff
@@ -0,0 +1,214 @@
...
+
+    @AfterMethod
+    @Override
+    protected void cleanup() throws Exception {
+        adminClient.close();
+        super.internalCleanup();
+    }
+
+    @Test
+    public void testGroupState() throws Exception {
```
BewareMyPower, line 72: It took about 1 minute to run this test in my local env. Is it necessary? Please and the timeout to `@Test` and reduce the test time. It's better shorter than 10 seconds even 5 seconds. If the test is unstable for shorter time, there would be some serious problems with the implementation.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
```diff
@@ -0,0 +1,291 @@
...
+
+    class BrokersChangeEvent implements KopEvent {
+        @Override
+        public void process() {
+            try {
+                getBrokers(metadataStore.getChildren(getBrokersChangePath()).join());
+            } catch (Exception e) {
+                log.error("BrokersChangeEvent process have an error {}", e.getMessage());
+                e.printStackTrace();
+            }
```
BewareMyPower, line 267: Synchronous call in callback is easily to cause deadlock problem. Could you change to asynchronous calls? For example,

```java
            metadataStore.getChildren(getBrokersChangePath()).whenComplete((brokers, e) -> {
                if (e != null) {
                    log.warn("BrokersChangeEvent process have an error {}", e.getMessage());
                    return;
                }
                getBrokers(brokers); // maybe it should also be async
            });
```

Or is there any reason to use the synchronous call?

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -674,11 +677,11 @@ protected void handleTopicMetadataRequest(KafkaHeaderAndRequest metadataHar,
                     ctx.channel(), metadataHar.getHeader(), e);
                 allNodes.add(newSelfNode());
                 MetadataResponse finalResponse =
-                    new MetadataResponse(
-                        allNodes,
-                        clusterName,
-                        controllerId,
-                        Collections.emptyList());
+                        new MetadataResponse(
+                                Lists.newArrayList(allNodes),
```
BewareMyPower, line 681: Why did you change `allNodes` to `Lists.newArrayList(allNodes)`?

BewareMyPower: Sorry it's my fault here, solve it.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -491,7 +492,9 @@ protected void handleTopicMetadataRequest(KafkaHeaderAndRequest metadataHar,
 
         // Command response for all topics
         List<TopicMetadata> allTopicMetadata = Collections.synchronizedList(Lists.newArrayList());
-        List<Node> allNodes = Collections.synchronizedList(Lists.newArrayList());
+        Set<Node> allNodes = Collections.synchronizedSet(Sets.newHashSet());
```
BewareMyPower, line 495: Why did you change `allNodes`'s type to `Set`? Is there any benefit? What I can think of is you want to remove repeated nodes. Please explain the reason.

BewareMyPower:
I've left some comments, PTAL. The main problem is that we should avoid waiting a future in any callback, which was very dangerous and hard to investigate the reason if deadlock happened. There was a similar problem before, see https://github.com/streamnative/kop/pull/556.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
wenbingshen: good comment!

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
wenbingshen: Thanks, I will do

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
wenbingshen: good, you are right. I have no objection.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
wenbingshen: Here should keep the `List`, I have changed it back. Thanks

wenbingshen:
> I've left some comments, PTAL. The main problem is that we should avoid waiting a future in any callback, which was very dangerous and hard to investigate the reason if deadlock happened. There was a similar problem before, see #556.

Thank you for your meaningful comments. Based on your comment, I fixed the relevant code, and for the test, I added timeout=6000 ms. In my local area, the actual test can be completed in just a few hundred milliseconds. It took 1 minute before because I got the judgment condition wrong. I mistakenly wrote `describeGroup.get(groupId).state().name().equals("DEAD")` as `describeGroup.get(groupId).state( ).name().equals("Dead"))`. PTAL :)

wenbingshen:
A single test only took around 190 milliseconds.

image

wenbingshen:
The failed test does not seem to be related to the PR.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
```diff
@@ -0,0 +1,318 @@
...
+            if (e != null) {
+                log.error("Get pulsar brokers {} failed", pulsarBrokers, e);
+                return;
+            }
+
+            Collection<? extends Node> oldKopBrokers = adminManager.getBrokers();
+            adminManager.addBrokers(brokers);
+            log.info("Refresh kop brokers new cache {}, old brokers cache {}",
+                    adminManager.getBrokers(), oldKopBrokers);
+        });
```
BewareMyPower, line 201: `kopBrokersFuture` is not exposed, the only usage is the callback. I think we can move this part in `metadataStore.get(getBrokersChangePath() + "/" + broker)`'s callback

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AdminManager.java
```diff
@@ -221,4 +228,17 @@ public void shutdown() {
         });
         return result;
     }
+
+    public Collection<? extends Node> getBrokers() {
+        return brokersCache;
+    }
+
+    public void addBrokers(Set<Node> brokers) {
```
BewareMyPower, line 236: It's better to rename it to `setBrokers`.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupMetadata.java
```diff
@@ -598,6 +605,27 @@ public boolean hasPendingOffsetCommitsFromProducer(long producerId) {
         ));
     }
 
+    public Set<TopicPartition> collectPartitionsWithTopics(Set<String> topics) {
```
BewareMyPower, line 608: Maybe it's not a big issue. Since the collection only calls `addAll` method, it's better to use `List` rather than `Set`.

BewareMyPower:
Overall LGTM, just left a few comments, PTAL.

wenbingshen:
@BewareMyPower I have addressed your comments. PTAL.

BewareMyPower (approved):

Issue #713, opened 2021-09-07
[FEATURE] Support close Idle connections when timeout


Demogorgon314:
When there are too many idle connections on the server side, new client connections may not be created, so we need to clear idle connections over time.

PR #714, opened 2021-09-07
[FEATURE] Support close Idle connections when timeout

Demogorgon314:
#713

## Motivation
When there are too many idle connections on the server side, new client connections may not be created, so we need to clear idle connections over time.

## Modifications
Implemented using `IdleStateHandler`.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -53,6 +53,7 @@ This section lists configurations that may affect the performance.
 | ----------------- | ------------------------------------------------------------ | ------- |
 | maxQueuedRequests | Limit the queue size for request, like `queued.max.requests` in Kafka server. | 500     |
 | requestTimeoutMs  | Limit the timeout in milliseconds for request, like `request.timeout.ms` in Kafka client.<br>If a request was not processed in the timeout, KoP would return an error response to client. | 30000   |
+| connectionMaxIdleMs | Idle connections timeout: the server handler close the connections that idle more than this. like `connections.max.idle.ms` in kafka server. | 600000 |
 | failedAuthenticationDelayMs | Connection close delay on failed authentication: this is the time (in milliseconds) by which connection close will be delayed on authentication failure, like `connection.failed.authentication.delay.ms` in Kafka server. | 300 |
 
 > **NOTE**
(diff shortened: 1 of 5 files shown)


BewareMyPower (approved):

docs/configuration.md
```diff
@@ -53,6 +53,7 @@ This section lists configurations that may affect the performance.
 | ----------------- | ------------------------------------------------------------ | ------- |
 | maxQueuedRequests | Limit the queue size for request, like `queued.max.requests` in Kafka server. | 500     |
 | requestTimeoutMs  | Limit the timeout in milliseconds for request, like `request.timeout.ms` in Kafka client.<br>If a request was not processed in the timeout, KoP would return an error response to client. | 30000   |
+| connectionMaxIdleMs | Idle connections timeout: the server handler close the connections that idle more than this. like `connections.max.idle.ms` in kafka server. <br>**NOTE**: Set -1 to disable idle timeout. | 600000 |
```
Huanli-Meng, line 56:
```suggestion
| connectionMaxIdleMs | The idle connection timeout in milliseconds. If the value is greater than the idle connection timeout (such as `connections.max.idle.ms` used in Kafka server), the server handler will close the connections. <br>**Note**: If it is set to `-1`, it indicates that the idle connection timeout is disabled. | 600000 |
```

Huanli-Meng (commented):
Leave a minor comment. Thanks

docs/configuration.md
Demogorgon314: IMO, it may be clearer to change it like this?
Before:
``` markdown
If the value is greater than the idle connection timeout (such as `connections.max.idle.ms` used in Kafka server), the server handler will close the connections.
```
After:
```markdown
If the idle connection timeout (such as `connections.max.idle.ms` used in Kafka server), the server handler will close this idle connections.
```

BewareMyPower: @Huanli-Meng PTAL again

Huanli-Meng: As confirmed with Kai, the sentence is to be updated as below:
`If the idle connection timeout (such as `connections.max.idle.ms` used in Kafka server) is reached, the server handler will close this idle connection.`

Huanli-Meng (approved):
LGTM

PR #718, opened 2021-09-08
[WIP] Fix NullPointerException on pulsar 2.9.0-SNAPSHOT

Demogorgon314:
Currently pulsar has a bug, we need fix pulsar side problem then upgrade pulsar dependency.


Demogorgon314:
Pulsar 2.9.0 current don't expose `getZkClient()`, I need find a way to use `getLocalMetadataStore()` instead.

PR #723, opened 2021-09-10
Implement authorization for GET_TOPIC like Pulsar WebService - DRAFT

eolivelli:
In the KOP handler it is possible to list topics using less permission in respect to Pulsar Admin API.

This patch tries to add the same validation that we do in the Pulsar REST API.
Unfortunately there is no API in Pulsar internals to do this verification exactly the same way we are doing in the REST API, so this patch is quite ugly.

This is only a proof-of-concept that tries to demonstrate the problem.


eolivelli:
@Demogorgon314 @BewareMyPower PTAL

eolivelli:
A few tests fail because we are actually requesting more permissions to the user.

eolivelli:
I just realised that we already have some code to perform authorisation.
I will use the existing framework once we agree that we should fix this here in KOP

Demogorgon314:
This change LGTM, you may need add to SimpleAclAuthorizer. Thx.

PR #734, opened 2021-09-17
[FEATURE] Add KopRequestManager and KopResponseManager

### LINKED ISSUE #458: [FEATURE] Process some requests using a synchronous way
Currently, KoP uses a global request queue to keep the process order of Kafka requests.

https://github.com/streamnative/kop/blob/86249c2e00b292c87f94de20181c84b2dcfdd155/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java#L62

> The element of `requestQueue` contains a original request and a future of response.

Because all Kafka requests are processed asynchronously.

https://github.com/streamnative/kop/blob/86249c2e00b292c87f94de20181c84b2dcfdd155/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java#L173-L180

Once a request completed, it would trigger `writeAndFlushResponseToClient` to iterate over the `responseQueue` to send all completed responses in order.

However, if a request took much time to complete, it would block all followed requests. Sometimes it's unnecessary because some followed responses may not rely on the previous response. We've added a request timeout in KoP side, see https://github.com/streamnative/kop/pull/429. But it still uses the `responseQueue` for **all request** in a channel.

- Handle the requests that have timeout fields (like `PRODUCE` and `FETCH`) individually.
- For some requests, use a synchronous way to process them to avoid queueing.
- Eventually, we need to remove the `responseQueue`.

wenbingshen:
Fixes #458

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaChannelInitializer.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaChannelInitializer.java
@@ -57,6 +57,8 @@ public class KafkaChannelInitializer extends ChannelInitializer<SocketChannel> {
     @Getter
     private final StatsLogger statsLogger;
     private final MetadataCache<LocalBrokerData> localBrokerDataCache;
+    private final KopRequestManager requestManager;
+    private final KopResponseManager responseManager;
 
     public KafkaChannelInitializer(PulsarService pulsarService,
                                    KafkaServiceConfiguration kafkaConfig,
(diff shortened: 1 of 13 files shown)


BewareMyPower:
Could you explain in details for why does this PR fix #458 since this PR contains many changes?

wenbingshen:
> Could you explain in details for why does this PR fix #458 since this PR contains many changes?

My original intention was to fix #458. There are still some problems with this PR. I will adjust the status to WIP first. Thank you for your attention to the PR. I will add a detailed description when this pr ready. thanks. :)

PR #741, opened 2021-09-23
Refactor TxnMarkerQueue

Demogorgon314:
## Motivation
In previous code pass a ArrayList and change ArrayList in method may not a good idea. And we should follow the Kafka code with is:
```scala
def forEachTxnTopicPartition[B](f:(Int, BlockingQueue[TxnIdAndMarkerEntry]) => B): Unit =
  markersPerTxnTopicPartition.foreach { case (partition, queue) =>
    if (!queue.isEmpty) f(partition, queue)
  }
```
## Modifications
Add a method `forEachTxnTopicPartition`.


gaoran10 (approved):
LGTM

PR #748, opened 2021-09-26
Adjust the log level

wenbingshen:
1. In SaslAuthenticator, we should keep the log level as debug, because authentication is a very frequent occurrence in high-traffic systems, which is often encountered in kafka, the connection and disconnection between the client and the server has been happening all the time . If we need to troubleshoot problems, we can dynamically adjust the log level.
2. Before executing the debug, keep judging whether to enable the debug, even if log4j2 writes the log asynchronously, it can avoid generating unnecessary LogEvent.


wenbingshen:
@BewareMyPower Can you take a quick look at this pr. Here are only minor adjustments to the log level. :)

BewareMyPower (approved):

Demogorgon314 (approved):

PR #752, opened 2021-09-26
[BUGFIX] Fix wrong broker cache with multiple listeners

### LINKED ISSUE #751: Kafka client can connect to a non-existing listener in advertisedListeners
**Describe the bug**
#742 supports multiple listeners based on configs like

```properties
kafkaListeners=kafka_internal://localhost:9092,kafka_external://localhost:19092
kafkaProtocolMap=kafka_internal:PLAINTEXT,kafka_external:PLAINTEXT
advertisedListeners=pulsar:pulsar://localhost:6650,kafka_internal:pulsar://localhost:9092,kafka_external:pulsar://localhost:19092
```

However, it looks like even a listener name doesn't appear in `advertisedListeners`, the Kafka client could still connect to the associated end point.

**To Reproduce**

Add a test to `KafkaListenerNameTest`:

```java
    @Test(timeOut = 20000)
    public void testConnectListenerNotExist() throws Exception {
        final int externalPort = PortManager.nextFreePort();
        super.resetConfig();
        conf.setAdvertisedAddress(null);
        conf.setKafkaListeners("kafka://0.0.0.0:" + kafkaBrokerPort + ",kafka_external://0.0.0.0:" + externalPort);
        conf.setKafkaProtocolMap("kafka:PLAINTEXT,kafka_external:PLAINTEXT");
        conf.setAdvertisedListeners("pulsar:pulsar://localhost:" + brokerPort
                + ",kafka_external:pulsar://0.0.0.0:" + externalPort);
        super.internalSetup();
        final Properties props = newKafkaProducerProperties();
        props.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:" + kafkaBrokerPort);
        final KafkaProducer<String, String> producer = new KafkaProducer<>(props);
        try {
            producer.send(new ProducerRecord<>("my-topic", "hello")).get();
            Assert.fail("send should fail");
        } catch (InterruptedException | ExecutionException e) {
            log.info("Send failed: {}", e.getMessage());
        }
        producer.close();
        super.internalCleanup();
    }
```

**Expected behavior**
When a producer sends messages to a listener that doesn't appear in `advertisedListeners`, it should fail.

**Additional context**
After I debugged this issue, I found it's because when the first lookup request with listener name `kafka` failed, the response would be wrong. What's worse is that it returns the wrong broker cache in `AdminManager` to Kafka client, then Kafka client connected to the listener of `kafka_external`.

wenbingshen:
Fixes #751 

### Motivation
The brokerCache in the current AdminManger does not distinguish the listener name, which causes the wrong broker metadata to be returned when there are multiple listenerNames

### Modifications
Support multiple listenerNames in BrokerCache

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/EndPoint.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/EndPoint.java
@@ -30,8 +30,7 @@ import org.apache.kafka.common.security.auth.SecurityProtocol;
  */
 public class EndPoint {
 
-    @Getter
-    private static final String END_POINT_SEPARATOR = ",";
+    public static final String END_POINT_SEPARATOR = ",";
     private static final String PROTOCOL_MAP_SEPARATOR = ",";
     private static final String PROTOCOL_SEPARATOR = ":";
     private static final String REGEX = "^(.*)://\\[?([0-9a-zA-Z\\-%._:]*)\\]?:(-?[0-9]+)";
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/EndPoint.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/EndPoint.java
@@ -30,6 +30,7 @@ import org.apache.kafka.common.security.auth.SecurityProtocol;
  */
 public class EndPoint {
 
+    @Getter
     private static final String END_POINT_SEPARATOR = ",";
     private static final String PROTOCOL_MAP_SEPARATOR = ",";
     private static final String PROTOCOL_SEPARATOR = ":";
@@ -52,9 +53,7 @@ public class EndPoint {
     public EndPoint(final String listener, final Map<String, SecurityProtocol> protocolMap) {
         this.originalListener = listener;
         final String errorMessage = "listener '" + listener + "' is invalid";
-        final Matcher matcher = PATTERN.matcher(listener);
-        checkState(matcher.find(), errorMessage);
-        checkState(matcher.groupCount() == 3, errorMessage);
+        final Matcher matcher = matcherListener(listener, errorMessage);
 
         this.listenerName = matcher.group(1);
         if (protocolMap == null || protocolMap.isEmpty()) {
@@ -81,8 +80,6 @@ public class EndPoint {
         this.port = Integer.parseInt(matcher.group(3));
         checkState(port >= 0 && port <= 65535, errorMessage + ": port " + port + " is invalid");
     }
-
-
 
     public InetSocketAddress getInetAddress() {
         return new InetSocketAddress(hostname, port);
@@ -148,4 +145,12 @@ public class EndPoint {
         }
         return protocolMap;
     }
+
+    public static Matcher matcherListener(String listener, String errorMessage) {
+        final Matcher matcher = PATTERN.matcher(listener);
+        checkState(matcher.find(), errorMessage);
+        checkState(matcher.groupCount() == 3, errorMessage);
+        return matcher;
+    }
+
 }
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -282,7 +282,7 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
                                Boolean tlsEnabled,
                                EndPoint advertisedEndPoint,
                                StatsLogger statsLogger) throws Exception {
-        super(statsLogger, kafkaConfig);
+        super(statsLogger, kafkaConfig, advertisedEndPoint);
         this.pulsarService = pulsarService;
         this.tenantContextManager = tenantContextManager;
         this.clusterName = kafkaConfig.getClusterName();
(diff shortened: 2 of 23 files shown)


wenbingshen:
@BewareMyPower @Demogorgon314 PTAL, Thanks.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
```diff
@@ -242,18 +251,26 @@ private JsonObject parseJsonObject(String info) {
     }
 
     @VisibleForTesting
```
BewareMyPower, line 253: Maybe it's a miss in previous review. Why did you add `@VisibleForTesting` here? I see `getNodes` is only called in private method `KopEventManager#getBrokers`.

BewareMyPower: Or should we add tests for this method?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
```diff
@@ -242,18 +251,26 @@ private JsonObject parseJsonObject(String info) {
...
+        HashMap<String, Set<Node>> nodesMap = Maps.newHashMap();
+        String[] kopBrokerArr = kopBrokerStrs.split(END_POINT_SEPARATOR);
+        for (String kopBrokerStr : kopBrokerArr) {
+            final String errorMessage = "kopBrokerStr " + kopBrokerStr + " is invalid";
+            final Matcher matcher = PATTERN.matcher(kopBrokerStr);
+            checkState(matcher.find(), errorMessage);
+            checkState(matcher.groupCount() == 3, errorMessage);
+            String listenerName = matcher.group(1);
+            String host = matcher.group(2);
+            String port = matcher.group(3);
```
BewareMyPower, line 264: This parse is nearly the same with what `EndPoint` does. Could you reuse the existing code? For example, add a conversion from `EndPoint` to `Node` or add some utils methods?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
BewareMyPower: Sorry it's a problem with my Idea (maybe cache was not cleared), resolve it.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -524,7 +524,7 @@ protected void handleTopicMetadataRequest(KafkaHeaderAndRequest metadataHar,
         List<TopicMetadata> allTopicMetadata = Collections.synchronizedList(Lists.newArrayList());
         List<Node> allNodes = Collections.synchronizedList(Lists.newArrayList());
         // Get all kop brokers in local cache
-        allNodes.addAll(adminManager.getBrokers());
+        allNodes.addAll(adminManager.getBrokers(metadataHar.getListenerName()));
```
BewareMyPower, line 527: Could we just use

```java
        allNodes.addAll(adminManager.getBrokers(advertisedEndPoint.getListenerName()));
```

here? Then we don't need to add a `listenerName` field to `KafkaHeaderAndRequest`.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/SaslPlainTestBase.java
```diff
@@ -113,8 +113,7 @@ protected void setup() throws Exception {
                     .build());
         admin.namespaces().createNamespace(TENANT + "/" + NAMESPACE);
         admin.topics().createPartitionedTopic(TOPIC, 1);
-        admin
-            .namespaces().grantPermissionOnNamespace(TENANT + "/" + NAMESPACE, SIMPLE_USER,
+        admin.namespaces().grantPermissionOnNamespace(TENANT + "/" + NAMESPACE, SIMPLE_USER,
```
BewareMyPower, line 116: We should avoid unrelated changes.

BewareMyPower:
My review is done, PTAL.

wenbingshen:
> My review is done, PTAL.

Thanks review. Addressed done. PTAL

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/EndPoint.java
```diff
@@ -30,6 +30,7 @@
  */
 public class EndPoint {
 
+    @Getter
```
BewareMyPower, line 33: For a `public static final` field, the getter is useless and looks weird. It's better to just modify it to public.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandlerTest.java
```diff
@@ -715,7 +715,8 @@ public void testMetadataForNonPartitionedTopic(short version) throws Exception {
         final MetadataRequest request = new MetadataRequest(Collections.singletonList(topic), false, version);
         final CompletableFuture<AbstractResponse> responseFuture = new CompletableFuture<>();
         handler.handleTopicMetadataRequest(
-                new KafkaHeaderAndRequest(header, request, PulsarByteBufAllocator.DEFAULT.heapBuffer(), null),
+                new KafkaHeaderAndRequest(header,
+                        request, PulsarByteBufAllocator.DEFAULT.heapBuffer(), null),
                 responseFuture);
```
BewareMyPower, line 720: Could you revert all unrelated changes?

BewareMyPower: In this file. Only some indent changes were made. It's confused and breaks the `git blame` history.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/EndPoint.java
wenbingshen: good idea

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandlerTest.java
wenbingshen: done. all unrelated changes have been reverted. PTAL again.

BewareMyPower (approved):

PR #753, opened 2021-09-27
Fix producer and reader not close when GroupMetadataManager shutdown

Demogorgon314:
## Motivation
When `GroupMetadataManager.shutdown()` the `scheduler` shutdown first, so there is no chance to execute `producer` and `reader` close.

## Modifications
Close `producer` and `reader` first.


BewareMyPower:
Could you give more details about why here is no chance to execute producer and reader close after the scheduler shutdown?

Demogorgon314:
@BewareMyPower 
See `producerCompletableFuture.thenComposeAsync(..., scheduler)` and `whenCompleteAsync(..., scheduler) ` gave a closed scheduler as executer.  

This future will never complete because throw a `RejectedExecutionException`, also because of all future is called by async, so this no exceptions throw in `shutdown()`.

BewareMyPower (approved):

PR #754, opened 2021-09-27
[WIP] fix:When pendingTopicFuture is completed, it is not removed from pend…

lordcheng10:
When pendingTopicFuture is completed, it is not removed from pendingTopicFuturesMap

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -1042,11 +1042,14 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
                 persistentTopicConsumer.accept(topicFuture.getNow(Optional.empty()));
             } else {
                 // topic is not available now
-                Consumer<TopicPartition> cleanCompletedPendingTopicFuture = (ignore) -> pendingTopicFuturesMap.remove(topicPartition);
+                Consumer<TopicPartition> cleanCompletedPendingTopicFuture = (ignore) -> {
+                    pendingTopicFuturesMap.remove(topicPartition);
+                };
                 pendingTopicFuturesMap
                         .computeIfAbsent(topicPartition, ignored ->
                                 new PendingTopicFutures(requestStats))
-                        .addListener(topicFuture, persistentTopicConsumer, exceptionConsumer, cleanCompletedPendingTopicFuture);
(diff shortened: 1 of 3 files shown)


BewareMyPower:
Why should we remove it from `pendingTopicFuturesMap`? What's the problem of the current implementation?

lordcheng10:
> pendingTopicFuturesMap

If we don't remove it, after a long time running, will there be too many objects in this map, and every connection seems to have a map.

> Why should we remove it from `pendingTopicFuturesMap`? What's the problem of the current implementation?

`            if (topicFuture.isDone()) {
                persistentTopicConsumer.accept(topicFuture.getNow(Optional.empty()));
            } else {
                // topic is not available now
                pendingTopicFuturesMap
                        .computeIfAbsent(topicPartition, ignored ->
                                new PendingTopicFutures(requestStats))
                        .addListener(topicFuture, persistentTopicConsumer, exceptionConsumer);
            }`
I consider two points:
1. Without cleaning up for a long time, the objects in the map have been backlogged, and each connection seems to have a map;
2. For the same topic partition, if pendingTopicFuture appears again, pendingTopicFuturesMap.computeIfAbsent will fail;

For the second point, when the pendingTopicFuture situation occurs again, topicFuture.isDone() usually returns true due to the cache, but the cache seems to be invalid when the topic is unloaded:
image

BewareMyPower:
> the objects in the map have been backlogged, and each connection seems to have a map;

The queue size is the total partition number of all PRODUCE requests. The memory usage should not increase. And when a connection is closed, the map could be garbage collected. And after your change, each time a topic future is completed, the `PendingTopicFutures` of the partition will be removed. Then next time a PRODUCE request arrived, `PendingTopicFutures` will be created again.

> but the cache seems to be invalid when the topic is unloaded:

Then it's better to clear the cache when topic is unloaded. However, before doing this, could you add a test to verify that this design would make KoP not handle the unload event well? Because current unload tests work well.

BewareMyPower:
@Demogorgon314 Please also take a look at this PR, thanks.

Demogorgon314:
I'm agree with @BewareMyPower , If we remove it from `pendingTopicFuturesMap`, the `pendingTopicFuturesMap` is no need to exist, because every time we use `pendingTopicFuturesMap` we just add and remove it. 

We should add a test to verify is any problem when namespace bundle unload or load.

> Without cleaning up for a long time, the objects in the map have been backlogged

Maybe we need remove `PendingTopicFutures` cache in `NamespaceBundleOwnershipListener`.

PR #755, opened 2021-09-27
Disable stats recorder for PulsarClient in group coordinator

BewareMyPower:
### Motivation
The `PulsarClient` object returned by `PulsarService#getClient` doesn't change the default `statsIntervalSeconds` config, Therebefore, each producer or reader created in `GroupMetadataManager` will has an internal `ProducerStatsRecorderImpl` or `ConsumerStatsRecorderImpl` object, which triggers a `TimerTask` whose period is 60 seconds. It's a waste of CPU usage especially when the offset topic's partitions number is large.

### Modifications
Since we have already defined our own `PulsarClient` wrapper (`LookupClient`) for topic lookup, this PR adds a common abstract class `AbstractPulsarClient` to customize the client config. Then add a `SystemTopicClient` class that configures `statsIntervalSeconds` with zero value to disable stats recorder.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AbstractPulsarClient.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AbstractPulsarClient.java
@@ -1,0 +1,104 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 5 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AbstractPulsarClient.java
```diff
@@ -0,0 +1,104 @@
...
+    public AbstractPulsarClient(@NonNull final PulsarClientImpl pulsarClient) {
+        this.pulsarClient = pulsarClient;
+    }
+
+    @Override
+    public void close() {
+        try {
+            pulsarClient.close();
+        } catch (PulsarClientException e) {
+            log.warn("Failed to close PulsarClient of LookupClient", e);
```
wenbingshen, line 49: Now it also may be `SystemTopicClient`

wenbingshen (commented):
This pr is useful for me. Just leave a minor comment.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AbstractPulsarClient.java
BewareMyPower: Now I printed the actual class name here.

wenbingshen (approved):
lgtm

Demogorgon314 (approved):

PR #757, opened 2021-09-27
[BUGFIX] Keep KopEventManager as a single-threaded event model

### LINKED ISSUE #756: [BUG] Keep KopEventManager as a single-threaded event model
**Describe the bug**
#735 implements tenant aware metadata topics. It instantiates a KopEventManager for each GroupCoordinator. In fact, this is wrong. KopEventManger should be a global event handler. Unless we need to use KopEventManager to implement some other models, we A single instance should be kept to handle events.

wenbingshen:
Fixes #756 

### Motivation
#735  implements tenant aware metadata topics. It instantiates a KopEventManager for each GroupCoordinator. In fact, this is wrong. KopEventManger should be a global event handler. Unless we need to use KopEventManager to implement some other models, we A single instance should be kept to handle events.

### Modifications
1. Keep KopEventManager as a single-threaded event model.
2. Leave a discussion on KopEventManager#307

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
@@ -32,7 +32,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
@@ -323,7 +322,8 @@ public class KopEventManager {
                             topicsFullNameDeletionsSets.add(kopTopic.getFullName());
                         });
 
-                        Iterable<GroupMetadata> groupMetadataIterable = groupCoordinator.getGroupManager().currentGroups();
+                        Iterable<GroupMetadata> groupMetadataIterable = groupCoordinator
+                                .getGroupManager().currentGroups();
                         HashSet<TopicPartition> topicPartitionsToBeDeletions = Sets.newHashSet();
 
                         groupMetadataIterable.forEach(groupMetadata -> {
@@ -345,7 +345,8 @@ public class KopEventManager {
                         }
 
                         log.info("Tenant {} GroupMetadata delete topics {}, no matching topics {}",
-                                tenant, curDeletedTopics, Sets.difference(topicsFullNameDeletionsSets, curDeletedTopics));
+                                tenant, curDeletedTopics,
+                                Sets.difference(topicsFullNameDeletionsSets, curDeletedTopics));
 
                         deletedTopics.addAll(curDeletedTopics);
                     }
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
@@ -303,19 +303,6 @@ public class KopEventManager {
                 final AtomicInteger pendingCoordinators = new AtomicInteger(currentCoordinators.size());
 
                 currentCoordinators.forEach((tenant, groupCoordinator) -> {
-
-                    // After we implement the creation and deletion of topics and other logic
-                    // updates to multi-tenancy. We need to pay attention to the TODO below
-                    // TODO
-                    // now we have tenant/coordinator/topicsDeletions
-                    // It is possible that different coordinators have the same Kafka topics?
-                    // Because there is no multi-tenancy on the Kafka side
-                    // The question here is, which topic does the kafka admin client delete?
-                    // Is it related to tenants or delete all kafka topics with the same name under all tenants?
-                    // The implementation here only follows the concept that there is no multi-tenancy in Kafka,
-                    // that mean deleted the Kafka topics with the same name under all tenants,
-                    // and leave the rest for everyone to discuss
-
                     if (groupCoordinator.isActive()) {
                         HashSet<String> topicsFullNameDeletionsSets = Sets.newHashSet();
                         HashSet<KopTopic> kopTopicsSet = Sets.newHashSet();
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
@@ -304,6 +304,8 @@ public class KopEventManager {
 
                 currentCoordinators.forEach((tenant, groupCoordinator) -> {
 
+                    // After we implement the creation and deletion of topics and other logic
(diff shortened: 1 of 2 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
```diff
@@ -288,54 +289,74 @@ void process(BiConsumer<String, Long> registerEventLatency,
...
+
+                    // TODO
+                    // now we have tenant/coordinator/topicsDeletions
+                    // It is possible that different coordinators have the same Kafka topics?
+                    // Because there is no multi-tenancy on the Kafka side
+                    // The question here is, which topic does the kafka admin client delete?
+                    // Is it related to tenants or delete all kafka topics with the same name under all tenants?
+                    // The implementation here only follows the concept that there is no multi-tenancy in Kafka,
+                    // that mean deleted the Kafka topics with the same name under all tenants,
+                    // and leave the rest for everyone to discuss
```
wenbingshen, line 314: leave a discuss

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
```diff
@@ -61,7 +62,7 @@
             new LinkedBlockingQueue<>();
     private final KopEventThread thread =
             new KopEventThread(kopEventThreadName);
-    private final GroupCoordinator coordinator;
+    private final Map<String, GroupCoordinator> groupCoordinatorsByTenant;
```
wenbingshen, line 65: There are no need to create a `KopEventManager` for each `GroupCoordinator`, but pass the `GroupCoordinator` list to `KopEventManager`

wenbingshen:
ping @eolivelli @BewareMyPower @Demogorgon314 please see see it. thx

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
```diff
@@ -288,54 +289,74 @@ void process(BiConsumer<String, Long> registerEventLatency,
...
-                            kopTopic -> collectDeleteTopics.contains(kopTopic.getFullName())
-                    ).map(KopTopic::getOriginalName).collect(Collectors.toSet());
-
-                    deletedTopics.forEach(deletedTopic -> {
-                        metadataStore.delete(
-                                getDeleteTopicsPath() + "/" + deletedTopic, Optional.of((long) -1));
-                    });
-                }
+                // Localize groupCoordinatorsByTenant to avoid multi-thread conflicts
+                final Map<String, GroupCoordinator> currentCoordinators = groupCoordinatorsByTenant;
```
eolivelli, line 300: This is not a copy operation

eolivelli (commented):
I will take a look more closer tomorrow

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
wenbingshen: You are right. I will use a copy operation to create a new Map. Thanks very much.

wenbingshen:
Kop only exposes external topics for the Kafka client, which will be converted to the fullName pulsar topic inside kop. Maybe we need to update `KafkaRequestHandler#handleDeleteTopics`: In this method, the deleted topic will be bounded to the tenant and updated to the fullName pulsar topic topic. Another doubt is whether different tenants or different namespaces have the same Kafka topic, for example:
Kafka topic: test-topic
tenant1/namespace1/test-topic
tenant1/namespace2/test-topic
tenant2/namespace1/test-topic
tenant2/namespace3/test-topic
Is this possible?

wenbingshen:
I found that KopTopic always uses the default tenant and namespace to create and delete topics by `public/default`,
see https://github.com/streamnative/kop/blob/master/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/KopTopic.java#L76

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
```diff
@@ -288,54 +290,76 @@ void process(BiConsumer<String, Long> registerEventLatency,
...
-
-                    deletedTopics.forEach(deletedTopic -> {
-                        metadataStore.delete(
-                                getDeleteTopicsPath() + "/" + deletedTopic, Optional.of((long) -1));
-                    });
-                }
+                // Localize groupCoordinatorsByTenant to avoid multi-thread conflicts
+                final Map<String, GroupCoordinator> currentCoordinators = new HashMap<>(groupCoordinatorsByTenant);
+                final Set<String> deletedTopics = Sets.newConcurrentHashSet();
+                final AtomicInteger pendingCoordinators = new AtomicInteger(currentCoordinators.size());
```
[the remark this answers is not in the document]
wenbingshen: Thanks review. You are right, I want to modify it to `currentCoordinators.entrySet().stream().forEach(()->{})`. Use concurrent execution to increase the processing speed, do you think it’s okay?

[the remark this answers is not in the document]
wenbingshen: sorry. it should be use `currentCoordinators.entrySet().parallelStream().forEach`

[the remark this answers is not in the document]
Demogorgon314: In small amount of data, use concurrent may increase context switching overhead, may reduce the speed. We need to evaluate whether it is worth it. @BewareMyPower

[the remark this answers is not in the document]
wenbingshen: Here, If I use the int type, idea prompts to use the atomic type.

image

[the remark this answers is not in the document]
Demogorgon314: Oh, my mistake forget it. It still need AtomicInteger.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
BewareMyPower: > which topic does the kafka admin client delete?

There're two configs `kafkaTenant` and `kafkaNamespace` that represent the default namespace in `KopTopic`.

Does it answer your questions? If yes, I think these comments can be removed.

wenbingshen: Yes. Let us remove these comments. :)

eolivelli (approved):
LGTM

wenbingshen:
> LGTM

Thanks very much for your review. :)

wenbingshen:
now @BewareMyPower @Demogorgon314 PTAL. Thanks.

BewareMyPower (approved):

PR #758, opened 2021-09-28
[FEATURE] Add authorization to create topic and delete topic requests

Demogorgon314:
## Motivation
Currently KoP don't support authorization when users use adminClient to create topic or delete topic. So we need support this feature. 

In Pulsar, create or delete topic should have tenantAdmin or SuperAdmin permission, this implementation will follow the pulsar way.

## Modifications
* Add authorization to `handleCreateTopics`
* Add authorization to `handleDeleteTopics`
* Add new test units for create or delete topic permissions
* Add new test units for create topic with authorization
* Add new test units for delete topic with authorization

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AdminManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AdminManager.java
@@ -32,6 +32,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.kafka.common.Node;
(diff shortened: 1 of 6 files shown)


Demogorgon314:
BTW, as @wenbingshen mentioned, the `KopTopic` always uses the default tenant and namespace, it should be fix.

BewareMyPower:
> the KopTopic always uses the default tenant and namespace, it should be fix.

Could you explain how to fix it? I think it's a feature to support short topic name, just like Pulsar maps `topic` to `persistent://public/default/topic`, what's different is KoP support configuring the default namespace by `kafkaTenant` and `kafkaNamespace`.


BTW, I found there are other requests that don't perform the authorization, like `LIST_GROUPS`, `DELETE_GROUPS`, most of them are admin operations. Could you make a task list and open a master issue?

Demogorgon314:
@BewareMyPower You're right, this problem should be: Should we support short topic name auto mapping to current login username(tenant/namespace). this feature depend on should we enforce users use "tenant/namespace" as username , instead of "tenant". 

After I thought about it, support auto mapping current login username is not good idea.

BewareMyPower:
>  Should we support short topic name auto mapping to current login username(tenant/namespace). 

No. The topic name and login resources should be separated.

BewareMyPower (approved):

Issue #539, opened 2021-05-31
[FEATURE]Can kop support low version kafka-client, such as 0.10.2


wking1986:
**Is your feature request related to a problem? Please describe.**
Currently kop supports high versions of kafka, such as 1.x and 2.x, but many low versions of kafka in production environments (such as 0.10.2), and this part of users is more willing to switch pulsar

BewareMyPower:
Currently we focused on some performance and stable issues for incoming KoP 2.8.0 (maybe a few days later), so it may be not possible to support 0.10.x.y before 2.8.0. But we'll try to support Kafka 0.10.x.y and 0.11.x.y for next releases.

wking1986:
Thanks for your reply, after supporting 0.10.x, we are going to launch pulsar on a large scale

zhaohaidao:
Hi, @BewareMyPower I would like to pick this up as we are use kafka-0.10.2 with a large scale, and we may start to test pulsar soon

BewareMyPower:
@zhaohaidao Thanks for your help. FYI, after you fixed the incompatibility with Kafka 0.10.x.y client, you can add a `kafka-0-10` module just like `kafka-1-0` module to test Kafka 0.10.x.y once #605 was merged.

BewareMyPower:
I'll assign this issue to you first @zhaohaidao

BewareMyPower:
@zhaohaidao Hi, just to confirm that are you still working on this issue now?

PR #791, opened 2021-10-07
[FEATURE] Support kafka LogValidator validate inner records and compression codec when handle producer request with entryFormat=kafka

### LINKED ISSUE #687: [FEATURE] Need Increase the guarantee of the correctness of the internal message offset
This issue was discovered when I fix #681 
In Kafka, the message offset is allocated locally by the partition leader, and the message written in kop will not display the allocated message offset. When the producer turns on compression or writes v0, v1 messages, the internal message set offset may be abnormal. When the consumer When reading the message, even if the down-conversion is done, only the batch baseoffset is modified, but the internal record message offset may still be wrong.

The batch offset in brokerEntryMetadata is correct. When down conversion is needed or the message version is v0 or v1, we should reallocate the internal message offset.

Because at present we are not doing internal message offset verification like Kafka when we write. If there is no better way, at least for now, we should make such a guarantee when we consume.

wenbingshen:
fixes #687 

### Motivation
Previously, in order to support low version kafka clients, such as 0.10 version clients, we verified the data when processing fetch requests. Because the lower message format would maintain the internal message set, according to the production situation I encountered, in a certain sometimes there is an error in the offset of the internal message set, such as the bug in the low version of the sarama client.

The problem is that for the low version message format, no matter whether there is a problem with the internal message set, we will regenerate the message records, which is unnecessary in some cases.

### Modifications
Support message set verification in production like kafka, and only need to verify when entryFormat=kafka.
Because when entryFormat=pulsar, we still cannot avoid doing message conversion during consumption.


wenbingshen:
If we can use this pr. Then we can use direct memory to rebuild kafka records like #673.
I can work for this after this pr has been merged.

BewareMyPower:
Please rebase to master since it has conflicts with your previous PR.

wenbingshen:
> Please rebase to master since it has conflicts with your previous PR.

Thanks for the reminder, I happen to be doing this job and the conflict has been resolved. :)

wenbingshen:
The result of my local performance test.

image

wenbingshen:
The failed test does not seem to be related to this pr.

wenbingshen:
> Here's my test result for encoding before and after this PR. **Before (Kafka)** represents `KafkaEntryFormatter` before this PR.
> 
> Before (Kafka)	Before (Pulsar)	After (Kafka)	After (Pulsar)
> Fixed records for 100 times	12	1246	48	1267
> Random records for 100 times	10	854	23	961
> Fixed records for 1000 times	78	7630	257	7230
> Random records for 1000 times	88	7376	170	6682
> We can see the encoding time of `KafkaEntryFormatter` has increased 2 to 3 times after this PR because the time of `PulsarEntryFormatter` didn't change significantly.
> 
> I think we cannot accept this performance lost just for compatibility with some **special cases** for Kafka 0.10 client.
> 
> If you insisted on this change, please add a new `EntryFormatter` implementation.

Okay, if we add a new `entryFormatter`, does it mean that the previous `KafkaEntryFormatter` no longer needs to support 0.10 clients and only supports magic=2?

wenbingshen:
> 1. If it doesn't support 0.10 clients, why could **kafka-0-10** module pass the tests of `io.streamnative.pulsar.handlers.kop.compatibility.BasicEndToEndKafkaTest`?

Starting from [KIP-32](https://cwiki.apache.org/confluence/display/KAFKA/KIP-31+-+Move+to+relative+offsets+in+compressed+message+sets), that is version kafka-0.10.0.0, timestamp and relative offset are introduced. Please check this issue [KAFKA-3025](https://issues.apache.org/jira/browse/KAFKA-3025) and this [code link](https://github.com/apache/kafka/blob/937054bb5e424d07075a758aad46abea129d0b24/clients/src/main/java/org/apache/kafka/common/record/AbstractLegacyRecordBatch.java#L370-L376) . The client will parse the compressed message set and calculate the absolute displacement through baseOffset and relative displacement. If all relative displacements are encountered as 0, the absolute displacement will always be the same, and errors will occur for submitting commitOffset and fetch request from consumers.

Why the compatibility test can pass, please check the comment section of [this code](https://github.com/apache/kafka/blob/6b005b2b4eece81a5500fb0080ef5354b4240681/core/src/main/scala/kafka/log/LogValidator.scala#L431-L437)

The old sarama client and filebeat are just a few of them. Due to my current production experience, I am not sure if any other clients or clients based on the sarama package have this problem.

> 2. If the problem only happens with Golang Sarama implementation, please make it clear that KafkaEntryFormatter supports official Java client 0.10 well but not Golang Sarama client for 0.10 Kafka protocol.

In KOP's 2.9.0 plan list, we have decided to support the [kafka-0.10 client](https://github.com/streamnative/kop/issues/539). Why, knowing this problem, we only marked that `KafkaEntryFormatter` is only applicable to java clients.

> 3. As for the reason I made this PR
I have described the problems caused by the current implementation in the `Motivation` of this PR:
The problem is that for the low version message format, no matter whether there is a problem with the internal message set, we will regenerate the message records, which is unnecessary in some cases. And `LogValidator` will only reconstruct the message when it is really needed.

BewareMyPower:
> And sarama and filebeat are still very widely used.

Yeah. So we need to let these users know, if you want the support for sarama and filebeat (I'm not sure if higher version has the same problem), please choose another `entryFormat` that has a significant performance overhead than `entryFormat=kafka` and make sure you can accept it.

But for those don't use lower version sarama, you can choose `entryFormat=kafka` for better performance.

BewareMyPower:
The original purpose to introduce `EntryFormatter` is to provide a most efficient way for Kafka users. If the performance was affected badly by the extra operations for compatibility or other things, we should give up the support and provide another `entryFormat` for them.

wenbingshen:
> The original purpose to introduce `EntryFormatter` is to provide a most efficient way for Kafka users. If the performance was affected badly by the extra operations for compatibility or other things, we should give up the support and provide another `entryFormat` for them.

What you said is very reasonable, I very much agree. Thanks for your discussion. Later I will implement an `entryFormatter` that supports kafka lower versions. For the current `KafkaEntryFormatter` it will be the default kafka entryFormmtter and only support `magic=2`.

wenbingshen:
> > And sarama and filebeat are still very widely used.
> 
> Yeah. So we need to let these users know, if you want the support for sarama and filebeat (I'm not sure if higher version has the same problem), please choose another `entryFormat` that has a significant performance overhead than `entryFormat=kafka` and make sure you can accept it.
> 
> But for those don't use lower version sarama, you can choose `entryFormat=kafka` for better performance.

Yeah. The higher versions of sarama and filebeat have fixed this problem, but I still can’t rule out whether other clients will cause this problem, but if the user encounters this problem, we can recommend them to use another entryFormatter we provide from this pr.

wenbingshen:
@Demogorgon314 I have addressed all of your comments above in the latest commit, PTAL. :)

wenbingshen:
> Leave my comments. PTAL. Thanks.

Thanks for your review. I have addressed your comment.

wenbingshen:
This is a known flaky test #809 
https://github.com/streamnative/kop/pull/791/checks?check_run_id=3902214148

BewareMyPower:
> @BewareMyPower And can you help add docs? Thanks very much.

There's a `doc-required` label for this PR. I'll add related docs after this PR is merged.

BewareMyPower:
Left some comments. In addition, I cannot see any different tests between `KafkaV1EntryFormatter` and `KafkaMixedEntryFormatter`. I think the newly added tests for `mixed_kafka` are not necessary, except tests under `io.streamnative.pulsar.handlers.kop.compatibility`.

We already have Kafka 0.10 client for tests. Does it mean only some older version Golang sarama client can make a difference between `kafka` and `mixed_kafka`?

Or could you do some simple tests just for `EntryFormatter#encode` method? Like creating a V0 message set, and show the different behaviors between `kafka` and `mixed_kafka`?

BewareMyPower:
If the newer sarama client has fixed the problem. Could you give the version?

wenbingshen:
> Left some comments. In addition, I cannot see any different tests between `KafkaV1EntryFormatter` and `KafkaMixedEntryFormatter`. I think the newly added tests for `mixed_kafka` are not necessary, except tests under `io.streamnative.pulsar.handlers.kop.compatibility`.
> 
> We already have Kafka 0.10 client for tests. Does it mean only some older version Golang sarama client can make a difference between `kafka` and `mixed_kafka`?
> 
> Or could you do some simple tests just for `EntryFormatter#encode` method? Like creating a V0 message set, and show the different behaviors between `kafka` and `mixed_kafka`?

All comments done. PTAL `io.streamnative.pulsar.handlers.kop.format.EntryFormatterTest` for the test to show the different behaviors between `kafka` and `mixed_kafka`.

PR #792, opened 2021-10-07
Add test case about multi-tenancy: cannot access other tenants data

eolivelli:
Add a new test case that ensures that a user cannot access other tenants data.
Testing:
- produce
- consume
- listTopics


eolivelli:
@Demogorgon314 PTAL as well
it is only a new test case

Demogorgon314:
Could you please rebase into master branch. and fix the code style. Thanks.

PR #813, opened 2021-10-15
TransactionCoordinator: prevent thread leaks during broker shutdown

eolivelli:


eolivelli:
@Demogorgon314 probably this patch is good to go now

Issue #818, opened 2021-10-19
[FEATURE] Smart listener support (KIP-103 compatibility)


EronWright:
### Motivation
As of [KIP-103](https://cwiki.apache.org/confluence/display/KAFKA/KIP-103%3A+Separation+of+Internal+and+External+traffic), Kafka automatically selects the appropriate listener based on the bind address that the topic metadata request arrives on.   Also, listener support is generalized; one may define numerous named listeners for the same security protocol.

This enhancement relates to [PIP-95 (Smart Listeners)](https://github.com/apache/pulsar/issues/12040).

### Use Case
Imagine that the broker is running in a Kubernetes environment, supporting a combination of external and internal clients.  The external clients connect via an API gateway (e.g. Istio), while the internal clients connect directly to the brokers.  Initially the client sends a topic metadata request to a configured service endpoint, obtains a broker address, and then connects to that broker.  As per KIP-103, the topic metadata response should vary based on which service endpoint was used, to provide the appropriate broker address.  The client need not supply a listener name in the metadata request; the correct behavior is automatic.

### Suggested Changes
1. Open a server socket for each configured listener.  Add configuration elements as necessary.
2. When a metadata request arrives on a given server socket, lookup the broker as normal, and automatically select the corresponding advertised listener for that broker.
3. Allow for numerous listeners to be defined for the same security protocol.  Do not use the security protocol as the key when looking up an advertised listener.

PR #742, opened 2021-09-24
support multiple listeners with multiple endpoints

wangjialing218:
Fixes #669
Fixes #574

To support multiple listeners with multiple protocols, we need to set follow configration, and deprecated `kafkaAdvertisedListeners` and `kafkaListenerName`.

> kafkaListeners=internal://0.0.0.0:9092,internal_ssl://0.0.0.0:9093,external://0.0.0.0:19002,external_ssl:0.0.0.0:19003
> kafkaProtocalMap=internal:PLAINTEXT,internal_ssl:SSL,external:PLAINTEXT,external_ssl:SSL
> advertisedListeners={pulsar's listeners},internal:pulsar//192.168.1.10:9092,internal_ssl:pulsar//192.168.1.10:9093,external:pulsar//172.16.1.10:19002,external_ssl:pulsar://172.16.1.10:19003

1. Define the listenerNames and port of each listenerName in `kafkaListeners`
2. Define the listenerNames and protocol of each listenerName in `kafkaProtocalMap`
3. Add listenerNames and advertised address of each listenerName into `advertisedListeners`

Kafka client should connect to the port of the listenerName according to KIP-103.

PR #864, opened 2021-10-28
Support multiple listeners without configuring broker's advertisedListeners

BewareMyPower:
Fixes https://github.com/streamnative/kop/issues/818

### Motivation

https://github.com/streamnative/kop/pull/742 adds the multiple listeners support for KoP 2.8.x. However, there're some flaws. 

First, the KoP advertised listeners must be added to `advertisedListeners` config. This config should be treated as the advertised listeners for broker, not protocol handlers. What `KafkaProtocolHandler#getProtocolDataToAdvertise` returns is not used though it has been written to ZK. It also makes `kafkaAdvertisedListeners` meaningless and it was marked as deprecated.

### Modification

Benefit by [PIP 95](https://github.com/apache/pulsar/issues/12040), Pulsar client doesn't need to configure listener name for topic lookup. So this PR simplifies the implementation of `LookupClient` by using a shared `PulsarClient` without configuring listener name.

Then this PR uses the `MetadataStoreCacheLoader` introduced from https://github.com/streamnative/kop/pull/820 to get the `kafkaAdvertisedListeners` from ZK. Since it has already been cached, this PR removes the `KOP_ADDRESS_CACHE`.

To verify the feature, this PR adds a test `testMetadataRequestForMultiListeners`. The KoP config is like

```properties
kafkaListeners=PLAINTEXT://0.0.0.0:<port1>,GW://0.0.0.0:<port2>
kafkaProtocolMap=PLAINTEXT:PLAINTEXT,GW:PLAINTEXT
kafkaAdvertisedListeners=PLAINTEXT://192.168.0.1:<port3>,GW://192.168.0.1:<port4>
```

And verify that each `KafkaRequestHandler` can handle the METADATA requests well.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
@@ -184,6 +184,8 @@ public class KafkaServiceConfiguration extends ServiceConfiguration {
         category = CATEGORY_KOP,
         doc = "Comma-separated list of URIs we will listen on and the listener names.\n"
                 + "e.g. PLAINTEXT://localhost:9092,SSL://localhost:9093.\n"
+                + "Each URI's scheme represents a listener name if `kafkaProtocolMap` is configured.\n"
+                + "Otherwise, the scheme must be a valid protocol in [PLAINTEXT, SSL, SASL_PLAINTEXT, SASL_SSL].\n"
                 + "If hostname is not set, bind to the default interface."
     )
     private String kafkaListeners;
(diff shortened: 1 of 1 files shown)


tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaListenerNameTest.java
```diff
@@ -44,6 +63,72 @@ protected void cleanup() throws Exception {
         // Clean up in the test method
     }
 
+    @Test(timeOut = 30000)
+    public void testMetadataRequestForMultiListeners() throws Exception {
+        final Map<Integer, InetSocketAddress> bindPortToAdvertisedAddress = new HashMap<>();
+        final int anotherKafkaPort = PortManager.nextFreePort();
+        bindPortToAdvertisedAddress.put(kafkaBrokerPort,
```
Hongten, line 70: I encountered an error `Listener 'PLAINTEXT://192.168.0.1/<unresolved>:15007' is invalid` when calling `super.internalSetup()`. Did you have the same issue when you run this test?

PR #871, opened 2021-10-29
[Issue 867]: Implement minimal DeleteRecords (KStreams support)

### LINKED ISSUE #867: [FEATURE]  Implement DeleteRecords API (KafkaStreams support)
I am testing Kafka Streams and it looks like it is using DeleteRecords to clean up the environment.

This implementation is not a blocker for using Kafka Streams but it will enhance the compatibility

We could leverage the Topic Truncate API introduced in Pulsar 2.8.0 
https://github.com/apache/pulsar/pull/10326

but the Kafka API is more powerful as you can truncate a topic (partition) at any given offset.

for the case of Kafka Streams maybe it is enough to be able to simply truncate the topic to the end, I have to dig into KStreams implementation

eolivelli:
KafkaStreams uses the DeleteRecords API to truncate topics to the very end of the topic.

https://github.com/apache/kafka/blob/ba0ebca7a516d4179b6327ddc60b0b49b1265347/streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java#L441

This patch implements DeleteRecords, using the `truncateAsync` API in Pulsar introduced in 2.8.0.

we cannot make it work for arbitrary offsets, by the way this implementation is still useful, at least for KStreams or for Kafka Users who want to empty a topic.

Please note that Kafka Streams does not break without this feature, because it tries to delete the records in a best effort way.

Fixes #867

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AdminManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AdminManager.java
@@ -35,6 +35,8 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.bookkeeper.mledger.Position;
+import org.apache.bookkeeper.mledger.impl.PositionImpl;
 import org.apache.kafka.clients.admin.NewPartitions;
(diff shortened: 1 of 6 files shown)


BewareMyPower:
@Demogorgon314 could you help review this PR?

Demogorgon314 (approved):
LGTM.

BewareMyPower (approved):

PR #875, opened 2021-11-02
[Transaction] Migrate transaction coordinator handleInitProducerId tests and fix behavior

Demogorgon314:
### Motivation

Currently, KoP support transaction but don't have units test to cover transaction coordinator code behavior, for future maintain, we should add units test from  Kafka.

### Modification
* Fix `handleInitProducerId` behavior.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java
@@ -19,6 +19,7 @@ import static io.streamnative.pulsar.handlers.kop.coordinator.transaction.Transa
 import static io.streamnative.pulsar.handlers.kop.coordinator.transaction.TransactionState.PREPARE_EPOCH_FENCE;
 import static org.apache.pulsar.common.naming.TopicName.PARTITIONED_TOPIC_SUFFIX;
 
+import com.google.api.client.util.Sets;
 import com.google.common.annotations.VisibleForTesting;
 import io.streamnative.pulsar.handlers.kop.KopBrokerLookupManager;
 import io.streamnative.pulsar.handlers.kop.SystemTopicClient;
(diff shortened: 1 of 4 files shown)


Demogorgon314:
Ping @gaoran10 @BewareMyPower PTAL. Thanks.

gaoran10 (approved):
Nice work, LGTM

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java
```diff
@@ -230,48 +232,70 @@ public void handleInitProducerId(String transactionalId,
...
+                        });
+                        return null;
                     });
-                    return null;
-                });
+                } else {
+                    responseCallback.accept(initTransactionError(Errors.UNKNOWN_SERVER_ERROR));
+                }
+            }).exceptionally(ex -> {
+                responseCallback.accept(initTransactionError(Errors.forException(ex.getCause())));
```
gaoran10, line 297: Just a trivial comment, maybe we could add an error log here.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java
```diff
@@ -230,48 +232,70 @@ public void handleInitProducerId(String transactionalId,
...
-                            .transactionalId(transactionalId)
-                            .producerId(pid)
-                            .producerEpoch(producerEpoch)
-                            .state(TransactionState.EMPTY)
-                            .topicPartitions(new HashSet<>())
-                            .txnLastUpdateTimestamp(time.milliseconds())
-                            .build();
-                    epochAndTxnMetaFuture.complete(txnManager.putTransactionStateIfNotExists(newMetadata));
-                });
+                if (existMeta.hasErrors()) {
```
gaoran10, line 235: If the existing metadata is present and has errors, maybe we need to remove the existing meta from the cache? Normally, if the metadata of one transaction is present, it shouldn't have errors.

Demogorgon314: In this case metadata don't exists and has errors, we should complete the future by follow the kafka transaction behavior.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java
Demogorgon314: You'r right. Addressed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java
gaoran10: Ok, get it.

BewareMyPower (approved):

PR #885, opened 2021-11-05
[Transaction] Fix transaction coordinator handleAddPartitionsToTransaction behavior and migrate tests

Demogorgon314:
### Motivation

Currently, KoP support transaction but don't have units test to cover transaction coordinator `handleAddPartitionsToTransaction` behavior, for future maintain, we should add units test from Kafka.

**In addition**: `handleAddOffsetsToTxn ` should response  `AddOffsetsToTxnResponse`, previous implementation use `AddPartitionsToTxnResponse` to response.

### Modification

* Fix `handleAddPartitionsToTransaction` behavior.
* Fix `handleAddOffsetsToTxn ` wrong response.


Demogorgon314:
@BewareMyPower PTAL

BewareMyPower (approved):

PR #887, opened 2021-11-08
[Transaction] Fix transaction coordinator handleEndTransaction behavior and migrate tests

Demogorgon314:
### Motivation

Currently, KoP support transaction but don't have units test to cover transaction coordinator `handleEndTransaction ` behavior, for future maintain, we should add units test from Kafka.

### Modification

* Fix `handleEndTransaction ` behavior.


Demogorgon314:
@BewareMyPower @gaoran10 PTAL, Thanks.

BewareMyPower (approved):

Issue #789, opened 2021-10-06
[BUG] Kafka consumer not working - Failed to get offset


Raven888888:
**Describe the bug**
Following #765 , went on to upgrade pulsar & kop from `2.7.0` to `2.8.1`, according to the [official upgrade guide](https://pulsar.apache.org/docs/en/administration-upgrade/).

Had a few strange observations after the upgrade.

1. **KafkaRequestHandler failed to get offset**
In pulsar `broker.log`:
```
11:17:51.144 [pulsar-web-40-4] INFO  org.eclipse.jetty.server.RequestLog - 192.168.3.1 - - [06/Oct/2021:11:17:51 +0000] "GET /admin/v2/persistent/public/kafka/test/partitions HTTP/1.1" 200 16 "-" "Pulsar-Java-v2.8.1" 1
11:17:51.245 [BookKeeperClientWorker-OrderedExecutor-0-0] ERROR io.streamnative.pulsar.handlers.kop.KafkaRequestHandler - [PersistentTopic{topic=persistent://public/kafka/test-partition-0}] Failed to get offset for position 7306:0
java.lang.NullPointerException: null
```
Not sure where did the position `7306` come from, could have been persisted from before the upgrade. If so, do I need to clear the `__kafka/__consumer_offsets` table after the upgrade so that kafka starts afresh?

2. **Kafka client consumer console failed to get offset**
Client version: `2.12-2.8.0`
First, get the available offsets:
`apache-kafka/bin/kafka-run-class.sh kafka.tools.GetOffsetShell --broker-list 192.168.3.1:9092 --topic test`
got
`test:0:1088`
Then, consume at that offset:
`apache-kafka/bin/kafka-console-consumer.sh --topic test --offset 1088 --partition 0 --bootstrap-server 192.168.3.1:9092`
got error
```
[2021-10-06 17:27:58,594] WARN [Consumer clientId=consumer-console-consumer-44214-1, groupId=console-consumer-44214] Unknown server error while fetching offset 1088 for topic-partition test-0 (org.apache.kafka.clients.consumer.internals.Fetcher)
```
Since KoP server lives in pulsar broker, I tried to search in `broker.log` but did not find anything related.

3. **`__kafka` namespace topics content abnormal?**
Using pulsar sql,
.
`select * from pulsar."public/kafka"."test"; `
returns correct data exist in the table.
.
`select * from pulsar."public/__kafka"."__consumer_offsets";`
[...]

PR #907, opened 2021-11-15
Skip messages without BrokerEntryMetadata

BewareMyPower:
### Motivation

When users upgrade from KoP (< 2.8.0) to KoP (>= 2.8.0), the topics produced by KoP (< 2.8.0) cannot be used, like https://github.com/streamnative/kop/issues/789. Therefore, the upgrade requires old version KoP users to delete these topics and create again if they want to reuse the same topic name.

This PR intends to skip these old messages that don't have a BrokerEntryMetadata part so that when users upgrade from KoP (< 2.8.0) to KoP (>= 2.8.0), the topics can be reused and the old messages will be skipped automatically.

### Modifications

- Add `MessageMetadataUtils#asyncFindPosition` that leverages `ManagedLedger#asyncFindPosition` to find the position associated with the offset. The most significant change is that `true` will be returned if the entry doesn't contain the BrokerEntryMetadata. Because it means the entry is older than the target entry in binary search.
- In `MessageMetadataUtils#getOffsetOfPosition`, return the 0 offset if the entry doesn't has the BrokerEntryMetadata.
- Add an `UpgradeTest`. First disable the `BrokerEntryMetadataInterceptor` and produce some messages, then restart the broker with `BrokerEntryMetadataInterceptor` enabled and produce some messages. There are 5 topics for test to cover different cases in binary search. And the tests cover the three cases of `LIST_OFFSETS` request (earliest, latest, timestamp)
  - `testOffsetsInSendCallback`: Verify the offsets in send callback are correct. When BrokerEntryMetadata is not enabled, the offset should be -1.
  - `testConsumeEarliest`: Run Kafka consumers to consume messages from earliest and verify the old messages are skipped.
  - `testConsumeLatest`: Run Kafka consumers to consume messages from latest. This tests's priority is lower because a new message will be sent in this test and it will affect other tests.
  - `testConsumeEarliest`: Verity listing offsets by timestamp work with corrupted old messages.


tests/src/test/java/io/streamnative/pulsar/handlers/kop/KopProtocolHandlerTestBase.java
```diff
@@ -329,24 +335,28 @@ protected final void internalCleanup() throws Exception {
...
-            if (mockZooKeeper != null) {
-                mockZooKeeper.shutdown();
-            }
             if (sameThreadOrderedSafeExecutor != null) {
                 sameThreadOrderedSafeExecutor.shutdown();
             }
-            if (bkExecutor != null) {
-                bkExecutor.shutdown();
-            }
+            cleanupStorage();
```
Demogorgon314, line 341: Why did we `cleanupStorage()` here ?  I think your original purpose is split broker and storage cleanups right?

BewareMyPower: It's right. I'll remove this line and try to verify the old data still exists after restart.

BewareMyPower:
TODO: Add an option to enable this feature

BewareMyPower:
After confirming with @codelipenghui, I think we don't need an extra option to enable skipping messages. So this PR is ready to review, PTAL @Demogorgon314

Demogorgon314 (approved):
LGTM

PR #389, opened 2021-03-01
add prometheus metric exporter framework

hangc0276:
### Motivation
KOP doesn't expose prometheus metrics to monitor running stats, which will be hard the debug performance problem.

### Changs
1. add Prometheus metrics provider for kop and use individual port to expose kop metrics
2. Due to BookKeeper Prometheus metric provider include basic metrics stats, such as jvm metrics, memory metrics, cpu metric etc, which will lead those metrics register twice. When run kop with pulsar broker, the prometheus metric thread will start failed.

PR #911, opened 2021-11-16
[Docs] Add Monitor KoP link to main page

BewareMyPower:
### Motivation

https://github.com/streamnative/kop/pull/389 exposed the metrics for KoP and added the docs. However, users cannot click into the metrics docs in the main web page.

### Modifications

Add the link of metrics docs to README.md. In addition, sync docs/kop.md with README.md so that upgrade.md and reference-metrics.md can both be navigated from SN hub. Then remove the **Upgrade** section because it's duplicated with upgrade.md.


BewareMyPower:
PTAL @Huanli-Meng

Huanli-Meng (approved):
LGTM. One more question, should we sync KoP Readme to SN hub? if need. We need to submit a PR in SN hub repo.

BewareMyPower:
@Huanli-Meng It's better not. Because it requires the docs using `{{protocol:version}}`, which leads to some wrong links in github pages.

PR #912, opened 2021-11-16
[Transaction] Support producer state manager

Demogorgon314:
Master issues: #915

This PR is a copy from #796, rebase from the master branch. And fix all tests. 

### Motivation
KoP doesn't support duplicate sequence number checks, it's used for the idempotent producers.

Each topic should have a `ProducerStateManager`, the `ProducerStateManager` will ensure idempotent when a producer has produced.

### Modifications
**Note**: Current implement just in-memory support, it should be persisted to some storage. And we only support `MAX_IN_FLIGHT_REQUESTS_PER_CONNECTION = 1` when using idempotent producer.

1. Support in memory producer state manager
2. Support idempotent producer
3. Add new units test `ProducerStateManagerTest`

**Addition change after #796**

1. Use exists `ReplicaManager#appendRecords` to write the transaction markers.
2. Refactor make the code cleaner.
3. Remove unused aborted Index logic in transaction coordinator.

### TODO
* Support Producer ID Expiration
* Support snapshot
* Support `MAX_IN_FLIGHT_REQUESTS_PER_CONNECTION = 5` when using idempotent producer(Is hard to support it, since KoP uses async message publishing, is hard to add lock in this method).


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/ProducerStateManager.java
```diff
@@ -0,0 +1,1048 @@
...
+            if (origin.equals(AppendOrigin.Client)) {
+                checkSequence(producerEpoch, firstSeq);
+            }
+        }
+
+        private void checkProducerEpoch(Short producerEpoch) {
+            if (producerEpoch < updatedEntry.producerEpoch) {
+                String message = String.format("Producer's epoch in %s is %s, which is smaller than the last seen "
+                        + "epoch %s", topicPartition, producerEpoch, currentEntry.producerEpoch);
+                throw new IllegalArgumentException(message);
```
BewareMyPower, line 436: Though current Kafka 2.0 dependency doesn't have the `InvalidProducerEpochException`, but we can implement it. Then in `KafkaRequestHandler#handlePartitionRecords`, we can catch this exception and return an `INVALID_PRODUCER_EPOCH` error. BTW, I think we should call `Errors.forException` in the catch block instead of just calling `completeExceptionally`.

BewareMyPower (commented):
Just have a simple review of this PR. At the moment I have only one suggestion that it's better to use class without access modifiers than static inner class.

We have already made some bad designs in KoP (and also in Pulsar). For example, `KafkaCommandDecoder#ResponseAndRequest`. The outer class can access any field of the inner class directly without calling any getter. The encapsulation (getters & setters) is corrupted.

Look back to this PR. The definitions of `SystemTopicProducerStateWriter` and `SystemTopicProducerStateReader` are both inside the `SystemTopicProducerStateClient` body, which makes code hard to read. These inner classes only access the `TOPIC_NAME_PROP` field. We can make it public and the inner classes outside the definition of `SystemTopicProducerStateClient`. However, for this case, a better solution is to change them to non-static inner classes because each of them holds a reference to the outer class.

```java
    @AllArgsConstructor
    public class SystemTopicProducerStateReader implements Reader<ByteBuffer> {

        private final org.apache.pulsar.client.api.Reader<ByteBuffer> reader;

        /* ... */

        @Override
        public SystemTopicClient<ByteBuffer> getSystemTopic() {
            return SystemTopicProducerStateClient.this;
        }
    }
```

The same goes for the inner classes of `ProducerStateManager` but it looks like some inner classes are used in tests and we need to access them in the same package. It's better to eliminate the usage of inner static classes IMO.

If you have some questions, ping me directly.

BewareMyPower:
IMO, this PR becomes too huge to review. (2,736 additions and 270 deletions)

I see in #915, take snapshot, recover and expiration are all independent tasks. Could you separate the related code into other PRs?

For example,
- `ProducerStateManager#takeSnapshot` is only called in `ProducerStateManagerTest#takeSnapshotAndWait`, which is used in `testTruncateAndReloadRemovesOutOfRangeSnapshots`.
- `ProducerStateManager#loadFromSnapshot` is only called in `ProducerStateManager#recover`, which is never called.

And I see you moved some classes outside the class. But it's not what I intended. I mean you can make these classes only visible for package only.

See these two commits of [my branch](https://github.com/BewareMyPower/kop/commits/TEST-PR-912) for more details.
- Remove snapshot and recover logic
- Don't expose some classes
- Remove expiration related code

Even in my branch, there are still 1,745 additions and 271 deletions, which could still take some time to review. But we can focus on the migration of producer state manager itself and idempotent producer. It would be great if you can still reduce the changes based on my branch.

BTW, I have some suggestions on the `SystemTopicProducerStateClient` but I think it's better to be discussed in another PR.

Demogorgon314:
@BewareMyPower 
Thank you for your suggestion, I addressed all you review suggestions and removed all unused code in this PR.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java
```diff
@@ -42,47 +47,157 @@
...
+                // compute the last stable offset without relying on additional index lookups.
+                Optional<CompletedTxn> maybeCompletedTxn =
+                        updateProducers(batch, updatedProducers, firstOffset, origin);
+                maybeCompletedTxn.ifPresent(completedTxns::add);
+            }
+        }
+        return new AnalyzeResult(updatedProducers, completedTxns, Optional.empty());
+    }
+
+    public void append(AnalyzeResult analyzeResult, long startOffset, LogAppendInfo appendInfo) {
```
BewareMyPower, line 137: It should be a private method. In addition, this method name is confused. It should be a callback, but `append` is similar to `appendRecords`. I recommend the name `publishComplete` and move the other logic into this method. 

Actually, I think there is no need to add a new method. I tried to add the code into `offsetFuture.whenComplete` directly like

```java
                final long lastOffset = offset + numMessages - 1;
                analyzeResult.updatedProducers().forEach((pid, producerAppendInfo) -> {
                    if (log.isDebugEnabled()) {
                        log.debug("Append pid: [{}], appendInfo: [{}], lastOffset: [{}]",
                                pid, producerAppendInfo, lastOffset);
                    }
                    // When we have real start offset, update current txn first offset.
                    producerAppendInfo.updateCurrentTxnFirstOffset(appendInfo.isTransaction(), offset);
                    producerStateManager.update(producerAppendInfo);
                });
                analyzeResult.completedTxns().forEach(completedTxn -> {
                    // update to real last offset
                    completedTxn.lastOffset(lastOffset);
                    long lastStableOffset = producerStateManager.lastStableOffset(completedTxn);
                    producerStateManager.updateTxnIndex(completedTxn, lastStableOffset);
                    producerStateManager.completeTxn(completedTxn);
                });
```

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/ProducerAppendInfo.java
```diff
@@ -0,0 +1,222 @@
...
+        updatedEntry.addBatch(epoch, lastSeq, lastOffset, lastSeq - firstSeq, lastTimestamp);
+    }
+
+    /**
+     * When message publish success, we should save the current txn first offset.
+     *
+     * @param isTransactional is transaction or not
+     * @param firstOffset transaction first offset
+     */
+    public void updateCurrentTxnFirstOffset(Boolean isTransactional, long firstOffset) {
```
BewareMyPower, line 171: I didn't find this method in Kafka 2.8.1. It looks like it's inside the `ProducerAppendInfo#appendDataBatch` method in Kafka.

Demogorgon314: The reason why I add this method is that we can't get a real offset when doing `ProducerAppendInfo#appendDataBatch`. In KoP we must update the txn first offset after append message into the storage. Otherwise, the aborted index might be wrong.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java
```diff
@@ -42,47 +47,157 @@
...
+
+        for (RecordBatch batch : records.batches()) {
+            if (batch.hasProducerId()) {
+                if (origin.equals(AppendOrigin.Client)) {
+                    Optional<ProducerStateEntry> maybeLastEntry =
+                            producerStateManager.lastEntry(batch.producerId());
+
+                    // if this is a client produce request,
+                    // there will be up to 5 batches which could have been duplicated.
+                    // If we find a duplicate, we return the metadata of the appended batch to the client.
```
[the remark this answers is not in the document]
Demogorgon314: Sure, Addressed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java
```diff
@@ -196,52 +314,103 @@ private void publishMessages(final Optional<PersistentTopic> persistentTopicOpt,
...
+
+            int batchSize = batch.sizeInBytes();
+            if (batchSize > kafkaConfig.getMaxMessageSize()) {
+                throw new RecordTooLargeException(String.format("Message batch size is %s "
+                                + "in append to partition %s which exceeds the maximum configured size of %s .",
+                        batchSize, topicPartition, kafkaConfig.getMaxMessageSize()));
+            }
+            if (batch.hasProducerId()) {
+                hasProducerId = true;
+            }
```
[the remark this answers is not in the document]
Demogorgon314: Removed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java
```diff
@@ -196,52 +314,103 @@ private void publishMessages(final Optional<PersistentTopic> persistentTopicOpt,
...
+        int shallowMessageCount = 0;
+        long lastOffset = -1L;
+        Optional<Long> firstOffset = Optional.empty();
+        long lastOffsetOfFirstBatch = -1L;
+        boolean readFirstMessage = false;
+        boolean monotonic = true;
+        boolean hasProducerId = false;
+        boolean isTransaction = false;
 
+        validateRecords(version, records);
```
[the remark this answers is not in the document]
Demogorgon314: Addressed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java
Demogorgon314: Addressed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -2130,92 +2165,106 @@ protected void handleEndTxn(KafkaHeaderAndRequest kafkaHeaderAndRequest,
...
+                    kafkaConfig.getRequestTimeoutMs(),
+                    true,
+                    currentNamespacePrefix(),
+                    controlRecords,
+                    PartitionLog.AppendOrigin.Coordinator,
+                    appendRecordsContext
+            ).whenComplete((result, ex) -> {
+                appendRecordsContext.recycle();
+                if (ex != null) {
+                    log.error("Append txn marker failed.", ex);
```
BewareMyPower, line 2212:
```suggestion
                    log.error("[{}] Append txn marker ({}) failed.", ctx.channel(), marker, ex);
```

Maybe we can also add a debug log when `ex == null`.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/IdempotentProducerTest.java
```diff
@@ -0,0 +1,108 @@
...
+        props.put(ProducerConfig.ENABLE_IDEMPOTENCE_CONFIG, "true");
+        props.put(ProducerConfig.ACKS_CONFIG, "all");
+        props.put(ProducerConfig.MAX_IN_FLIGHT_REQUESTS_PER_CONNECTION, 1);
+        props.put(CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG, "localhost:" + kafkaBrokerPort);
+        props.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, StringSerializer.class.getName());
+        props.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, StringSerializer.class.getName());
+        props.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class.getName());
+        props.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class.getName());
+        props.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+        props.put(ConsumerConfig.GROUP_ID_CONFIG, "test_group");
```
BewareMyPower, line 83: Could you avoid sharing a common `Properties`?

Here is an example:

```java
        final Properties producerProps = newKafkaProducerProperties();
        producerProps.put(ProducerConfig.ENABLE_IDEMPOTENCE_CONFIG, true);
        producerProps.put(ProducerConfig.MAX_IN_FLIGHT_REQUESTS_PER_CONNECTION, 1);

        @Cleanup
        KafkaProducer<String, String> producer = new KafkaProducer<>(producerProps)

        @Cleanup
        KafkaConsumer<String, String> consumer = new KafkaConsumer<>(newKafkaConsumerProperties());
```

BTW, I'm not sure whether `acks` should be configured.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaAuthorizationTestBase.java
```diff
@@ -202,7 +202,8 @@ public void testAuthorizationSuccess() throws PulsarAdminException {
```
BewareMyPower, line 202: Are the changes of this file and `KafkaNonPartitionedTopicTest.java` related to this PR?

tests/src/test/java/io/streamnative/pulsar/handlers/kop/compatibility/saslplain/SaslPlainEndToEndTest.java
```diff
@@ -265,7 +265,6 @@ void testKafkaProduceAndConsumeWithSaslPlain() throws Exception {
 
             // ensure that we can list the topic
             Map<String, List<PartitionInfo>> result = consumer.listTopics(1000);
-            assertEquals(result.size(), 1);
```
BewareMyPower, line 268: The same question as I asked in previous comment. Why did you remove this check?

tests/src/test/java/io/streamnative/pulsar/handlers/kop/storage/ProducerStateManagerTest.java
```diff
@@ -0,0 +1,582 @@
...
+        } catch (OutOfOrderSequenceException e) {
+            log.info("Expected behavior.");
+        }
+
+        // Invalid sequence number (greater than next expected sequence number)
+        try {
+            append(stateManager, producerId, epoch, 5, 0L, 2L, false);
+            fail("Invalid sequence number (greater than next expected sequence number).");
+        } catch (OutOfOrderSequenceException e) {
+            log.info("Expected behavior.");
```
BewareMyPower, line 103: Please remove all "Expected behavior." logs because they don't help at all.

And I think it's better to use `assertThrows` for this case, like

```java
        assertThrows(OutOfOrderSequenceException.class, () ->
                append(stateManager, producerId, epoch, 5, 0L, 2L, false));
```

tests/src/test/java/io/streamnative/pulsar/handlers/kop/storage/ProducerStateManagerTest.java
```diff
@@ -0,0 +1,582 @@
...
+        assertEquals(0, secondEntry.lastSeq().intValue());
+    }
+
+    @Test(timeOut = defaultTestTimeout, expectedExceptions = OutOfOrderSequenceException.class)
+    public void testProducerSequenceInvalidWrapAround() {
+        short epoch = 15;
+        Integer sequence = Integer.MAX_VALUE;
+        long offset = 735L;
+        append(stateManager, producerId, epoch, sequence, offset);
+        append(stateManager, producerId, epoch, 1, offset + 500);
```
BewareMyPower, line 172: I found the original test in Kafka is

```scala
    append(stateManager, producerId, epoch, sequence, offset, origin = AppendOrigin.Replication)
    assertThrows(classOf[OutOfOrderSequenceException], () => append(stateManager, producerId, epoch, 1, offset + 500)
```

The first `append` should not throw exceptions. Could you remove `expectedExceptions` and use `assertThrows` instead?

tests/src/test/java/io/streamnative/pulsar/handlers/kop/storage/ProducerStateManagerTest.java
```diff
@@ -0,0 +1,582 @@
...
+        long offset = 735L;
+        append(stateManager, producerId, epoch, sequence, offset);
+
+        Optional<ProducerStateEntry> maybeLastEntry = stateManager.lastEntry(producerId);
+        assertTrue(maybeLastEntry.isPresent());
+
+        ProducerStateEntry lastEntry = maybeLastEntry.get();
+        assertEquals(epoch, lastEntry.producerEpoch().shortValue());
+        assertEquals(sequence, lastEntry.firstSeq());
+        assertEquals(sequence, lastEntry.lastSeq());
```
BewareMyPower, line 188: The original test in Kafka has following assertions:

```scala
    assertEquals(offset, lastEntry.lastDataOffset)
    assertEquals(offset, lastEntry.firstDataOffset)
```

I found in KoP, the `lastDataOffset` and `firstDataOffset` are both -1L. Could you explain it?

BewareMyPower:
Now I have reviewed nearly all the changes except some tests in `ProducerStateManagerTest`, PTAL.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/ProducerStateManager.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    public void completeTxn(CompletedTxn completedTxn) {
+        TxnMetadata txnMetadata = ongoingTxns.remove(completedTxn.firstOffset());
+        if (txnMetadata == null) {
+            String msg = String.format("Attempted to complete transaction %s on partition "
+                    + "%s which was not started.", completedTxn, topicPartition);
+            throw new IllegalArgumentException(msg);
+        }
+        txnMetadata.lastOffset(completedTxn.lastOffset());
```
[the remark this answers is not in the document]
Demogorgon314: The `unreplicatedTxns` is useless in KoP. We can remove it.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -2130,92 +2165,106 @@ protected void handleEndTxn(KafkaHeaderAndRequest kafkaHeaderAndRequest,
...
+                Map<TopicPartition, Errors> currentErrors = new HashMap<>();
+                result.forEach(((topicPartition, partitionResponse) -> {
+                    currentErrors.put(topicPartition, partitionResponse.error);
+                }));
+                updateErrors.accept(producerId, currentErrors);
+
+                Set<TopicPartition> successfulOffsetsPartitions = result.keySet()
+                        .stream()
+                        .filter(topicPartition ->
+                                TopicName.get(topicPartition.topic()).getLocalName().equals(GROUP_METADATA_TOPIC_NAME))
```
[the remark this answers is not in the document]
Demogorgon314: This case can pass, but when the topic has a partition suffix, it doesn't work (In this we never get a partitioned topic), I changed it to the below code for more rigorous:
```java
String partitionedTopicName = TopicName.get(fullTopicName).getPartitionedTopicName();
partitionedTopicName.endsWith("/" + GROUP_METADATA_TOPIC_NAME);
```

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaAuthorizationTestBase.java
Demogorgon314: The previous change introduced a new system topic, when we do the `listTopics`, the system topic is in the result, is should add in the next PR. We can remove it for now.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/storage/ProducerStateManagerTest.java
Demogorgon314: In KoP we don't use the offset now, since we just use the seq to check order

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
```diff
@@ -410,9 +410,10 @@ private void handleEntries(final List<Entry> entries,
         final long highWatermark = MessageMetadataUtils.getHighWatermark(cursor.getManagedLedger());
         // Add new offset back to TCM after entries are read successfully
         tcm.add(cursorOffset.get(), Pair.of(cursor, cursorOffset.get()));
-
+        PartitionLog partitionLog =
+                requestHandler.getReplicaManager().getPartitionLog(topicPartition, namespacePrefix);
         final long lso = (readCommitted
-                ? tc.getLastStableOffset(TopicName.get(fullTopicName), highWatermark) : highWatermark);
+                ? partitionLog.firstUndecidedOffset().orElse(highWatermark) : highWatermark);
```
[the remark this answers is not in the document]
Demogorgon314: Sure. removed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -2130,92 +2165,106 @@ protected void handleEndTxn(KafkaHeaderAndRequest kafkaHeaderAndRequest,
...
+                        .collect(Collectors.toSet());
+                if (successfulOffsetsPartitions.size() > 0) {
```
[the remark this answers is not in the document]
Demogorgon314: Addressed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
Demogorgon314: Sure, Addressed.

BewareMyPower (approved):

PR #925, opened 2021-11-23
Split the message storage logic into independent classes

Demogorgon314:
### Motivation

Currently, the KoP store message logic is in `KafkaRequestHandler`, to avoid duplicate code and easier to support idempotent produce, we must refactor the message storage logic.

### Modifications
* Refactor message storage logic.
* The `ReplicaManager` is a public interface to store messages, we should always use `ReplicaManager` to store the message in any situation.

* `PartitionLog` mapping to Kafka is `Log`, the reason for using this name is to avoid naming conflict, like: `log.debug()`.

* `PartitionLogManager` hold all `PartitionLog`, key is full partition name, value is `PartitionLog`.

### TODO
* Add fetch operation

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
@@ -132,6 +132,7 @@ public class KafkaProtocolHandler implements ProtocolHandler, TenantContextManag
                 throw new IllegalStateException(e);
             }
             return new ReplicaManager(
+                    kafkaConfig,
                     Time.SYSTEM,
                     entryFormatter,
                     transactionCoordinatorOptional,
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
@@ -26,6 +26,8 @@ import io.streamnative.pulsar.handlers.kop.coordinator.group.GroupCoordinator;
 import io.streamnative.pulsar.handlers.kop.coordinator.group.OffsetConfig;
 import io.streamnative.pulsar.handlers.kop.coordinator.transaction.TransactionConfig;
 import io.streamnative.pulsar.handlers.kop.coordinator.transaction.TransactionCoordinator;
+import io.streamnative.pulsar.handlers.kop.format.EntryFormatter;
+import io.streamnative.pulsar.handlers.kop.format.EntryFormatterFactory;
 import io.streamnative.pulsar.handlers.kop.stats.PrometheusMetricsProvider;
 import io.streamnative.pulsar.handlers.kop.stats.StatsLogger;
 import io.streamnative.pulsar.handlers.kop.storage.ReplicaManager;
@@ -122,15 +124,18 @@ public class KafkaProtocolHandler implements ProtocolHandler, TenantContextManag
             if (kafkaConfig.isEnableTransactionCoordinator()) {
                 transactionCoordinatorOptional = Optional.of(getTransactionCoordinator(tenant));
             }
+            EntryFormatter entryFormatter;
             try {
-                return new ReplicaManager(kafkaConfig,
-                        Time.SYSTEM,
-                        transactionCoordinatorOptional,
-                        producePurgatory);
-            } catch (Exception e) {
-                log.error("Failed to init ReplicaManager for tenant {}", tenant, e);
+                entryFormatter = EntryFormatterFactory.create(kafkaConfig);
+            } catch (IllegalArgumentException e) {
+                log.error("Failed to init create enter formatter {}", tenant, e);
                 throw new IllegalStateException(e);
             }
+            return new ReplicaManager(kafkaConfig,
+                    Time.SYSTEM,
+                    entryFormatter,
+                    transactionCoordinatorOptional,
+                    producePurgatory);
         });
     }
 
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
@@ -28,6 +28,7 @@ import io.streamnative.pulsar.handlers.kop.coordinator.transaction.TransactionCo
 import io.streamnative.pulsar.handlers.kop.coordinator.transaction.TransactionCoordinator;
 import io.streamnative.pulsar.handlers.kop.stats.PrometheusMetricsProvider;
 import io.streamnative.pulsar.handlers.kop.stats.StatsLogger;
+import io.streamnative.pulsar.handlers.kop.storage.ReplicaManager;
 import io.streamnative.pulsar.handlers.kop.utils.ConfigurationUtils;
 import io.streamnative.pulsar.handlers.kop.utils.MetadataUtils;
 import io.streamnative.pulsar.handlers.kop.utils.delayed.DelayedOperation;
@@ -97,6 +98,7 @@ public class KafkaProtocolHandler implements ProtocolHandler, TenantContextManag
 
     private final Map<String, GroupCoordinator> groupCoordinatorsByTenant = new ConcurrentHashMap<>();
     private final Map<String, TransactionCoordinator> transactionCoordinatorByTenant = new ConcurrentHashMap<>();
+    private final Map<String, ReplicaManager> replicaManagerByTenant = new ConcurrentHashMap<>();
 
     @Override
     public GroupCoordinator getGroupCoordinator(String tenant) {
@@ -111,6 +113,18 @@ public class KafkaProtocolHandler implements ProtocolHandler, TenantContextManag
     @Override
     public TransactionCoordinator getTransactionCoordinator(String tenant) {
         return transactionCoordinatorByTenant.computeIfAbsent(tenant, this::createAndBootTransactionCoordinator);
+    }
+
+    @Override
+    public ReplicaManager getReplicaManager(String tenant) {
+        return replicaManagerByTenant.computeIfAbsent(tenant, s -> {
+            try {
+                return new ReplicaManager(kafkaConfig, Time.SYSTEM, producePurgatory, fetchPurgatory);
+            } catch (Exception e) {
+                log.error("Failed to init ReplicaManager for tenant {}", tenant, e);
+                throw new IllegalStateException(e);
+            }
+        });
     }
 
     /**
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
@@ -131,7 +131,7 @@ public class KafkaProtocolHandler implements ProtocolHandler, TenantContextManag
                 log.error("Failed to init create enter formatter {}", tenant, e);
                 throw new IllegalStateException(e);
             }
-            return new ReplicaManager(kafkaConfig,
+            return new ReplicaManager(
                     Time.SYSTEM,
                     entryFormatter,
                     transactionCoordinatorOptional,
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
@@ -126,8 +126,7 @@ public class KafkaProtocolHandler implements ProtocolHandler, TenantContextManag
                 return new ReplicaManager(kafkaConfig,
                         Time.SYSTEM,
                         transactionCoordinatorOptional,
-                        producePurgatory,
-                        fetchPurgatory);
+                        producePurgatory);
             } catch (Exception e) {
                 log.error("Failed to init ReplicaManager for tenant {}", tenant, e);
                 throw new IllegalStateException(e);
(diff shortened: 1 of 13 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
```diff
@@ -113,6 +115,25 @@ public TransactionCoordinator getTransactionCoordinator(String tenant) {
...
+            }
+            try {
+                return new ReplicaManager(kafkaConfig,
+                        Time.SYSTEM,
+                        transactionCoordinatorOptional,
+                        producePurgatory);
+            } catch (Exception e) {
+                log.error("Failed to init ReplicaManager for tenant {}", tenant, e);
+                throw new IllegalStateException(e);
+            }
```
BewareMyPower, line 133: Remove the try-catch block. The exception could only be thrown from `EntryFormatter#create`. If you want to fail fast, you can create the `EntryFormatter` in advance.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -894,231 +893,84 @@ private void completeSendOperationForThrottling(long msgSize) {
...
+                ).thenAccept(response -> {
+                    appendRecordsContext.recycle();
+                    Map<TopicPartition, PartitionResponse> mergedResponse = new HashMap<>();
+                    mergedResponse.putAll(response);
+                    mergedResponse.putAll(unauthorizedTopicResponsesMap);
+                    resultFuture.complete(new ProduceResponse(mergedResponse));
+                }).exceptionally(ex -> {
+                    appendRecordsContext.recycle();
+                    resultFuture.completeExceptionally(ex.getCause());
+                    return null;
```
BewareMyPower, line 942: If you are sure that no exceptions could be thrown from `thenAccept` block, using `whenComplete` is better.

Otherwise, if exceptions were thrown from `thenAccept` block, `appendRecordsContext.recycle()` would be called twice.

BewareMyPower:
**NOTE: I'll go out for a while and the review is not completed. You can address these comments first.**

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
Demogorgon314: You're right. Addressed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
Demogorgon314: Good idea.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -894,231 +893,84 @@ private void completeSendOperationForThrottling(long msgSize) {
...
-            }
-            // add the topicPartition with timeout error if it's not existed in responseMap
-            produceRequest.partitionRecordsOrFail().keySet().forEach(topicPartition -> {
-                if (!responseMap.containsKey(topicPartition)) {
-                    responseMap.put(topicPartition, new PartitionResponse(Errors.REQUEST_TIMED_OUT));
+        String namespacePrefix = currentNamespacePrefix();
+        final AtomicInteger unfinishedAuthorizationCount = new AtomicInteger(numPartitions);
+        Consumer<Runnable> completeOne = (action) -> {
+            // When complete one authorization or failed, will do the action first.
+            action.run();
```
BewareMyPower, line 914: I think the `action` parameter here is not necessary. Instead of calling

```java
completeOne.accept(() -> someMap.put(someKey, someValue));
```

, I think the following code is more clear.

```java
someMap.put(someKey, someValue)
completeOne.run();
```

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java
```diff
@@ -0,0 +1,332 @@
...
+                                                 final short version,
+                                                 final AppendRecordsContext appendRecordsContext) {
+        return append(records, version, false, appendRecordsContext);
+    }
+
+    /**
+     * Append this message to pulsar.
+     *
+     * This method will generally be responsible for assigning offsets to the messages,
+     * however if the assignOffsets=false flag is passed we will only check that the existing offsets are valid.
```
BewareMyPower, line 98: It looks like these two lines were copied from Kafka? Here it's a little different in KoP. We don't have `assignOffsets=false` flag. Instead, we only allows `MixedKafkaEntryFormatter` to assign offsets and the original `KafkaV1EntryFormatter` (it actually also supports V0) doesn't need to assign offsets.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/KopTopic.java
```diff
@@ -103,4 +106,10 @@ public static String toString(String topic, int partition, String namespacePrefi
         return (new KopTopic(topic, namespacePrefix)).getPartitionName(partition);
     }
 
+    public static boolean isInternalTopic(final String fullTopicName) {
+        String partitionedTopicName = TopicName.get(fullTopicName).getPartitionedTopicName();
+        return partitionedTopicName.endsWith("/" + GROUP_METADATA_TOPIC_NAME)
+                || partitionedTopicName.endsWith("/" + TRANSACTION_STATE_TOPIC_NAME);
+    }
```
BewareMyPower, line 113: `KafkaRequestHandler#isInternalTopic` is still used in `handleTopicMetadataRequest`, please reuse it.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java
```diff
@@ -0,0 +1,332 @@
...
+                // Append Message into pulsar
+                final CompletableFuture<Optional<PersistentTopic>> topicFuture =
+                        topicManager.getTopic(fullPartitionName);
+                if (topicFuture.isCompletedExceptionally()) {
+                    topicFuture.exceptionally(e -> {
+                        appendFuture.completeExceptionally(e);
+                        return Optional.empty();
+                    });
+                    return appendFuture;
+                }
```
BewareMyPower, line 129: When we refactor the code, please make sure the behavior is nearly the same as before. I see you copied Kafka's logic here, like calling `analyzeAndValidateRecords` and `trimInvalidBytes`, even the `lock`. Maybe some are right. However, we should not do that in this PR. This PR should migrate the logic from the current KoP source.

Demogorgon314: Ok, I removed it.

Maybe I'll add this logic in ProducerStateManager's PR.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/KopTopic.java
Demogorgon314: Sure. Addressed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java
Demogorgon314: Addressed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
Demogorgon314: You're right. Addressed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java
```diff
@@ -0,0 +1,257 @@
...
+@Slf4j
+@AllArgsConstructor
+public class PartitionLog {
+    private KafkaServiceConfiguration kafkaConfig;
+    private Time time;
+    private TopicPartition topicPartition;
+    private String namespacePrefix;
+    private String fullPartitionName;
+    private EntryFormatter entryFormatter;
+    private Optional<TransactionCoordinator> transactionCoordinator;
```
BewareMyPower, line 61: Make these fields `final`.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java
```diff
@@ -0,0 +1,257 @@
...
+            if (topicFuture.isDone() && !topicFuture.getNow(Optional.empty()).isPresent()) {
+                appendFuture.completeExceptionally(Errors.NOT_LEADER_FOR_PARTITION.exception());
+                return appendFuture;
+            }
+            final Consumer<Optional<PersistentTopic>> persistentTopicConsumer = persistentTopicOpt -> {
+                if (!persistentTopicOpt.isPresent()) {
+                    appendFuture.completeExceptionally(Errors.NOT_LEADER_FOR_PARTITION.exception());
+                    return;
+                }
+                // TODO: validateMessagesAndAssignOffsets here.
```
BewareMyPower, line 111: `validateMessagesAndAssignOffsets` is called in `MixedKafkaEntryFormatter#encode`.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java
```diff
@@ -0,0 +1,257 @@
...
+    private String namespacePrefix;
+    private String fullPartitionName;
+    private EntryFormatter entryFormatter;
+    private Optional<TransactionCoordinator> transactionCoordinator;
+
+    public CompletableFuture<Long> appendRecords(final MemoryRecords records,
+                                                 final short version,
+                                                 final AppendRecordsContext appendRecordsContext) {
+        return append(records, version, appendRecordsContext);
+    }
```
BewareMyPower, line 67: Should `appendRecords` be replaced by `append`? Nothing more is done in `appendRecords`.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java
```diff
@@ -0,0 +1,257 @@
...
+                final EncodeResult encodeResult = entryFormatter.encode(encodeRequest);
+                encodeRequest.recycle();
+                requestStats.getProduceEncodeStats().registerSuccessfulEvent(
+                        time.nanoseconds() - beforeRecordsProcess, TimeUnit.NANOSECONDS);
+                appendRecordsContext.getStartSendOperationForThrottling()
+                        .accept(encodeResult.getEncodedByteBuf().readableBytes());
+                if (log.isDebugEnabled()) {
+                    log.debug("Produce messages for topic {} partition {}",
+                            topicPartition.topic(), topicPartition.partition());
+                }
```
BewareMyPower, line 129: Actually I think this debug logs doesn't help. The original debug logs are

```java
                if (log.isDebugEnabled()) {
                    log.debug("[{}] Request {}: Produce messages for topic {} partition {}, "
                                    + "request size: {} ", ctx.channel(), produceHar.getHeader(),
                            topicPartition.topic(), topicPartition.partition(), numPartitions);
                }
```

BTW, the `numPartitions` is wrong, it should not represent the request size.

I think you can just remove these lines or pass the channel context and request header to `AppendRecordsContext`.

BewareMyPower:
LGTM but left some minor comments.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java
Demogorgon314: In idempotent producer, we need to assign offset here first and we need to support all formatter.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java
Demogorgon314: Sure. Addressed.

BewareMyPower (approved):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -894,231 +883,80 @@ private void completeSendOperationForThrottling(long msgSize) {
...
-                    topicPartition, new PartitionResponse(Errors.NONE, offset, -1L, -1L));
-            final Consumer<Errors> errorsConsumer =
-                    errors -> addPartitionResponse.accept(topicPartition, new PartitionResponse(errors));
-            final Consumer<Throwable> exceptionConsumer =
-                    e -> addPartitionResponse.accept(topicPartition, new PartitionResponse(Errors.forException(e)));
             final String fullPartitionName = KopTopic.toString(topicPartition, namespacePrefix);
-
             authorize(AclOperation.WRITE, Resource.of(ResourceType.TOPIC, fullPartitionName))
                     .whenComplete((isAuthorized, ex) -> {
                         if (ex != null) {
```
Hongten, line 940: It seems those two `if` can combine together. Like 
```
if (ex != null || !isAuthorized) {
    if (ex != null) {
        log.error("Write topic authorize failed, topic - {}. {}",
                fullPartitionName, ex.getMessage());
    }
    unauthorizedTopicResponsesMap.put(topicPartition,
            new ProduceResponse.PartitionResponse(Errors.TOPIC_AUTHORIZATION_FAILED));
    completeOne.run();
    return;
}
```

Issue #924, opened 2021-11-23
[FEATURE] Adjust pom.xml for dependency management


BewareMyPower:
**Is your feature request related to a problem? Please describe.**
Currently many dependencies are in the `<dependency>` block but not the `<dependencyManagement>` block. It's not a good design because if we want to add a new module, the module will inherit these dependencies by default.

**Describe the solution you'd like**
Adjust all the `pom.xml` files. Move the dependencies from `<dependency>` to `<dependencyManagement> in the root `pom.xml`.

PR #926, opened 2021-11-24
Adjust pom.xml for dependency management

BewareMyPower:
Fixes https://github.com/streamnative/kop/issues/924

### Motivation

The current dependency management is bad. We should ensure new added module doesn't depend on some unnecessary dependencies.

### Modifications

The main changes are:
- Move all dependencies from `<dependencies>` under `<dependencyManagement>` block in root `pom.xml`.
- Add required dependencies for each module.

In addition, some minor changes were made:
1. Remove google common dependencies from modules of different Kafka version, like `kafka-0-9`. It's redundant to introduce a dependency just for a `Maps.newHashMap` method.
2. Rename `mavem-shade-plugin.version` to `maven-shade-plugin.version`.

### DIFF
--- a/kafka-0-10/pom.xml
+++ b/kafka-0-10/pom.xml
@@ -46,7 +46,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-shade-plugin</artifactId>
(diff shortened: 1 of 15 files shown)


Demogorgon314 (approved):
LGTM. Nice work!

PR #928, opened 2021-11-24
Internal brokers communication and distribution of topics deleted messages

### LINKED ISSUE #927: [BUG] Topics that are not subscribed will not be deleted from zk
**Describe the bug**

When a topic is not consumed, call KafkaAdminClient to delete this topic. This topic will create a child node under the zk path /kop//delete_topics. Since there is no subscription information, it will not be deleted from io.streamnative.pulsar. handlers.kop.KopEventManager.DeleteTopicsEvent

The following test method can verify that the zk node will always exist on zk after deleting this topic.

**To Reproduce**
This is an existing test method which from io.streamnative.pulsar.handlers.kop.KafkaRequestHandlerTest#testDeleteTopicsAndCheckChildPath
···
    @Test(timeOut = 10000)
    public void testDeleteTopicsAndCheckChildPath() throws Exception {
        Properties props = new Properties();
        props.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:" + getKafkaBrokerPort());

        @Cleanup
        AdminClient kafkaAdmin = AdminClient.create(props);
        Map<String, Integer> topicToNumPartitions = new HashMap<String, Integer>(){{
            put("testCreateTopics-0", 1);
            put("testCreateTopics-1", 3);
            put("my-tenant/my-ns/testCreateTopics-2", 1);
            put("persistent://my-tenant/my-ns/testCreateTopics-3", 5);
        }};
        // create
        createTopicsByKafkaAdmin(kafkaAdmin, topicToNumPartitions);
        verifyTopicsCreatedByPulsarAdmin(topicToNumPartitions);
        // delete
        deleteTopicsByKafkaAdmin(kafkaAdmin, topicToNumPartitions.keySet());
        verifyTopicsDeletedByPulsarAdmin(topicToNumPartitions);
        // check deleted topics path
        Set<String> deletedTopics = handler.getPulsarService()
                .getBrokerService()
                .getPulsar()
                .getLocalMetadataStore()
                .getChildren(KopEventManager.getDeleteTopicsPath())
                .join()
                .stream()
                .map((TopicNameUtils::getTopicNameWithUrlDecoded))
                .collect(Collectors.toSet());

        assertEquals(topicToNumPartitions.keySet(), deletedTopics);
    }
···

wenbingshen:
Fixes #927 

1.Message forwarding between kop brokers through a given internal communication protocol.
2.Reuse the DeleteTopicsRequest request protocol. If other request protocols are introduced, it will conflict with subsequent Kafka dependent version upgrades.
3.Because the existing protocol is reused, the message is forwarded directly through the KafkaAdminClient management tool, without additional definition of NetWorkClient

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
@@ -19,10 +19,10 @@ import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
-import com.google.gson.JsonParser;
 import io.streamnative.pulsar.handlers.kop.coordinator.group.GroupCoordinator;
 import io.streamnative.pulsar.handlers.kop.coordinator.group.GroupMetadata;
 import io.streamnative.pulsar.handlers.kop.stats.StatsLogger;
+import io.streamnative.pulsar.handlers.kop.utils.AdminClientUtils;
 import io.streamnative.pulsar.handlers.kop.utils.KopTopic;
 import io.streamnative.pulsar.handlers.kop.utils.MetadataUtils;
 import io.streamnative.pulsar.handlers.kop.utils.ShutdownableThread;
@@ -32,9 +32,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
-import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.LinkedBlockingQueue;
@@ -47,11 +45,8 @@ import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.bookkeeper.common.util.MathUtils;
 import org.apache.kafka.clients.admin.AdminClient;
-import org.apache.kafka.clients.admin.AdminClientConfig;
-import org.apache.kafka.clients.admin.KafkaAdminClient;
 import org.apache.kafka.common.Node;
 import org.apache.kafka.common.TopicPartition;
-import org.apache.kafka.common.config.SaslConfigs;
 import org.apache.pulsar.broker.loadbalance.LoadManager;
 import org.apache.pulsar.common.util.Murmur3_32Hash;
 import org.apache.pulsar.metadata.api.MetadataStore;
@@ -190,7 +185,7 @@ public class KopEventManager {
                         }
 
                         if (brokerData.isPresent()) {
-                            JsonObject jsonObject = parseJsonObject(
+                            JsonObject jsonObject = AdminClientUtils.parseJsonObject(
                                     new String(brokerData.get().getValue(), StandardCharsets.UTF_8));
                             JsonObject protocols = jsonObject.getAsJsonObject("protocols");
                             JsonElement element = protocols.get("kafka");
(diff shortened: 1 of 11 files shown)


BewareMyPower (changes requested):
I'm afraid this feature has gone too far.

The initial purpose is to delete the z-nodes created when topics without subscription were deleted. A simple solution is adding a periodical task to check these z-nodes. Because the existing z-node doesn't affect the recreation of the topic.

It looks too complex.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
```diff
@@ -337,23 +346,33 @@ public void process(BiConsumer<String, Long> registerEventLatency,
...
+                    if (!advertisedEndPoint.getHostname().equals(broker.host())
+                            || advertisedEndPoint.getPort() != broker.port()) {
+                        AdminClient adminClient =
+                                AdminClientUtils.getKafkaAdminClient(kafkaConfig, broker.host(), broker.port());
+                        adminClients.add(adminClient);
+                        adminClient.deleteTopics(topicsDeletions);
+                        log.info("Send deleteTopics {} to kop broker {}:{}",
+                                topicsDeletions, broker.host(), broker.port());
+                    }
+                });
```
BewareMyPower, line 370: What if there are many brokers? Why does each broker need to process a DELETE_TOPICS request?

BewareMyPower:
@Demogorgon314 Could you also take a look at this PR?

wenbingshen:
> I'm afraid this feature has gone too far.
> 
> The initial purpose is to delete the z-nodes created when topics without subscription were deleted. A simple solution is adding a periodical task to check these z-nodes. Because the existing z-node doesn't affect the recreation of the topic.
> 
> It looks too complex.

I am also thinking of some better ways to solve this problem. This is one of the ideas I thought of. If there is a better way, I agree to implement it.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopEventManager.java
wenbingshen: In kafka, DeleteTopicsRequest will only be sent to the controller, while in kop, the node processing the metadataRequest will return itself to become the controller node.

wenbingshen: So I want to use this feature of the client and the processing logic of kop, and the kop node that receives the client's delete topic request will forward this request to other kop nodes.

wenbingshen: Therefore, creating an adminClient for each kop broker can ensure that the corresponding kop broker becomes the controller of this request.

wenbingshen: The request to update metadata in Kafka is also distributed by the Controller node to all Kafka brokers.

BewareMyPower: Kafka controller is used to manage metadata of the whole cluster, while each Pulsar broker exposes a HTTP endpoint for metadata management based on ZK (metadata store). Pulsar doesn't need to maintain leader and ISRs so the metadata is not as much as Kafka has.

Considering a case that you found sending the request to all brokers brings a performance overhead, maybe you would like to add a controller role in KoP?

Before going so far, we should avoid broker-to-broker communication from the beginning. If we want to share some metadata among the brokers, use the metadata store or create a system topic.

wenbingshen: I don't want to add the controller role to kop because I also noticed that pulsar also uses PulsarAdmin to forward messages. However, you are right. It is best to use metadata store or create system topic. As you can see, the previous method is based on metadata store.

BewareMyPower:
Close this PR first.

PR #929, opened 2021-11-24
Improve TransactionMarkerChannelManager

eolivelli:
This patch adds failure handling on TransactionMarkerChannelHandler.
In case of failure on the socket we must fail every pending request, otherwise the system will be stuck

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionMarkerChannelManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionMarkerChannelManager.java
@@ -33,11 +33,13 @@ import java.util.Set;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiConsumer;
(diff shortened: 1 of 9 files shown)


Demogorgon314:
Please ping me if this PR is ready to review : ) 

Nice catch BTW.

eolivelli:
@Demogorgon314 all tests passed.
I believe that this is a good first cleanup.

I figured out that most of the problems I saw happen when you have two brokers and the load balancer starts to unload topics and move them to another broker or when the bootstrap of one of the brokers is slower.
This is way I cannot reproduce my failures in simple unit tests.

I have not added tests yet, probably we should add tests with more then one broker and simulate failures, but it will take much time to code them.

I am continuing to run my system tests and see if this patch is enough to reach perfect stability (100% tests passing for a good number of times)

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionMarkerChannelManager.java
```diff
@@ -235,13 +258,20 @@ public void addTxnMarkersToBrokerQueue(String transactionalId,
             CompletableFuture<Void> addFuture = new CompletableFuture<>();
             addressFutureList.add(addFuture);
             addressFuture.whenComplete((address, throwable) -> {
-                if (throwable != null) {
+                if (throwable != null)  {
```
Demogorgon314, line 261:
```suggestion
                if (throwable != null) {
```

Demogorgon314 (approved):
Overall LGTM. Maybe we could add the units test in the next PR.

eolivelli:
@Demogorgon314 I have committed your suggestion

PR #931, opened 2021-11-25
Avoid waiting a future when channel is closed

BewareMyPower:
### Motivation

`KafkaRequestChannel#close` calls `KafkaTopicManager#removePersistentTopicAndReferenceProducer` for each topic in cache. But `CompletableFuture#get` is called in `removePersistentTopicAndReferenceProducer`, which is called in an I/O thread. It will lead to a deadlock issue.

image

### Modifications

Instead of calling `get` for a `PersistentTopic` object, call `thenAccept` and `exceptionally` method on the future to do the cleanup work for `PersistentTopic`.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
@@ -312,6 +312,9 @@ public class KafkaTopicManager {
                             topicName, (producer.getTopic() == null) ? "null" : producer.getTopic().getName());
                 }
             }
+        }).exceptionally(e -> {
+            log.error("Failed to get topic '{}' in removeTopicAndReferenceProducer", topicName, e);
+            return null;
         });
     }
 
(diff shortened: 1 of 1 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicManager.java
```diff
@@ -316,9 +312,7 @@ private static void removePersistentTopicAndReferenceProducer(final String topic
                             topicName, (producer.getTopic() == null) ? "null" : producer.getTopic().getName());
                 }
             }
-        } catch (InterruptedException | ExecutionException e) {
-            log.error("Failed to get topic '{}' in removeTopicAndReferenceProducer", topicName, e);
-        }
+        });
```
Demogorgon314, line 315: Do we need to handle the exceptionally? I see topicFuture might complete exceptionally.

BewareMyPower: Addressed.

Demogorgon314 (approved):

PR #937, opened 2021-11-29
fix quick start link error

yzydtc:
when i read quick start,i found some problems about this links,it would be 404 bot found,so i fixed this links to right address.


BewareMyPower:
These links are not purposed to be read from Github. They are converted to the actual links in by replacing `{{protocol:version}}` with the actual version in a release. See [Release 2.8.1.24](https://github.com/streamnative/kop/blob/branch-2.8.1.24/docs/kop.md#how-to-use-kop) for example. We did this because the docs will be grabbed by [StreamNative hub](https://hub.streamnative.io/protocol-handlers/kop/v2.8.1.24/#how-to-use-kop).

PR #940, opened 2021-11-29
Fix problem with Transactions and Multibroker cluster

eolivelli:
When we handle WriteTxMarkers and we use short topic name there is a confusion while applying the default namespace.
This happens only during the Lookup operation that we need to find the Broker to connect to in order to send the TxMarkers.

In this case the Markers must be sent to the topic in the "user" space (usually public/default), not in the metadata namespace (usually public/__kafka).

In this patch we explicitly give a name to the prefix for metadata and for user topics and this confusion is solved.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java
@@ -69,7 +69,8 @@ import org.apache.pulsar.metadata.api.extended.MetadataStoreExtended;
 @Slf4j
 public class TransactionCoordinator {
 
-    private final String namespacePrefix;
+    private final String namespacePrefixForMetadata;
+    private final String namespacePrefixForUserTopics;
     private final TransactionConfig transactionConfig;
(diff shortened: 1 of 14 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java
```diff
@@ -133,18 +136,20 @@ public static TransactionCoordinator of(TransactionConfig transactionConfig,
...
-                                            String namespacePrefix) {
+                                            String namespacePrefixForMetadata,
+                                            String namespacePrefixForUserTopics) {
         TransactionStateManager transactionStateManager =
                 new TransactionStateManager(transactionConfig, txnTopicClient, scheduler, time);
         return new TransactionCoordinator(
                 transactionConfig,
                 new TransactionMarkerChannelManager(null, transactionStateManager,
-                        kopBrokerLookupManager, false, namespacePrefix),
+                        kopBrokerLookupManager, false, namespacePrefixForUserTopics),
```
eolivelli, line 146: This is actually the fix for the problem

Demogorgon314 (approved):
LGTM.

PR #941, opened 2021-11-30
Transactions: implement support for JWT token authentication between brokers

eolivelli:
**Motivation**
When you enable JWT token authentication transactions do not work because transactions need broker-to-broker communication but there is currently no support for the SASL handshake.

**Changes**
Now when you enable authentication and you configure Token Authentication for Broker-To-Broker communications in Pulsar, we perform SASL PLAIN authentication while preparing the connection to other brokers.

**Tests**
I have extended existing SASL tests in order to use Transactions. I have copy/pasted some parts of the TransactionTest, but there was not enough code to make sense to share the test code.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
@@ -567,7 +567,7 @@ public abstract class KafkaCommandDecoder extends ChannelInboundHandlerAdapter {
     protected abstract void
     handleCreatePartitions(KafkaHeaderAndRequest kafkaHeaderAndRequest, CompletableFuture<AbstractResponse> response);
 
-    static class KafkaHeaderAndRequest implements Closeable {
+    public static class KafkaHeaderAndRequest implements Closeable {
 
         private static final String DEFAULT_CLIENT_HOST = "";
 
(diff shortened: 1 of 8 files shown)


eolivelli:
@Demogorgon314 @BewareMyPower 
I have ported part of the KOP Proxy code (copy/paste) regarding broker-to-broker authentication.
One day if we accept the KOP Proxy here in the main repository we can share a single class for Proxy-To-Broker and for Broker-To-Broker communications.

https://github.com/datastax/starlight-for-kafka/blob/41b59fc67885db66846d945606db44062ff12f79/proxy/src/main/java/io/streamnative/pulsar/handlers/kop/ConnectionToBroker.java#L49

eolivelli:
@Demogorgon314 
unfortunately the new test fails here in KOP repo

in my fork the test is passing.

This patch #849 fixes a problem with Multi-Tenant configuration and SASL, but it is not enough to make the test pass.

Today I will update #849 and address your comments, so that our codebase will be more one close to each other

eolivelli:
it looks like #849 is not needed.

in my fork I am missing the latest refactor about ReplicaManager, probably that's the cause.

can you help me in debugging the failure ?
@Demogorgon314

Demogorgon314:
I found the reason why the test failed.

First of all, you need rebase master onto the current branch, then set `conf.setKafkaTransactionCoordinatorEnabled(true);` in `SaslPlainTestBase` since the `TransactionCoordinator` default is disabled.

Demogorgon314 (approved):
LGTM. Nice work!

PR #943, opened 2021-12-01
Add Kafka payload processor 

BewareMyPower:
### Motivation

It's recommended to configure `entryFormat=kafka` for better performance among Kafka clients in KoP. However, since it writes the original messages of Kafka format to BK, Pulsar consumer cannot consume these messages.

[PIP-96](https://github.com/apache/pulsar/wiki/PIP-96%3A-Message-payload-processor-for-Pulsar-client), which will be included in Pulsar 2.9.0, introduced a payload processor for Pulsar consumer. Based on PIP-96, this PR intends to add a payload processor for messages of Kafka format.

### Modifications

- Add a `kafka-payload-processor` module that contains the implementation of `KafkaPayloadProcessor`.
- Add tests to cover null values, keys and headers.

### DIFF
--- a/kafka-payload-processor/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaPayloadProcessor.java
+++ b/kafka-payload-processor/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaPayloadProcessor.java
@@ -16,7 +16,9 @@ package io.streamnative.pulsar.handlers.kop;
 import io.netty.buffer.ByteBuf;
 import io.netty.util.concurrent.FastThreadLocal;
 import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 import java.util.function.Consumer;
+import org.apache.kafka.common.header.Header;
 import org.apache.kafka.common.record.MemoryRecords;
 import org.apache.kafka.common.record.Record;
 import org.apache.kafka.common.serialization.StringDeserializer;
@@ -105,6 +107,12 @@ public class KafkaPayloadProcessor implements MessagePayloadProcessor {
             valueBuffer = null;
             singleMessageMetadata.setNullValue(true);
             singleMessageMetadata.setPayloadSize(0);
+        }
+
+        for (Header header : record.headers()) {
+            singleMessageMetadata.addProperty()
+                    .setKey(header.key())
+                    .setValue(new String(header.value(), StandardCharsets.UTF_8));
         }
 
         final ByteBuf buf = PulsarByteBufAllocator.DEFAULT.buffer(
--- a/kafka-payload-processor/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaPayloadProcessor.java
+++ b/kafka-payload-processor/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaPayloadProcessor.java
@@ -21,6 +21,7 @@ import java.util.function.Consumer;
 import org.apache.kafka.common.header.Header;
 import org.apache.kafka.common.record.MemoryRecords;
 import org.apache.kafka.common.record.Record;
+import org.apache.kafka.common.record.RecordBatch;
 import org.apache.kafka.common.serialization.StringDeserializer;
 import org.apache.pulsar.client.api.Message;
 import org.apache.pulsar.client.api.MessagePayload;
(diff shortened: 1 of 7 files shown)


BewareMyPower:
There are something to be done in following PRs:
1. Document how Pulsar users should use `KafkaPayloadProcessor`.
2. More tests should be added to `io.streamnative.pulsar.handlers.kop.compatibility` package so that more versions of Kafka client can be verified.

kafka-payload-processor/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaPayloadProcessor.java
```diff
@@ -0,0 +1,127 @@
...
+        if (!isKafkaFormat(context)) {
+            DEFAULT.process(payload, context, schema, messageConsumer);
+            return;
+        }
+
+        final ByteBuf buf = MessagePayloadUtils.convertToByteBuf(payload);
+        try {
+            final MemoryRecords records = MemoryRecords.readableRecords(buf.nioBuffer());
+            int numMessages = 0;
+            for (Record ignored : records.records()) {
```
Demogorgon314, line 63: We might need to check this batch is a control batch or not, because users might use transactions in KoP, and some markers will write in the normal topics, Pulser consumers don't want to consume these messages.
```java
for (MutableRecordBatch recordBatch : records.batches()) {
    boolean controlBatch = recordBatch.isControlBatch();
}
```

BewareMyPower: It's right, I'll fix it soon.

BewareMyPower: I added the check but no tests are added. I also noticed `AbstractEntryFormatter#decode` also doesn't handle the case. Is it possible to cover the case in tests currently?

BewareMyPower: PTAL again, @Demogorgon314

kafka-payload-processor/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaPayloadProcessor.java
```diff
@@ -0,0 +1,136 @@
...
+                numMessages++;
+            }
+            int index = 0;
+            for (RecordBatch batch : records.batches()) {
+                if (batch.isControlBatch() || batch.isTransactional()) {
```
Demogorgon314, line 69: The normal batch if is sent by transaction, then the `batch.isTransactional` is always true.  We only need to skip when the batch is the control batch.

BewareMyPower: Good point, addressed.

Demogorgon314:
Can we add a test to send the message by Kafka transaction and when producing over we use Pulser consumer to consume the message?

BewareMyPower:
@Demogorgon314 There is a thing that I have asked before.

>  I also noticed AbstractEntryFormatter#decode also doesn't handle the case. Is it possible to cover the case in tests currently?

I tried to add the Pulsar consumer to `TransactionTest`, but when I debugged the code, no control batch record is found.

```java
    @Test
    public void test() throws Exception {
        final String topic = "test";
        final Properties props = newKafkaProducerProperties();
        props.put(ProducerConfig.TRANSACTIONAL_ID_CONFIG, "txn-10");
        final KafkaProducer<String, String> producer = new KafkaProducer<>(props);
        producer.initTransactions();
        producer.beginTransaction();
        producer.send(new ProducerRecord<>(topic, "hello")).get();
        producer.commitTransaction();
        producer.close();

        final Consumer<byte[]> consumer = pulsarClient.newConsumer()
                .topic(topic)
                .subscriptionName("my-sub")
                .subscriptionInitialPosition(SubscriptionInitialPosition.Earliest)
                .messagePayloadProcessor(new KafkaPayloadProcessor())
                .subscribe();
        final Message<byte[]> message = consumer.receive();
        System.out.println(new String(message.getValue()));
        consumer.close();
    }
```

So the question is, how to send the control batch?

Demogorgon314:
Oh, sorry, I forgot before. The current implementation of the control batch is “incorrect”, we might need to merge the  #912, before adding this test.

The reason why you can't see the control batch is current implementation is using pulsar format as the control batch.

Demogorgon314:
Maybe you can manually build a control batch and use the pulsar producer save it to the topic just like this:

```java
EndTransactionMarker marker = new EndTransactionMarker(ControlRecordType.COMMIT, 0);
MemoryRecords memoryRecords = MemoryRecords.withEndTransactionMarker(0, (short) 0, marker);
```

Demogorgon314 (approved):
LGTM. Nice work! I approve first, you can choose to add the test in this PR, or the next PR.

BewareMyPower:
> Maybe you can manually build a control batch and use the pulsar producer save it to the topic

I think it's better to add a real world test for the control record. If we cannot use Kafka producer to send a control batch currently, the test will be meaningless. Because we can use a Pulsar producer to send any message.

Therefore, the test should be added after #912 is merged.

PR #947, opened 2021-12-03
Implement ability to not handle namespaces automatically: kafkaManageSystemNamespaces

eolivelli:
**Motivation**

When you have a system with many independent tenants having a global configuration that is the same for every tenant is problematic.
Therefore there are cases in which you don't want that the PH creates namespaces or overrides the configuration, because there is some other component (like a separate control plane) that is in charge of managing all the system configurations for each tenant.

**Changes**
Add a new configuration option `kafkaManageSystemNamespaces` to disable any operation about creating namespaces or setting any configuration.
This option will allow to move the management of such things to an external entity.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MetadataUtils.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MetadataUtils.java
@@ -98,6 +98,10 @@ public class MetadataUtils {
                                                      KopTopic kopTopic,
                                                      int partitionNum)
         throws PulsarAdminException {
+        if (!conf.isKafkaManageSystemNamespaces()) {
+            log.info("Skipping initialization of topic {} for tenant {}", kopTopic.getFullName(), tenant);
+            return;
+        }
         String cluster = conf.getClusterName();
         String kafkaMetadataNamespace = tenant + "/" + conf.getKafkaMetadataNamespace();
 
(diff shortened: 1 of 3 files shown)


Demogorgon314:
Can you please add more descriptions?

Demogorgon314:
Ping @eolivelli

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MetadataUtils.java
```diff
@@ -98,6 +98,10 @@ private static void createKafkaMetadataIfMissing(String tenant,
                                                      KopTopic kopTopic,
                                                      int partitionNum)
         throws PulsarAdminException {
+        if (!conf.isKafkaManageSystemNamespaces()) {
```
BewareMyPower, line 101: The config name is better to be `skipCreatingKafkaMetadata` (and the default value is false) according to the method name.

eolivelli: Initially I thought about it but
it is not only about "creating" but it is also about setting retention and other parameters

we don't want that parameters set by the user are overridden by the PH.

think about an environment where there are tons of independent tenants, and you want total independence of the configurations (and also you want that some system control plane is in change of managing system namespaces)

Demogorgon314: Can we call the config name `kopAutoManageMetadataEnable`(the default value is true)? Do you have a better idea?

BewareMyPower: Ping @eolivelli. Is there any update?  I think it's only a matter of config name.

BewareMyPower (approved):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MetadataUtils.java
Demogorgon314: Make sense.

PR #948, opened 2021-12-05
Refactor append records analyze and validate records

Demogorgon314:
### Motivation
Currently, when appending records we should analyze and validate the append records, but some operates can be optimized, such as :
1. Shouldn't check the max message size in another loop
2. Shouldn't calculate num of messages in another loop
3. Shouldn't get the source codec in another loop

### Modifications
1. Check the max message size in the `analyzeAndValidateRecords` method.
2. Calculate num of messages in the `analyzeAndValidateRecords` method.
3. Get the source codec in the `analyzeAndValidateRecords` method.
4. Test of analyze and validate records.


BewareMyPower:
I found #912 did much work for the similar thing. Does this PR conflict with #912?

Demogorgon314:
> I found #912 did much work for the similar thing. Does this PR conflict with #912?

Since PR #912 already moved `LogAppendInfo` to the outside of class, this PR needs to find a way to access the `LogAppendInfo`, when we need to test it.

BewareMyPower:
> this PR needs to find a way to access the LogAppendInfo, when we need to test it.

We can access the class if the test is under the same package (`io.streamnative.pulsar.handlers.kop.storage`).

kafka-impl/src/test/java/io/streamnative/pulsar/handlers/kop/format/EntryFormatterTest.java
```diff
@@ -104,22 +120,25 @@ public void testEntryFormatterEncode(CompressionType compressionType, byte magic
 
         MemoryRecords records = prepareRecords(compressionType, magic);
 
+        PartitionLog.LogAppendInfo appendInfo =
```
Demogorgon314, line 123: @BewareMyPower See this class.

BewareMyPower: I see. That's because you pass the `LogAppendInfo` to the constructor of `EncodeRequest`.

Generally, a class with default access modifier should only be used in its package. But since you want classes in other packages to access this class (`LogAppendInfo`), then it should be exposed as public class.

BewareMyPower: So in this PR, making `LogAppendInfo` a static inner class is okay.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLogTest.java
```diff
@@ -0,0 +1,146 @@
...
+    @Test(dataProvider = "compressionTypes")
+    public void testAnalyzeAndValidateRecords(CompressionType compressionType) {
+        MemoryRecords memoryRecords = buildMemoryRecords(new int[]{3, 2, 5}, compressionType, 100);
+        PartitionLog.LogAppendInfo appendInfo =
+                PARTITION_LOG.analyzeAndValidateRecords(memoryRecords);
+        Assert.assertEquals(appendInfo.shallowCount(), 3);
+        Assert.assertEquals(appendInfo.numMessages(), 10);
+        Assert.assertEquals(appendInfo.validBytes(), memoryRecords.validBytes());
+        Assert.assertEquals(appendInfo.sourceCodec().codec(), compressionType.id);
+        Assert.assertFalse(appendInfo.isTransaction());
```
BewareMyPower, line 65: I found all tests only validate the following fields:

```java
        private int numMessages;
        private int shallowCount;
        private boolean isTransaction;
        private int validBytes;
```

Only this test validates the `sourceCodec` field. So I think we can add a method to validate the other fields of `appendInfo` are the default values.

Demogorgon314: Addressed.

BewareMyPower: The `lastOffsetOfFirstBatch` field is still tested, PTAL.

Demogorgon314: Addressed. PTAL.

BewareMyPower (approved):

Issue #768, opened 2021-09-30
[BUG] KoP and pulsar-2.8.0 compatibility


wenbingshen:
**Describe the bug**

I use pulsar-2.8.0 and kop's current master branch to test, kafka consumers cannot consume data, kop does not throw any exceptions, and the Fetch request will never be completed until timeout, then will be cancelled; after @BewareMyPower prompt, I tested kop-2.8.0.16 and pulsar-2.8.0 also have this problem, which is a compatibility problem introduced by kop.
We should remind users about compatibility issues in this area, or we should branch out a patch for the latest kop-2.8.0.16 to solve this compatibility issue.

BewareMyPower:
You can push a fix to branch-2.8. But since latest Pulsar 2.8.x version is 2.8.1 now, we might not release any KoP 2.8.0.x. So it's better to add docs to mark the affected minor versions.

wenbingshen:
> You can push a fix to branch-2.8. But since latest Pulsar 2.8.x version is 2.8.1 now, we might not release any KoP 2.8.0.x in future. So it's better to add docs to mark the affected minor versions.

I agree to add docs to mark the affected minor versions.

PR #951, opened 2021-12-06
Fix NPE problems caused by mockZooKeeper API changes when dump version to 2.8.1.26

Demogorgon314:
Fix https://github.com/streamnative/kop/runs/4425667410?check_suite_focus=true#step:5:283

#950

### Motivation

MockZookeeper API has changed, the `CreateMode` shouldn't be null. We should use MetadataStore instead of mockZooKeeper since KoP already remove all zkClient usage.

### Modifications

Use MetadataStore instead of mockZooKeeper in ProducerIdManagerTest

### DIFF
--- a/tests/src/test/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/ProducerIdManagerTest.java
+++ b/tests/src/test/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/ProducerIdManagerTest.java
@@ -17,8 +17,9 @@ import static org.testng.AssertJUnit.assertEquals;
 import static org.testng.AssertJUnit.fail;
 
 import io.streamnative.pulsar.handlers.kop.KopProtocolHandlerTestBase;
+import java.util.Optional;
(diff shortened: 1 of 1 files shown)


BewareMyPower:
Could you upgrade the Pulsar dependency to 2.8.1.26 as well? (Current dependency is 2.8.0.15) This change was introduced from https://github.com/apache/pulsar/pull/13066, we should also note it in PR description.

BewareMyPower:
It's a breaking change but it only affects the tests. So I think upgrading Pulsar to 2.8.1.26 is OK and won't cause the compatibility issue like https://github.com/streamnative/kop/issues/768. The version upgrade is just to ensure the fix works.

BewareMyPower:
Sorry I missed the fact this is a PR to master branch. We can do the dependency upgrade later. I'll merge this PR first.

BewareMyPower (approved):

BewareMyPower:
Remove the 2.8.1 related labels because this PR uses an API only for 2.9.0 or later.

PR #952, opened 2021-12-06
Bump version to 2.8.1.26

BewareMyPower:
https://github.com/apache/pulsar/pull/13066 introduced a breaking change to `MockZooKeeper` API and causes test failure. https://github.com/streamnative/kop/pull/951 fixed this bug and this PR is to verify it.

Since the change only affects tests, this PR won't cause the breaking change.


BewareMyPower:
There is a problem with #951, I'll rebase to master once the fix has been merged to branch-2.8.1.

Issue #958, opened 2021-12-09
[BUG] KafkaPayloadProcessor cannot be used with Pulsar shaded dependencies


BewareMyPower:
**Describe the bug**

`pulsar-client` and `pulsar-client-all` both use Maven shaded plugin to rename some classes. However, KoP's `kafka-payload-processor` package doesn't relocate the package name.

**To Reproduce**

Run KoP standalone with `entryFormat=kafka`. Then send some messages via Kafka CLI.

Create a Maven project with following dependencies.

```xml
    <dependency>
      <groupId>io.streamnative</groupId>
      <artifactId>pulsar-client</artifactId>
      <version>2.9.0-rc-202110221101</version>
    </dependency>

    <dependency>
      <groupId>io.streamnative.pulsar.handlers</groupId>
      <artifactId>kafka-payload-processor</artifactId>
      <version>2.9.0-SNAPSHOT</version>
    </dependency>
```

Here is the example code snippet.

```java
            final Consumer<byte[]> consumer = client.newConsumer()
                    .topic("my-topic")
                    .subscriptionName("my-sub-2")
                    .subscriptionInitialPosition(SubscriptionInitialPosition.Earliest)
                    .messagePayloadProcessor(new KafkaPayloadProcessor())
                    .subscribe();
            while (true) {
                Message<byte[]> message = consumer.receive(1, TimeUnit.SECONDS);
                if (message != null) {
                    System.out.println(new String(message.getValue(), StandardCharsets.UTF_8));
                    break;
                }
            }
```

When we ran the program, the following error happened.

```
java.lang.NoClassDefFoundError: io/netty/util/concurrent/FastThreadLocal
    at Main.main (Main.java:22)
    at org.codehaus.mojo.exec.ExecJavaMojo$1.run (ExecJavaMojo.java:254)
    at java.lang.Thread.run (Thread.java:748)
Caused by: java.lang.ClassNotFoundException: io.netty.util.concurrent.FastThreadLocal
    at java.net.URLClassLoader.findClass (URLClassLoader.java:382)
    at java.lang.ClassLoader.loadClass (ClassLoader.java:418)
    at java.lang.ClassLoader.loadClass (ClassLoader.java:351)
    at Main.main (Main.java:22)
    at org.codehaus.mojo.exec.ExecJavaMojo$1.run (ExecJavaMojo.java:254)
    at java.lang.Thread.run (Thread.java:748)
```

Even after we added the Netty dependency.

```xml
    <dependency>
[...]

PR #959, opened 2021-12-09
Add kafka-payload-processor shaded module

BewareMyPower:
Fixes https://github.com/streamnative/kop/issues/958

### Motivation

Current `kafka-payload-processor` dependency is not compatible with `pulsar-client` dependency.

### Modifications

- Change the original `kafka-payload-processor` artifact id to `kafka-payload-processor-original`, which is compatible with `pulsar-client-original` dependency.
- Add a new module `kafka-payload-processor-shaded` whose artifact id is `kafka-payload-processor`. This module is compatible with `pulsar-client` dependency.
- Add a new module `kafka-payload-processor-shaded-tests` that contains a basic test for artifact `kafka-payload-processor` with `pulsar-client` dependency.


BewareMyPower:
Now this PR is ready to review, PTAL @Demogorgon314

kafka-payload-processor-shaded-tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaPayloadProcessorTest.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.List;
+import org.apache.pulsar.client.api.Schema;
+import org.apache.pulsar.client.impl.MessageIdImpl;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class KafkaPayloadProcessorTest {
+
+    @Test
+    public void testValues() throws Exception {
```
Demogorgon314, line 27: Can we add some tests to verify bypassing the control batch?

BewareMyPower: This PR only tests the `kafka-payload-processor` dependency could work with `pulsar-client` dependency. So this test is more like a smoke test. I think we can use other PRs to add more tests, including tests for keys and headers.

Demogorgon314 (approved):
LGTM.

PR #964, opened 2021-12-12
[BUGFIX] Fix the internal topics should be controlled in the meta namespace

wenbingshen:
`public/default/__consumer_offsets` is incorrectly considered as an internal metadata topic
image

1. Based on #793 , we can control whether users are allowed to access the `metadata` namespace

2. I am still not sure whether we should allow users to access the kafka topics which under other namespaces (such as `default` ) have the same name with the internal topics. The current implementation of the kafka topics under the `default` namespace have the same name with the internal topics will be used as the internal topics and returned to the kafka client.

3. However, we should bind to the `metadata` namespace when judging the `internal` topics.

This PR solves problem 3. When problem 3 is solved, we can treat the kafka topics which under other namespaces have the same name with the internal topics  as user topics.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/KopTopic.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/KopTopic.java
@@ -17,6 +17,7 @@ import static org.apache.kafka.common.internals.Topic.GROUP_METADATA_TOPIC_NAME;
 import static org.apache.kafka.common.internals.Topic.TRANSACTION_STATE_TOPIC_NAME;
 import static org.apache.pulsar.common.naming.TopicName.PARTITIONED_TOPIC_SUFFIX;
 
+import io.streamnative.pulsar.handlers.kop.KafkaServiceConfiguration;
 import io.streamnative.pulsar.handlers.kop.exceptions.KoPTopicException;
 import lombok.Getter;
 import org.apache.kafka.common.TopicPartition;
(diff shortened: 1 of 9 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/KopTopic.java
```diff
@@ -106,15 +107,20 @@ public static String toString(String topic, int partition, String namespacePrefi
...
+    public static boolean isGroupMetadataTopicName(final String fullTopicName, KafkaServiceConfiguration conf) {
         String partitionedTopicName = TopicName.get(fullTopicName).getPartitionedTopicName();
-        return partitionedTopicName.endsWith("/" + GROUP_METADATA_TOPIC_NAME);
+        return partitionedTopicName.endsWith(conf.getKafkaMetadataNamespace() + "/" + GROUP_METADATA_TOPIC_NAME);
+    }
+
+    public static boolean isTransactionMetadataTopicName(final String fullTopicName, KafkaServiceConfiguration conf) {
+        String partitionedTopicName = TopicName.get(fullTopicName).getPartitionedTopicName();
+        return partitionedTopicName.endsWith(conf.getKafkaMetadataNamespace() + "/" + TRANSACTION_STATE_TOPIC_NAME);
     }
```
BewareMyPower, line 124: Could you avoid passing the whole `KafkaServiceConfiguration` as the argument? I think here we only need a namespace name.

Since we've already used `TopicName` here, we can did the following check:

```java
    private static boolean isInternalTopic(final String fullTopicName,
                                           final String metadataNamespace,
                                           final Function<String, Boolean> condition) {
        final TopicName topicName = TopicName.get(fullTopicName);
        if (!topicName.getNamespace().equals(metadataNamespace)) {
            return false;
        }

        final String localName = topicName.getLocalName();
        return condition.apply(topicName.isPartitioned()
                ? localName.substring(0, localName.lastIndexOf(TopicName.PARTITIONED_TOPIC_SUFFIX))
                : localName);
    }

    public static boolean isInternalTopic(final String fullTopicName, final String metadataNamespace) {
        return isInternalTopic(fullTopicName, metadataNamespace,
                topic -> topic.equals(GROUP_METADATA_TOPIC_NAME) || topic.equals(TRANSACTION_STATE_TOPIC_NAME));
    }

    public static boolean isGroupMetadataTopicName(final String fullTopicName, final String metadataNamespace) {
        return isInternalTopic(fullTopicName, metadataNamespace, topic -> topic.equals(GROUP_METADATA_TOPIC_NAME));
    }
```

BTW, `isTransactionMetadataTopicName` is unused.

wenbingshen: I have addressed your comments,PTAL. Thanks.

wenbingshen:
`TopicName.getNamespace()` will return like `public/__kafka`, so replace with `TopicName.getNamespacePortion()` will return `__kafka`

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/KopTopic.java
wenbingshen: That's nothing. :)

BewareMyPower (approved):

PR #973, opened 2021-12-16
Fetch: trigger pending fetches when producing messages.

eolivelli:
When the Kafka client issues a Fetch and sets a maxWait time we are already scheduling a DelayedFetch, but there is no way to trigger that Fetch and it is deemed to wait for the fully timeout.
This adds latency spikes on the Kafka Consumer.

With this patch we are triggering any pending DelayedFetch in case of writing any record to one of the partitions interested by the Fetch.

This is only a first implementation, in the future we can make it better and do not trigger at the first record, but wait in any case for more records to come.
With this implementation the Fetch result will contain usually only 1 record, but this is enough to let the Kafka Client start a new Fetch cycle and do not waste time in doing nothing (waiting for maxWait).

Changes:
- trigger pending Fetches while producing to the topic
- add new metric WAITING_FETCHES_TRIGGERED
- add DelayedOperation#wakeup, that means the operation should wake up due to some trigger  (in this case the Production of records to the topic)
- add a new test that would fail without this patch (because the tests asserts that there is no idle cycle in the Consumer loop)

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -986,7 +986,7 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
     }
 
     private void notifyPendingFetches(TopicPartition topicPartition) {
-       ctx.executor().execute( () -> {
+       ctx.executor().execute(() -> {
            DelayedOperationKey.TopicPartitionOperationKey key =
                    new DelayedOperationKey.TopicPartitionOperationKey(topicPartition);
            int matches = fetchPurgatory.checkAndComplete(key);
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -997,7 +997,7 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
                }
            }
        });
-    };
+    }
 
     private void validateRecords(short version, MemoryRecords records) {
         if (version >= 3) {
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -56,6 +56,7 @@ import io.streamnative.pulsar.handlers.kop.utils.MetadataUtils;
 import io.streamnative.pulsar.handlers.kop.utils.OffsetFinder;
 import io.streamnative.pulsar.handlers.kop.utils.TopicNameUtils;
 import io.streamnative.pulsar.handlers.kop.utils.delayed.DelayedOperation;
+import io.streamnative.pulsar.handlers.kop.utils.delayed.DelayedOperationKey;
 import io.streamnative.pulsar.handlers.kop.utils.delayed.DelayedOperationPurgatory;
 import java.net.InetSocketAddress;
 import java.nio.ByteBuffer;
@@ -941,6 +942,11 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
                     mergedResponse.putAll(unauthorizedTopicResponsesMap);
                     mergedResponse.putAll(invalidRequestResponses);
                     resultFuture.complete(new ProduceResponse(mergedResponse));
+                    mergedResponse.forEach((_topicPartition, _response) -> {
+                        if (_response.error == Errors.NONE) {
+                            notifyPendingFetches(_topicPartition);
+                        }
+                    });
                 });
             }
         };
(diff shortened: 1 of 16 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/RequestStats.java
```diff
@@ -112,6 +113,12 @@
     )
     private final OpStatsLogger fetchDecodeStats;
 
+    @StatsDoc(
+            name = WAITING_FETCHES_TRIGGERED,
+            help = "number of pending fetches that woke up due to some data produced"
+    )
+    private final Counter waitingFetchesTriggered;
```
BewareMyPower, line 120: Could you add this new metric to docs/reference-metrics.md?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -979,6 +985,20 @@ protected void handleProduceRequest(KafkaHeaderAndRequest produceHar,
...
+                   new DelayedOperationKey.TopicPartitionOperationKey(topicPartition);
+           int matches = fetchPurgatory.checkAndComplete(key);
+           if (matches > 0) {
+               requestStats.getWaitingFetchesTriggered().add(matches);
+               if (log.isDebugEnabled()) {
+                   log.debug("{} DelayedFetch woke up for {}", matches, topicPartition);
+               }
+           }
+       });
+    };
```
BewareMyPower, line 1000:
```suggestion
    }
```

BewareMyPower (commented):
I think adding a parameter to `tryComplete` brings too many changes. Most of them only accept a `false` argument.

I think it's better to add a `wakeup()` method to `DelayedOperation`.

```java
    public boolean wakeup() {
        // No ops
        return true;
    }
```

Then in `DelayedFetch`, override this method.

```java
    @Override
    public boolean wakeup() {
        // if we are here then we were waiting for the condition
        // someone wrote some messages to one of the topics
        // trigger the Fetch from scratch
        restarted.set(true);
        messageFetchContext.onDataWrittenToSomePartition();
        return true;
    }
```

Finally, modify `DelayedOperationPurgatory#Watchers#tryCompleted`:

```java
                } else if (curr.wakeup() && curr.maybeTryComplete()) {
```

BTW, I did a trick that `wakeup()` return a boolean so that `wakeup()` and `maybeTryComplete()` calls can be combined in a line. The `wakeup()` method always return true.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/RequestStats.java
eolivelli: done

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
eolivelli: done

eolivelli (commented):
@BewareMyPower I have addressed your comments.

Nice suggestions !

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/delayed/DelayedOperationPurgatory.java
```diff
@@ -340,7 +340,7 @@ public void watch(T t) {
         }
 
         // traverse the list and try to complete some watched elements
-        public int tryCompleteWatched() {
+        public int tryCompleteWatched(Object key) {
```
BewareMyPower, line 343: Could you explain why did you add an argument here? Before this PR, the `key` is a field of `Watchers` and is used in following code

```java
            if (operations.isEmpty()) {
                removeKeyIfEmpty(key, this);
            }
```

However, after this change, the `key` no longer represents the field of `Watchers`. Instead, it's passed by `DelayedOperationPurgatory#checkAndComplete`. Then the behavior changes. Could you explain the difference of these two `key`s?

BewareMyPower:
I've left some comments, PTAL. BTW, please update the PR description because the current design changes a bit from the initial design.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessageFetchContext.java
```diff
@@ -195,15 +199,48 @@ private void addErrorPartitionResponse(TopicPartition topicPartition, Errors err
...
+    /**
+     * Restart this Fetch, we were waiting for some data (minBytes)
+     * and someone wrote something on any of the watched partitions.
+     */
+    public void onDataWrittenToSomePartition() {
+        decodeResults.forEach(DecodeResult::recycle);
+        decodeResults.clear();
+        bytesRead.set(0);
+        hasComplete.set(false);
+        log.info("onDataWrittenToSomePartition current resposnses {}", responseData);
```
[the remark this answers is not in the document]
eolivelli: oh, sorry, I left it for debugging in this PR.
I will remove it

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/delayed/DelayedOperationPurgatory.java
eolivelli: I added this in order to have a way in the future to pass to wakeup the motivation of the trigger.
this patch currently triggers a new Fetch execution for every partition that is involved, but in the future we can make it better and re-read the data only from the partition that received a message.
I didn't want to complicate this patch.
I can remove this parameter here and I will add it back in a follow up patch (that I still haven't implemented, because it sounds premature to me, at least for my tests and use cases)

eolivelli:
@BewareMyPower description updated.
if you want I can drop the "key" parameter as it is currently useless

eolivelli:
@BewareMyPower  I have removed that parameter
PTAL

eolivelli:
@BewareMyPower @Demogorgon314 CI passed

BewareMyPower (approved):

eolivelli:
Thank you @BewareMyPower
I will work on a fix.
Probably early next week

PR #974, opened 2021-12-16
Bump log4j to 2.16.0

wenbingshen:
### Motivation
Bump log4j to 2.16.0 for branch-2.8.0

### Modifications
Update log4j version from 2.13.3 to 2.16.0.

We need to prepare new patch releases for the affected kop version.


BewareMyPower:
branch-2.8.0 is not maintained anymore.

Issue #908, opened 2021-11-15
Flaky-test: KafkaAuthorizationKafkaMultitenantTenantMetadataTest.testListTopic


BewareMyPower:
KafkaAuthorizationKafkaMultitenantTenantMetadataTest is flaky. The testListTopic test method fails sporadically.

[example failure](https://github.com/streamnative/kop/runs/4210083340?check_suite_focus=true)

```
Error:  Tests run: 15, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 24.993 s <<< FAILURE! - in io.streamnative.pulsar.handlers.kop.KafkaAuthorizationKafkaMultitenantTenantMetadataTest
Error:  testListTopic(io.streamnative.pulsar.handlers.kop.KafkaAuthorizationKafkaMultitenantTenantMetadataTest)  Time elapsed: 0.078 s  <<< FAILURE!
java.lang.AssertionError: expected [2] but found [0]
	at org.testng.Assert.fail(Assert.java:96)
	at org.testng.Assert.failNotEquals(Assert.java:776)
	at org.testng.Assert.assertEqualsImpl(Assert.java:137)
	at org.testng.Assert.assertEquals(Assert.java:118)
	at org.testng.Assert.assertEquals(Assert.java:652)
	at org.testng.Assert.assertEquals(Assert.java:662)
	at io.streamnative.pulsar.handlers.kop.KafkaAuthorizationTestBase.testListTopic(KafkaAuthorizationTestBase.java:337)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:124)
	at org.testng.internal.InvokeMethodRunnable.runOne(InvokeMethodRunnable.java:54)
	at org.testng.internal.InvokeMethodRunnable.run(InvokeMethodRunnable.java:44)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

PR #991, opened 2021-12-24
Fix list topic and list offset behavior

Demogorgon314:
#908
### Motivation
In the current implementation, list topic and list offset might get the wrong response, because the authorized future is not finished yet. We need to make sure to authorize future finish correctly.

### Modifications
* Ensure all future is finished when the response

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -580,36 +580,47 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
         final Map<String, TopicName> nonPartitionedTopicMap = Maps.newConcurrentMap();
 
         final String metadataNamespace = kafkaConfig.getKafkaMetadataNamespace();
+        pulsarTopicsFuture = new CompletableFuture<>();
 
         if (topics == null || topics.isEmpty()) {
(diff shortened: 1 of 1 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -580,36 +580,47 @@ protected void handleTopicMetadataRequest(KafkaHeaderAndRequest metadataHar,
...
+                                                    Collections.emptyList()));
+                                            completeOne.run();
+                                            return;
+                                        }
+                                        topicMap.computeIfAbsent(
+                                                topic,
+                                                ignored -> Collections.synchronizedList(new ArrayList<>())
+                                        ).add(topicName);
+                                        completeOne.run();
+                                    })));
```
BewareMyPower, line 620: Could you avoid the indent changes? I've compared carefully for a while and only found two lines (612 and 619) are actually needed.

Currently the indent style is not restricted. We should avoid any change about the existing indent except nearly the whole code block is changed.

BewareMyPower:
LGTM except a comment about code style.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
Demogorgon314: Addressed.

BewareMyPower (approved):

PR #1011, opened 2022-01-10
Implement placeholder 'advertisedAddress' in kafkaAdvertisedListeners

eolivelli:
With this patch when you configure kafkaAdvertisedListeners you can use the special placeholder `advertisedAddress` that picks up the value configured for `advertisedAddress` in broker.conf.

Something like:
`kafkaAdvertisedListeners=PLAINTEXT://advertisedAddress:9092`

Without this placeholder you have to inject the value using other external means, like adding stuff to the Helm Chart (status.podIP) and it is not trivial.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
@@ -477,7 +477,7 @@ public class KafkaServiceConfiguration extends ServiceConfiguration {
                 } catch (UnknownHostException e) {
                     throw new IllegalStateException("hostname is empty and localhost is unknown: " + e.getMessage());
                 }
-            } else if (hostname.equalsIgnoreCase("advertisedAddress")) {
+            } else if (hostname.equals("advertisedAddress")) {
                 hostname = getAdvertisedAddress();
                 listenersReBuilder.append(matcher.group(1))
                         .append("://")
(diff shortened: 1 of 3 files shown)


eolivelli:
I will add docs if the feature is accepted

eolivelli:
@Demogorgon314 @BewareMyPower PTAL

eolivelli:
@Demogorgon314 did you have time to take a look ?

Demogorgon314 (approved):
LGTM.

eolivelli:
@BewareMyPower @Demogorgon314 could we commit this patch ?
or should I add docs together this patch ?
no problem from my side

BewareMyPower:
What I concern is that the placeholder doesn't contain the special characters. Is there a possibility that `advertisedAddress` is a valid domain name.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
```diff
@@ -477,6 +477,13 @@ private String checkAdvertisedListeners(String advertisedListeners) {
                 } catch (UnknownHostException e) {
                     throw new IllegalStateException("hostname is empty and localhost is unknown: " + e.getMessage());
                 }
+            } else if (hostname.equalsIgnoreCase("advertisedAddress")) {
```
BewareMyPower, line 480: If we treat `advertisedAddress` as a special hostname, it's better to use `equals` instead of `equalsIgnoreCase` to avoid a lot of (2^16) candidate hostnames are treated as the special hostname.

eolivelli:
> What I concern is that the placeholder doesn't contain the special characters. Is there a possibility that `advertisedAddress` is a valid domain name.

@BewareMyPower the fact is that special characters don't play well with Helm/Bash/YAML...it is always a pain with escaping  (underscore, dot, dollar symbol...)

So I opted to use the same configuration entry that we have in broker.conf.
It is unlikely that there is a hostname named "advertisedAddress" so it is a safe placeholder.

We will document this very well

I would add that initially I thought to resolve to advertisedAddress the case of "empty hostname", but that would have changed the current behaviour, so "advertisedAddress" is safer and backward compatible

BewareMyPower (approved):

eolivelli:
Thanks!

I will follow up with docs next Monday

PR #1073, opened 2022-02-13
Transactions: don't create a thread per-tenant

eolivelli:
**Modifications**
- use a shared threadpool
- lazily start the drainQueuedTransactionMarkers activity only when needed
- set period to 100ms (instead of 1ms)

**Result**
If a tenant never uses transactions, then you never start the background activity, also we are able to scale the number for active tenants.

A future improvement may be to shutdown the activity when a tenant is idle for a while

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionMarkerChannelManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionMarkerChannelManager.java
@@ -220,6 +220,7 @@ public class TransactionMarkerChannelManager {
                                     TransactionMetadata txnMetadata,
                                     TransactionMetadata.TxnTransitMetadata newMetadata,
                                     String namespacePrefix) {
+        ensureDrainQueuedTransactionMarkersActivity();
         String transactionalId = txnMetadata.getTransactionalId();
         PendingCompleteTxn pendingCompleteTxn = new PendingCompleteTxn(
                 transactionalId,
@@ -244,6 +245,7 @@ public class TransactionMarkerChannelManager {
     }
 
     public void maybeWriteTxnCompletion(String transactionalId) {
+        ensureDrainQueuedTransactionMarkersActivity();
         PendingCompleteTxn pendingCompleteTxn = transactionsWithPendingMarkers.get(transactionalId);
         if (!hasPendingMarkersToWrite(pendingCompleteTxn.txnMetadata)
                 && transactionsWithPendingMarkers.remove(transactionalId, pendingCompleteTxn)) {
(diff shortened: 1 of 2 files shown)


eolivelli:
CI failed, checking

eolivelli:
Ci passed

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionMarkerChannelManager.java
```diff
@@ -475,8 +474,24 @@ private void drainQueuedTransactionMarkers() {
         }
     }
 
+    private synchronized void ensureDrainQueuedTransactionMarkersActivity() {
+        if (drainQueuedTransactionMarkersHandle != null || closed) {
+            return;
+        }
+        drainQueuedTransactionMarkersHandle = scheduler.scheduleWithFixedDelay(() -> {
+            drainQueuedTransactionMarkers();
+        }, 100, 100, TimeUnit.MILLISECONDS);
```
BewareMyPower, line 483: Is there any reason to change the period from 1 second to 100 milliseconds? BTW, I think the 2nd argument (`initialDelay` should be 0).

To keep the original behavior, I think we can pass a `ThreadFactory` to create thread whose name prefix is `kop-transaction-channel-manager-`.

Demogorgon314: ping @eolivelli, PTAL : )

eolivelli: sorry for late reply:

> Is there any reason to change the period from 1 second to 100 milliseconds
1 ms is like having a busy wait, we are using lot of CPU for something that is not doing much work.
we don't have in Pulsar other activities that run so often, also we will start one of this activities per each tenant and this have a significant impact on the Broker.
Should we make this configurable ?

> initialDelay should be 0
there is not much different between 0 and 1ms and 100ms.
I chose to set non-zero because when you start this activity it is likely that you don't need this as soon as possible and so we can let the bootstrap less resource intensive.
If you feel strong I can set this to 0

> ThreadFactory...
I don't understand this point. We cannot set the thread name or the Factory because we are going to use a shared threadpool.
Do you mean that we should create a separated threadpool ?

@BewareMyPower @Demogorgon314

BewareMyPower: > 1 ms is like having a busy wait,

It makes sense.

> If you feel strong I can set this to 0

No, I asked these questions just to confirm your points for the minor behavior changes.

> Do you mean that we should create a separated threadpool ?

No. I just missed the fact that it used a shared thread pool.

This PR is okay to me now. No need to change.

BewareMyPower (approved):

PR #1042, opened 2022-01-28
fix create topic fail when using advertised listener

wangjialing218:
Fixes #1038

controller in METADATA response did not return the advertised listener, which cause create topic fail.

PR #1092, opened 2022-02-16
Fix CreateTopics failure when kafkaAdvertisedListeners are different with kafkaListeners

BewareMyPower:
This PR is a follow up bug fix of https://github.com/streamnative/kop/pull/1042 for branch-2.8.2. It also migrate some util methods and tests from https://github.com/streamnative/kop/pull/864.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -78,6 +78,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
+import java.util.regex.Matcher;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import lombok.Getter;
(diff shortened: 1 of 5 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -2547,7 +2548,14 @@ static Node newNode(InetSocketAddress address) {
...
 
     Node newSelfNode() {
-        return newNode(advertisedEndPoint.getInetAddress());
+        String advertisedListeners = kafkaConfig.getKafkaAdvertisedListeners();
+        String listener = EndPoint.findListener(advertisedListeners, advertisedEndPoint.getListenerName());
+        if (listener == null) {
+            return newNode(advertisedEndPoint.getInetAddress());
+        }
+        final Matcher matcher = EndPoint.matcherListener(listener,
+                listener + " cannot be split into 3 parts");
```
Demogorgon314, line 2557: Can we move this code to `KafkaRequestHandler` constructors?

Demogorgon314 (approved):
LGTM. Just left one question. PTAL.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
BewareMyPower: Good find. I think we can apply this change for master and other branches as well.

PR #1093, opened 2022-02-16
Optimize the way to get the controllerId

wenbingshen:
Related to this #1092 

1. The current brokers metadata is maintained in the AdminManager, and obtaining the controllerId from here can ensure data consistency.
2. Avoid extra complex parsing listeners and handling exceptions.
3. The node ID is calculated from the hostname and port hash. If kop is not configured with ip or hostname, the program will automatically obtain the hostname. One of the problems is: when kop is running, if a new hostname address is added to the local domain name self-resolution, and as for the first position, it may cause the program to return a different host alias, the controller node id will inconsistent with the Broker cache in metadataStore. You can see the PR #648 I submitted a long time ago.
4. Reuse `testMetadataRequestForMultiListeners` test validation controller.
5. A random node from the broker cache is used as the controller. There is no controller concept in kop. It is the same whether it returns to the node itself or randomly selects a node.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AdminManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AdminManager.java
@@ -430,6 +430,7 @@ class AdminManager {
     }
 
     public Collection<? extends Node> getBrokers(String listenerName) {
+
         if (brokersCache.containsKey(listenerName)) {
             return brokersCache.get(listenerName);
         }
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AdminManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AdminManager.java
@@ -430,27 +430,15 @@ class AdminManager {
     }
 
     public Collection<? extends Node> getBrokers(String listenerName) {
-        brokersCacheLock.readLock().lock();
-        try {
-            if (brokersCache.containsKey(listenerName)) {
-                return brokersCache.get(listenerName);
-            }
-
-            return Collections.emptyList();
-        } finally {
-            brokersCacheLock.readLock().unlock();
+        if (brokersCache.containsKey(listenerName)) {
+            return brokersCache.get(listenerName);
         }
+
+        return Collections.emptyList();
     }
 
-    // Since getBrokers method in KopEventThread uses asynchronous,
-    // so we should use readLock here.
     public Map<String, Set<Node>> getAllBrokers() {
-        brokersCacheLock.readLock().lock();
-        try {
-            return brokersCache;
-        } finally {
-            brokersCacheLock.readLock().unlock();
-        }
+        return brokersCache;
     }
 
     public void setBrokers(Map<String, Set<Node>> newBrokers) {
@@ -465,27 +453,21 @@ class AdminManager {
 
     // only set when setBrokers
     private void setControllerId(Map<String, Set<Node>> newBrokers) {
+        Map<String, Integer> newControllerId = Maps.newHashMap();
         newBrokers.forEach((listenerName, brokers) -> {
             if (brokers.size() == 0) {
-                controllerId.put(listenerName, MetadataResponse.NO_CONTROLLER_ID);
+                newControllerId.put(listenerName, MetadataResponse.NO_CONTROLLER_ID);
             } else {
                 List<Node> nodes = Lists.newArrayList(brokers);
-                controllerId.put(listenerName,
+                newControllerId.put(listenerName,
                         nodes.size() > 1 ? nodes.get(random.nextInt(brokers.size())).id() : nodes.get(0).id());
             }
         });
+        this.controllerId = newControllerId;
     }
 
     // always get the controllerId directly from the cache
     public int getControllerId(String listenerName) {
-        brokersCacheLock.readLock().lock();
-        try {
-            if (controllerId.containsKey(listenerName)) {
-                return controllerId.get(listenerName);
-            }
-            return MetadataResponse.NO_CONTROLLER_ID;
-        } finally {
-            brokersCacheLock.readLock().unlock();
-        }
+        return controllerId.getOrDefault(listenerName, MetadataResponse.NO_CONTROLLER_ID);
     }
 }
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AdminManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AdminManager.java
@@ -16,6 +16,7 @@ package io.streamnative.pulsar.handlers.kop;
 import static io.streamnative.pulsar.handlers.kop.utils.delayed.DelayedOperationKey.TopicKey;
 import static org.apache.kafka.common.requests.CreateTopicsRequest.TopicDetails;
 
+import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import io.streamnative.pulsar.handlers.kop.exceptions.KoPTopicException;
 import io.streamnative.pulsar.handlers.kop.utils.KopTopic;
@@ -28,6 +29,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
@@ -49,6 +51,7 @@ import org.apache.kafka.common.protocol.Errors;
 import org.apache.kafka.common.requests.ApiError;
 import org.apache.kafka.common.requests.CreateTopicsRequest;
 import org.apache.kafka.common.requests.DescribeConfigsResponse;
+import org.apache.kafka.common.requests.MetadataResponse;
 import org.apache.pulsar.client.admin.PulsarAdmin;
 import org.apache.pulsar.client.admin.PulsarAdminException;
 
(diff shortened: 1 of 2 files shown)


BewareMyPower:
I think we should use a separate lock for `controllerId`. Here is my design:

```java
    private void setControllerId(Map<String, Set<Node>> newBrokers) {
        final Map<String, Integer> listenerToControllerId = new HashMap<>();
        newBrokers.forEach((listenerName, brokers) -> {
            if (brokers.size() == 0) {
                listenerToControllerId.put(listenerName, MetadataResponse.NO_CONTROLLER_ID);
            } else {
                List<Node> nodes = Lists.newArrayList(brokers);
                listenerToControllerId.put(listenerName,
                        nodes.size() > 1 ? nodes.get(random.nextInt(brokers.size())).id() : nodes.get(0).id());
            }
        });
        synchronized (this.listenerToControllerId) {
            this.listenerToControllerId.clear();
            this.listenerToControllerId.putAll(listenerToControllerId);
        }
    }

    public int getControllerId(String listenerName) {
        synchronized (this.listenerToControllerId) {
            return listenerToControllerId.getOrDefault(listenerName, MetadataResponse.NO_CONTROLLER_ID);
        }
    }

    private final Map<String, Integer> listenerToControllerId = Maps.newHashMap();
```

wenbingshen:
@BewareMyPower Thanks for your review. Sorry, there is a problem with my understanding, here reading does not require any read locks, we only need to ensure the `synchronization` of writing; `synchronized` will cause unnecessary competition between the request processing thread and `KopEventThread` , as well as competition between different request processing, This will bring serious performance problems, so we don't need any read-related lock operations here, because for the kafka client, the short-term expired metadata cache will not cause any problems for the client, the client only Just retry the `metadata` request.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AdminManager.java
```diff
@@ -439,10 +445,30 @@ private void handleUpdatePartitionsAsync(String topic,
...
+                newControllerId.put(listenerName,
+                        nodes.size() > 1 ? nodes.get(random.nextInt(brokers.size())).id() : nodes.get(0).id());
+            }
+        });
+        this.controllerId = newControllerId;
+    }
+
+    // always get the controllerId directly from the cache
+    public int getControllerId(String listenerName) {
+        return controllerId.getOrDefault(listenerName, MetadataResponse.NO_CONTROLLER_ID);
```
BewareMyPower, line 472: Could you ensure `getControllerId` and `setControllerId` are called in the same thread? If not, it would be not thread-safe without the lock. The `volatile` keyword can only ensure the `controllerId` is the latest reference, but you can't ensure a `null` value is read.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AdminManager.java
```diff
@@ -439,10 +445,30 @@ private void handleUpdatePartitionsAsync(String topic,
     public void setBrokers(Map<String, Set<Node>> newBrokers) {
         brokersCacheLock.writeLock().lock();
         try {
+            setControllerId(newBrokers);
```
BewareMyPower, line 448: BTW, based on my example code before, this method can move out of the locked block.

```java
    public void setBrokers(Map<String, Set<Node>> newBrokers) {
        setControllerId(newBrokers);
        brokersCacheLock.writeLock().lock();
        try {
            this.brokersCache = newBrokers;
        } finally {
            brokersCacheLock.writeLock().unlock();
        }
    }

    private void setControllerId(Map<String, Set<Node>> newBrokers) {
        final Map<String, Integer> listenerToControllerId = /* create a local map using a local variable `newBrokers` */;
        synchronized (this.listenerToControllerId) { // local the field to make it thread safe, there's no need to protect it by `brokersCacheLock`.
            this.listenerToControllerId.clear();
            this.listenerToControllerId.putAll(listenerToControllerId);
        }
    }

    private final Map<String, Integer> listenerToControllerId = Maps.newHashMap();
```

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AdminManager.java
wenbingshen: Outdated `controllerId` is allowed, this will not cause any problems, it is better for the client to briefly retry the metadata request than the server to use lock synchronization, otherwise we may see all the request handle threads are stuck getting the `controllerId` , and the `KopEventThread` is updating the `controllerId` now.

The possible consistency issue here is the visibility order of `controllerId` and `brokerCache`, if you are very strong about this, we can define a `Metadata` or `MetadataCache` class to wrap `controllerId` and `brokerCache` for synchronization purposes.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AdminManager.java
wenbingshen: you can see this issue: https://issues.apache.org/jira/browse/KAFKA-7019

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AdminManager.java
BewareMyPower: | Time | Operation                                                    | `controllerId` |
| ---- | ------------------------------------------------------------ | -------------- |
| 1    | setControllerId({"A": 1})                                    | {"A": 1}       |
| 2    | getControllerId("B") read `controllerId`: {"A": 1}           | {"A": 1}       |
| 3    | setControllerId({"B": 2})                                    | {"B": 2}       |
| 4    | getControllerId("B") **getOrDefault: operation on an outdated reference** |                |

I'm not sure whether the operation on an outdated reference is valid.

wenbingshen: Get controllerId in kafka:
https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/server/metadata/ZkMetadataCache.scala#L297

BewareMyPower (approved):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/AdminManager.java
```diff
@@ -439,10 +445,30 @@ private void handleUpdatePartitionsAsync(String topic,
...
+                List<Node> nodes = Lists.newArrayList(brokers);
+                newControllerId.put(listenerName,
+                        nodes.size() > 1 ? nodes.get(random.nextInt(brokers.size())).id() : nodes.get(0).id());
+            }
+        });
+        this.controllerId = newControllerId;
+    }
+
+    // always get the controllerId directly from the cache
+    public int getControllerId(String listenerName) {
```
eolivelli, line 471: This code is wrong.
This is a HashMap and it is not accessed inside the "lock".
it is strange that spotbugs does not complaint.

btw accessing a HashMap in a non threadsafe way may lead to unpredictable behaviour.
we should fix this
@Demogorgon314 @BewareMyPower

BewareMyPower: There was a discussion in https://github.com/streamnative/kop/pull/1093#discussion_r809048190. Does the explanation of @wenbingshen make sense to you?

From my perspective, the internal state of `controllerId` never changes once it's set with a new value. Since `controllerId` is a volatile field, we can always see the latest object it points to, and the object has only one possible state to watch.

See the only two references of `controllerId`:

```java
    private void setControllerId(Map<String, Set<Node>> newBrokers) {
        /* ... */
        this.controllerId = newControllerId; // [1] points to another map
    }

    public int getControllerId(String listenerName) {
        return controllerId.getOrDefault(listenerName, MetadataResponse.NO_CONTROLLER_ID); // [2] call getOrDefault, which doesn't modify the internal state.
    }
```

eolivelli: I believe that it is better to follow the best practices and so do not have bad surprises in production :-)

The "volatile" keyword applies only to the access to the reference to the map, not to the contents. So in a multithread env you are not guaranteed that another thread "sees" the new contents of the map, even if it "sees" the map

we must:
- use ConcurentHashMap e use the "readLock"
- remove "volatile" keyword

I am really sure that not following well established patterns will bring us bad surprises and problems to users

BewareMyPower: The point is, the content could never be published to another thread.

The map is first created here:

```java
    private void setControllerId(Map<String, Set<Node>> newBrokers) {
        Map<String, Integer> newControllerId = Maps.newHashMap();
        /* ... */
        this.controllerId = newControllerId;
    }
```

After it's assigned to `controllerId`, there is no way to modify the content because we only expose the immutable view by `getOrDefault` method.

BewareMyPower: JMM's happens-before rule can guarantee another thread can see the latest content before `[3]` because `[3]` is an assignment of a `volatile` variable so that `[2.1]` and `[2.2]` can be seen by other threads.

```java
   private void setControllerId(Map<String, Set<Node>> newBrokers) {
        Map<String, Integer> newControllerId = Maps.newHashMap(); // [1] create an empty map
        newBrokers.forEach((listenerName, brokers) -> {
            if (brokers.size() == 0) {
                newControllerId.put(listenerName, MetadataResponse.NO_CONTROLLER_ID); // [2.1] put a key-value
            } else {
                List<Node> nodes = Lists.newArrayList(brokers);
                newControllerId.put(listenerName, // [2.2] put a key-value
                        nodes.size() > 1 ? nodes.get(random.nextInt(brokers.size())).id() : nodes.get(0).id());
            }
        });
        this.controllerId = newControllerId; // [3]
```

eolivelli: Good. I missed [3].

Thanks for your patience

BewareMyPower:
Hi, could you try to migrate this PR to branch-2.8.2? Because the code of branch-2.8.2 is a little different from master branch.

wenbingshen:
> Hi, could you try to migrate this PR to branch-2.8.2? Because the code of branch-2.8.2 is a little different from master branch.

Sure, please wait a moment.

PR #1094, opened 2022-02-17
[Transaction] Able to disable transaction coordinator

### LINKED ISSUE #1090: [BUG] Kafka transaction coordinator unable to disable
**Describe the bug**

Currently, even if we config `kafkaTransactionCoordinatorEnabled=false`, the Kafka transaction coordinator is still started.

Demogorgon314:
Fixes  #1090

### Motivation
Currently, even if we config kafkaTransactionCoordinatorEnabled=false, the Kafka transaction coordinator is still started.
It happened when the broker has a consumer fetch message.

### Modifications
When `kafkaTransactionCoordinatorEnabled = false`, ensure the transaction coordinator don't start.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -267,7 +267,10 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
     }
 
     public TransactionCoordinator getTransactionCoordinator() {
-        return tenantContextManager.getTransactionCoordinator(getCurrentTenant());
+        if (kafkaConfig.isKafkaTransactionCoordinatorEnabled()) {
+            return tenantContextManager.getTransactionCoordinator(getCurrentTenant());
+        }
+        return null;
     }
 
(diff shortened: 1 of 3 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -267,7 +267,10 @@ public GroupCoordinator getGroupCoordinator() {
     }
 
     public TransactionCoordinator getTransactionCoordinator() {
-        return tenantContextManager.getTransactionCoordinator(getCurrentTenant());
+        if (kafkaConfig.isKafkaTransactionCoordinatorEnabled()) {
+            return tenantContextManager.getTransactionCoordinator(getCurrentTenant());
+        }
+        return null;
```
BewareMyPower, line 273: It's better to throw an exception here rather than returning a `null` value. Otherwise, the caller would have to call `throwIfTransactionCoordinatorDisabled` each time before `getTransactionCoordinator` is called.

Demogorgon314: Good idea.

BewareMyPower (approved):

Issue #1105, opened 2022-02-22
[DISCUSS] Should we distinguish the authorization exception and failure?


BewareMyPower:
See https://github.com/streamnative/kop/blob/459c6cb2de814a362a3900050d650a78a764f898/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java#L666-L677 for example.

Currently, no matter if `authorize` failed with an exception or the topic doesn't have the proper permission, in both case `completeOneAuthFailedTopic.accept(topic, fullTopicName);` will be called. Generally it doesn't make a difference because Kafka client will get the error in both cases.

However, IMO, if an exception was thrown in `authorize`, we should not continue the following process because there might be something wrong at the broker side.

@Demogorgon314 PTAL.

PR #1107, opened 2022-02-23
Refactor the handler for METADATA request

### LINKED ISSUE #1104: [Enhancement] Refactor the handler for METADATA request
`KafkaRequestHandler#handleTopicMetadataRequest` is too long that it has over 300 lines. The code style is bad and it's hard to maintain. It's better to refactoring this method to advance the code readability.

BewareMyPower:
Fixes #1104

### Motivation

In addition to #1104 explained, the existing code has too many nested code. It relies on an atomic integer and multiple callbacks to wait all futures until completed. Since the futures' callbacks are not guaranteed to be executed in a single thread, the helper containers must be thread safe. The better solution is leveraging `CompletableFuture.allOf` method to get all futures' results in a single callback and process the results or handling them to another future.

### Modifications
- Add some helper methods into `CoreUtils` to wait all futures and perform conversion among lists and maps.
- Add `ListPair` class to avoid null check when processing the result of `groupingBy`.
- Add unit tests for methods and class above.
- Add `TopicAndMetadata` class as a cache of the full topic name and the number of partitions, which can be negative to indicate an error.
- Use a more functional programming code style to rewrite the handler for METADATA request.
- Correct an incorrect check for all topics METADATA request, see https://github.com/apache/kafka/blob/8d88b20b2779faa413ffc4c6d2546800e225213f/clients/src/main/java/org/apache/kafka/common/requests/MetadataRequest.java#L162-L164, if `topics` field is empty, only version 0 METADATA request is for all topics.

### TODO
The `authorizeTopicsAsync` method and methods from `CoreUtils` can also be used to simplify other request handlers.


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -499,348 +532,174 @@ protected void handleInactive(KafkaHeaderAndRequest kafkaHeaderAndRequest,
...
+                ListPair.of(futureMap.entrySet()
+                        .stream()
+                        .collect(Collectors.groupingBy(e -> e.getValue().join()))
+                ).map(Map.Entry::getKey));
+    }
+
+    private CompletableFuture<List<TopicAndMetadata>> findTopicMetadata(final ListPair<String> listPair,
+                                                                        final boolean allowTopicAutoCreation) {
+        final Map<String, CompletableFuture<Integer>> futureMap = CoreUtils.listToMap(
+                listPair.getSuccessfulList(),
```
eolivelli, line 613: what happens to topics with errors that are not AUTHORIZATION_FAILURE  ?

it seems to me that below we are reporting them all with AUTHORIZATION_FAILURE

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/TopicAndMetadata.java
```diff
@@ -0,0 +1,98 @@
...
+        return CoreUtils.waitForAll(stream()
+                .map(TopicName::get)
+                .map(lookupFunction)
+                .collect(Collectors.toList()), partitionMetadataList ->
+                new TopicMetadata(
+                        error(),
+                        getOriginalTopic.apply(topic),
+                        KopTopic.isInternalTopic(topic, metadataNamespace),
+                        partitionMetadataList
+                ));
```
wenbingshen, line 77: we can replace 
```
partitionMetadataList ->
                new TopicMetadata(
                        error(),
                        getOriginalTopic.apply(topic),
                        KopTopic.isInternalTopic(topic, metadataNamespace),
                        partitionMetadataList
                )
```
with

```
partitionMetadataList -> toTopicMetadata(getOriginalTopic, metadataNamespace)
```

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -499,348 +532,174 @@ protected void handleInactive(KafkaHeaderAndRequest kafkaHeaderAndRequest,
...
+                Collectors.toMap(
+                        topic -> topic,
+                        topic -> authorize(aclOperation, Resource.of(ResourceType.TOPIC, topic))
+                ));
+        return CoreUtils.waitForAll(futureMap.values()).thenApply(__ ->
+                ListPair.of(futureMap.entrySet()
+                        .stream()
+                        .collect(Collectors.groupingBy(e -> e.getValue().join()))
+                ).map(Map.Entry::getKey));
+    }
```
wenbingshen, line 608: We should handle exceptions that occur when checking permissions. I see that `io.streamnative.pulsar.handlers.kop.KafkaRequestHandler#authorize()` may throw exception.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -499,348 +532,174 @@ protected void handleInactive(KafkaHeaderAndRequest kafkaHeaderAndRequest,
...
-        return topicMapFuture;
+        return topicAndMetadataList;
+    }
+
+    private CompletableFuture<List<String>> authorizeNamespacesAsync(final Collection<String> namespaces,
+                                                                     final AclOperation aclOperation) {
+        final Map<String, CompletableFuture<Boolean>> futureMap = namespaces.stream().collect(
+                Collectors.toMap(
+                        namespace -> namespace,
+                        namespace -> authorize(aclOperation, Resource.of(ResourceType.NAMESPACE, namespace))
```
wenbingshen, line 575: handle exceptions when authorize namespaces.

wenbingshen (changes requested):
Great work., leave some comments. PTAL.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
BewareMyPower: > we are reporting them all with AUTHORIZATION_FAILURE

No. Only those topics are failed to be authorized are reported with AUTHORIZATION_FAILURE.

```java
        ).thenApply(topicAndMetadataList -> { // [1]
            final List<TopicAndMetadata> list = CoreUtils.listToList(listPair.getFailedList(),
                    topic -> new TopicAndMetadata(topic, TopicAndMetadata.AUTHORIZATION_FAILURE)); // [2]
            list.addAll(topicAndMetadataList);
            return list;
        });
```
- `[1]` (`topicAndMetadataList`) is the lookup result by `getPartitionedTopicMetadataAsync`, the error (`numPartitions` field) could also be `INVALID_PARTITIONS`.
- `[2]` (the initial `list`) contains the failed topics from `authorizeTopicsAsync` so they are all reported with `AUTHORIZATION_FAILURE`.

Since `[1]` could be an empty list `Collections.emptyList()` so we cannot call `addAll` on `[1]`. Instead, we should create a new list and call `addAll`, which accepts `[1]` as the argument.

See `getTopicsAsync`:

```java
            return authorizeTopicsAsync(fullTopicNames, AclOperation.DESCRIBE)
                    .thenCompose(authorizedTopicsPair -> findTopicMetadata(authorizedTopicsPair, allowTopicCreation));
```

BewareMyPower: I'll rename them to make the code more readable.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
BewareMyPower: `authorize` doesn't throw an exception, instead, the future will be completed exceptionally. In this case, the future returned by `CoreUtils.waitForAll` will be completed exceptionally. The exception will eventually be processed in `getTopicsAsync`'s `whenComplete`.

It should be noted that there are two failed cases:
- `authorize` returns false, which means the topic doesn't have the expected permission. In this case, an `AUTHORIZATION_FAILURE` will be set.
- `authorize` failed with exception. In this case, we treat it as an unrecoverable failure and prevent the followed operations.

See https://github.com/streamnative/kop/issues/1105. This PR changes the current behavior for the 2nd case.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
BewareMyPower: The reason is similar in the previous comment. There are two cases:

1. Some namespaces doesn't have the proper permission: printed the warning logs and ignore them

```java
            futureMap.entrySet().stream().filter(e -> {
                if (!e.getValue().join()) {
                    log.warn("Failed to authorize {} for ACL operation {}", e.getKey(), aclOperation);
                    return false;
                }
```
2. The authorization failed exceptionally. Go to the `getTopicsAsync`'s callback:

```java
        getTopicsAsync(request, fullTopicNameToOriginal.keySet()).whenComplete((topicAndMetadataList, e) -> {
            if (e != null) 
```

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -499,348 +532,174 @@ protected void handleInactive(KafkaHeaderAndRequest kafkaHeaderAndRequest,
...
+        // Check missed partitions
+        final List<TopicAndMetadata> topicAndMetadataList = new ArrayList<>();
+        topicToPartitionIndexes.forEach((topic, partitionIndexes) -> {
+            Collections.sort(partitionIndexes);
+            final int lastIndex = partitionIndexes.get(partitionIndexes.size() - 1);
+            if (lastIndex < 0) {
+                topicAndMetadataList.add(
+                        new TopicAndMetadata(topic, TopicAndMetadata.NON_PARTITIONED_NUMBER));
+            } else {
+                if (lastIndex == partitionIndexes.size() - 1) {
```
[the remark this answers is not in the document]
BewareMyPower: Yes.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
BewareMyPower: See https://github.com/streamnative/kop/pull/1107/commits/9424ea9a4b7016629e27d5f695a997df2d951364.

BewareMyPower:
Thanks for all your reviews because it's better to have more views on such a refactor, though the existing test already covered much. I've addressed all comments now, PTAL again. @eolivelli @wenbingshen @Demogorgon314

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
wenbingshen: One of my concerns is that if handle permission validation with an unrecoverable error, instead of having no permissions, the client won't be able to receive the exception from the server, even though it doesn't care about the exception, but since the client just got an empty TopicMetadata , the client will continue to retry the `MetadataRequest` , and a large number of `MetadataRequest` will consume more CPU of the broker. In addition, the client may not find that there is a problem with its own program. 

I looked at the implementation of `handleMetadataRequest` in kafka. If an broker level fault is encountered, the broker will return the error to the client.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -499,348 +534,173 @@ protected void handleInactive(KafkaHeaderAndRequest kafkaHeaderAndRequest,
...
-                resultFuture.complete(finalResponse);
+                log.error("[{}] Request {}: Exception fetching metadata, will return null Response",
+                        ctx.channel(), metadataHar.getHeader(), e);
+                resultFuture.complete(KafkaResponseUtils.newMetadata(
+                        allNodes,
+                        clusterName,
+                        controllerId,
+                        Collections.emptyList()));
                 return;
             }
```
wenbingshen, line 685: One of my concerns is that if handle permission validation with an unrecoverable error, instead of having no permissions, the client won't be able to receive the exception from the server, even though it doesn't care about the exception, but since the client just got an empty TopicMetadata , the client will continue to retry the MetadataRequest , and a large number of MetadataRequest will consume more CPU of the broker. In addition, the client may not find that there is a problem with its own program.

I looked at the implementation of handleMetadataRequest in kafka. If an broker level fault is encountered, the broker will return the error to the client.

wenbingshen:
> Thanks for all your reviews because it's better to have more views on such a refactor, though the existing test already covered much. I've addressed all comments now, PTAL again. @eolivelli @wenbingshen @Demogorgon314

A great code refactoring!

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
BewareMyPower: It changes the current behavior as well. I think if we can handle this exceptional case in another PR so that we can add some tests to verify it.

BewareMyPower: After rethinking for a while, I think we need to complete it exceptionally here. See the latest commit.

wenbingshen: Good work! Thanks. :)

wenbingshen (approved):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/CoreUtils.java
```diff
@@ -76,4 +81,38 @@
             ));
     }
 
+    public static <T> List<T> concatList(final List<T> list1, final List<T> list2) {
+        try {
+            // if `list` can be modifiable, modify the `list1` in place.
+            list1.addAll(list2);
+            return list1;
+        } catch (UnsupportedOperationException ignored) {
```
eolivelli, line 89: Personally I don't like this very much.
because it is not clear if the input list will be modified or not.

why can't we always create a new List ?

```
List<T> result = new ArrayList<>(list1.size() + list2.size());
result.addAll(list1);
result.addAll(list2);
return result;
```

it should be more efficient and less error prone.

Please note that `list1` will be resized when you add `list2` elements so in anycase you have a arraycopy behind the scenes, but you don't know, because it depends on the internals of `list1`.

eolivelli (changes requested):
very good
I left one comment about `concatList` I believe it is better to code it another way

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/CoreUtils.java
BewareMyPower: Yeah. It's right. The "in place" solution doesn't looks so good.

BewareMyPower: I used `Streams#concat` now. See https://github.com/streamnative/kop/pull/1107/commits/b79b36a560322711754b008617256a1fec79b645.

Demogorgon314: Why don't we use @eolivelli 's way? Look like it is more efficient.

eolivelli (approved):
LGTM

Demogorgon314 (approved):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/CoreUtils.java
BewareMyPower: Good point. BTW, I found we can just use `ListUtils#union` instead of reimplementing the same logic.

BewareMyPower: In addition, I think it's not in a critical path. The performance gap usually doesn't make a difference.

BewareMyPower:
This PR will be cherry-picked to branch-2.8.2 after https://github.com/streamnative/kop/pull/1120 is merged.

Issue #1053, opened 2022-02-08
Flaky-test: IdempotentProducerTest.testIdempotentProducer


BewareMyPower:
IdempotentProducerTest is flaky. The testIdempotentProducer test method fails sporadically.

[example failure](https://github.com/streamnative/kop/runs/5107060964?check_suite_focus=true)

```
Error:  testIdempotentProducer(io.streamnative.pulsar.handlers.kop.IdempotentProducerTest)  Time elapsed: 1.34 s  <<< FAILURE!
org.junit.ComparisonFailure: expected:<test[510]> but was:<test[824]>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at io.streamnative.pulsar.handlers.kop.IdempotentProducerTest.testIdempotentProducer(IdempotentProducerTest.java:108)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:124)
	at org.testng.internal.InvokeMethodRunnable.runOne(InvokeMethodRunnable.java:54)
	at org.testng.internal.InvokeMethodRunnable.run(InvokeMethodRunnable.java:44)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
```

BewareMyPower:
The test is still flaky, see [example failure](https://github.com/streamnative/kop/runs/5232538770?check_suite_focus=true). PTAL. @Demogorgon314 

```
Error:  testIdempotentProducer(io.streamnative.pulsar.handlers.kop.IdempotentProducerTest)  Time elapsed: 20.014 s  <<< FAILURE!
[...]

PR #1124, opened 2022-02-28
[flaky-test] Increase test idempotent produce timeout time

Demogorgon314:
Fixes: #1053

### Motivation

After examining the timeout test log, no blocking phenomenon was found. So the current solution is to increase test idempotent produce timeout time. Maybe it took too much time to send 1000 messages.


BewareMyPower (approved):

BewareMyPower:
It looks like this test is still flaky, see https://github.com/streamnative/kop/runs/5439450488?check_suite_focus=true.

Demogorgon314:
> It looks like this test is still flaky, see https://github.com/streamnative/kop/runs/5439450488?check_suite_focus=true.

Yes, I saw it, too, I'll check it out.

PR #1125, opened 2022-03-01
Transactions/Multi tenancy: use a Pulsar topic to generate Producer Ids

eolivelli:
### Motivation

Currently we are using ZooKeeper to generate the Producer ids.
This has two problems:
- it does not play well the multi-tenancy story, as the id sequence is shared among the tenants
- it does not scale well in the number of tenants

### Implementation
- Add a new configuration option `kafkaTransactionProducerIdsStoredOnPulsar` (disabled by default)
- Create a new topic to store the data
- Allocate ids in blocks (for efficiency)
- Support compaction (we write always with the same key, and read from the compacted topic)
- In case of lost topic or deleted messages the counter starts from scratch (ideally we should set infinite retention on the topic)

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
@@ -14,9 +14,13 @@
 package io.streamnative.pulsar.handlers.kop.coordinator.transaction;
 
 import java.math.BigInteger;
+import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Collectors;
+import java.util.stream.LongStream;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.pulsar.client.api.Message;
 import org.apache.pulsar.client.api.MessageId;
@@ -33,12 +37,13 @@ import org.apache.pulsar.client.api.Reader;
  */
 @Slf4j
 public class PulsarStorageProducerIdManagerImpl implements ProducerIdManager {
-
+    static final int BLOCK_SIZE = 1000;
     private final AtomicLong nextId = new AtomicLong(0);
     private final String topic;
     private final PulsarClient pulsarClient;
     private CompletableFuture<Reader<byte[]>> reader;
     private CompletableFuture<Void> currentReadHandle;
+    private final ConcurrentLinkedQueue<Long> availableIdsLocally = new ConcurrentLinkedQueue<>();
 
     private synchronized CompletableFuture<Reader<byte[]>> ensureReaderHandle() {
         if (reader == null) {
@@ -108,6 +113,12 @@ public class PulsarStorageProducerIdManagerImpl implements ProducerIdManager {
 
     @Override
     public synchronized CompletableFuture<Long> generateProducerId() {
+        Long booked = availableIdsLocally.poll();
+        if (booked != null) {
+            log.debug("Returning pre-allocated id {} for {}", booked, topic);
+            return CompletableFuture.completedFuture(booked);
+        }
+        log.debug("Allocating new block of ids for {}", topic);
         // we could get rid of the Exclusive Producer if we had Message.getIndex()
         // introduced in 2.9.0 https://github.com/apache/pulsar/pull/11553
         CompletableFuture<Producer<byte[]>> producerHandle = pulsarClient.newProducer()
@@ -121,17 +132,23 @@ public class PulsarStorageProducerIdManagerImpl implements ProducerIdManager {
             // wait for local cache to be up-to-date
             CompletableFuture<Long> dummy = ensureLatestData(true)
                     .thenCompose((___) -> {
-                        final long result = nextId.get() + 1;
+                        long start = nextId.get();
+                        List<Long> block = generateBlock(start + BLOCK_SIZE, BLOCK_SIZE);
+                        final long nextAvailableId = start + BLOCK_SIZE;
                         // write to Pulsar
-                        byte[] serialized = BigInteger.valueOf(result).toByteArray();
+                        byte[] serialized = BigInteger.valueOf(nextAvailableId).toByteArray();
                         CompletableFuture<Long>  res =  opProducer
                                 .newMessage()
                                 .key("") // always the same key, this way we can rely on compaction
                                 .value(serialized)
                                 .sendAsync()
                                 .thenApply((msgId) -> {
-                                    log.debug("{} written {} as {}", this, result, msgId);
-                                    nextId.set(result);
+                                    log.debug("{} written {} as {}", this, nextAvailableId, msgId);
+                                    nextId.set(nextAvailableId);
+                                    availableIdsLocally.addAll(block);
+                                    long result = availableIdsLocally.remove();
+                                    log.debug("Returning allocated id {} for {}, new range: {}-{}",
+                                            result, topic, start, nextAvailableId);
                                     return result;
                                 });
                         return res;
@@ -142,6 +159,12 @@ public class PulsarStorageProducerIdManagerImpl implements ProducerIdManager {
             });
             return dummy;
         });
+    }
+
+    private static List<Long> generateBlock(long start, int blockSize) {
+        return LongStream.range(start, start + blockSize)
+                .boxed()
+                .collect(Collectors.toList());
     }
 
     public PulsarStorageProducerIdManagerImpl(String topicName, PulsarClient pulsarClient) {
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
@@ -1,0 +1,161 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.streamnative.pulsar.handlers.kop.coordinator.transaction;
+
+import java.math.BigInteger;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.atomic.AtomicLong;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.pulsar.client.api.Message;
+import org.apache.pulsar.client.api.MessageId;
+import org.apache.pulsar.client.api.Producer;
+import org.apache.pulsar.client.api.ProducerAccessMode;
+import org.apache.pulsar.client.api.PulsarClient;
+import org.apache.pulsar.client.api.Reader;
+
+/**
+ * ProducerIdManager is the part of the transaction coordinator that provides ProducerIds in a unique way
+ * such that the same producerId will not be assigned twice across multiple transaction coordinators.
+ *
+ * ProducerIds are managed via a Pulsar non-partitioned topic.
+ */
+@Slf4j
+public class PulsarStorageProducerIdManagerImpl implements ProducerIdManager {
+
+    private final AtomicLong nextId = new AtomicLong(0);
+    private final String topic;
+    private final PulsarClient pulsarClient;
+    private CompletableFuture<Reader<byte[]>> reader;
+    private CompletableFuture<Void> currentReadHandle;
+
+    private synchronized CompletableFuture<Reader<byte[]>> ensureReaderHandle() {
+        if (reader == null) {
+            reader = pulsarClient.newReader()
+                    .topic(topic)
+                    .startMessageId(MessageId.earliest)
+                    .startMessageIdInclusive()
+                    .readCompacted(true)
+                    .createAsync();
+        }
+        return reader;
+    }
+
+    private CompletableFuture<Void> readNextMessageIfAvailable(Reader<byte[]> reader) {
+        return reader
+                .hasMessageAvailableAsync()
+                .thenCompose(hasMessageAvailable -> {
+                    if (hasMessageAvailable == null
+                            || !hasMessageAvailable) {
+                        return CompletableFuture.completedFuture(null);
+                    } else {
+                        CompletableFuture<Message<byte[]>> opMessage = reader.readNextAsync();
+                        return opMessage.thenCompose(msg -> {
+                            byte[] value = msg.getValue();
+                            long newId = new BigInteger(value).longValue();
+                            log.debug("{} Read {} from {}", this, newId, msg.getMessageId());
+                            nextId.set(newId);
+                            return readNextMessageIfAvailable(reader);
+                        });
+                    }
+                });
+    }
+
+    synchronized CompletableFuture<Void> ensureLatestData(boolean beforeWrite) {
+        if (currentReadHandle != null) {
+            if (beforeWrite) {
+                // we are inside a write loop, so
+                // we must ensure that we start to read now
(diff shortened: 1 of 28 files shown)


eolivelli:
The Codacy StaticCode Analysis error is wrong, I am  using "==" because I want to compare the references. 

image

BewareMyPower:
Could you take a first look at this PR? @Demogorgon314

codelipenghui:
@eolivelli @BewareMyPower I have disabled the rule

image

But not sure how to rerun the check, I would like to close and reopen the PR.

BewareMyPower:
It looks like the Codacy check didn't rerun. Could you force push the branch again to see whether it works?

image

Demogorgon314 (commented):
Great work! Overall LGTM. Just left some small comments.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
```diff
@@ -0,0 +1,197 @@
...
+                .hasMessageAvailableAsync()
+                .thenCompose(hasMessageAvailable -> {
+                    if (hasMessageAvailable == null
+                            || !hasMessageAvailable) {
+                        return CompletableFuture.completedFuture(null);
+                    } else {
+                        CompletableFuture<Message<byte[]>> opMessage = reader.readNextAsync();
+                        return opMessage.thenCompose(msg -> {
+                            byte[] value = msg.getValue();
+                            long newId = new BigInteger(value).longValue();
```
BewareMyPower, line 72: Should we use `Schema.INT64.encode(value)` instead of usage of `BigInteger`? Or just use a `Producer<Long>`?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
```diff
@@ -0,0 +1,197 @@
...
+    @Override
+    public synchronized CompletableFuture<Long> generateProducerId() {
+        Long booked = availableIdsLocally.poll();
+        if (booked != null) {
+            log.debug("Returning pre-allocated id {} for {}", booked, topic);
+            return CompletableFuture.completedFuture(booked);
+        }
+        log.debug("Allocating new block of ids for {}", topic);
+        // we could get rid of the Exclusive Producer if we had Message.getIndex()
+        // introduced in 2.9.0 https://github.com/apache/pulsar/pull/11553
```
BewareMyPower, line 124: Just a note. I think we can add a separate PR to leverage `Message.getIndex` API for Pulsar 2.9.0 or higher in future.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
```diff
@@ -0,0 +1,197 @@
...
+                                            result, topic, start, nextAvailableId);
+                                    return result;
+                                });
+                        return res;
+                    });
+            // ensure that we release the exclusive producer in any case
+            dummy.whenComplete((___, err) -> {
+                opProducer.closeAsync();
+            });
+            return dummy;
```
BewareMyPower, line 161: Maybe we can return the future returned by `whenComplete` like:

```java
            return dummy.whenComplete((___, err) -> {
                opProducer.closeAsync();
            });
```

In addition, it's better to print the error message if `closeAsync` failed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
```diff
@@ -0,0 +1,197 @@
...
+    public CompletableFuture<Void> initialize() {
+       return ensureLatestData(false);
+    }
+
+    @Override
+    public synchronized void shutdown() {
+        if (reader != null) {
+            reader.whenComplete((r, e) -> {
+                if (r != null) {
+                    r.closeAsync();
```
BewareMyPower, line 191: It's better to print the error message if closeAsync failed.

Demogorgon314:
@eolivelli Please check the review comments, and rebase to the code. Thanks!

eolivelli:
sure, 
I will update this patch today.
thanks

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
eolivelli: good idea

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
eolivelli: fixed

tests/src/test/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImplTest.java
```diff
@@ -0,0 +1,211 @@
...
+        super.internalSetup();
+    }
+
+    @AfterClass
+    @Override
+    protected void cleanup() throws Exception {
+        super.internalCleanup();
+    }
+
+    @Test
```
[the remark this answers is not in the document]
eolivelli: added to all the methods

eolivelli (commented):
Thank you @BewareMyPower  and @Demogorgon314 
I have updated the patch and rebased it to latest master

eolivelli:
it looks like Codacy indeed passed
https://github.com/streamnative/kop/pull/1125/checks?check_run_id=5534807834

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
```diff
@@ -0,0 +1,210 @@
...
+                    .createAsync();
+        }
+        return reader;
+    }
+
+    private CompletableFuture<Void> readNextMessageIfAvailable(Reader<byte[]> reader) {
+        return reader
+                .hasMessageAvailableAsync()
+                .thenCompose(hasMessageAvailable -> {
+                    if (hasMessageAvailable == null
```
BewareMyPower, line 65: When could `hasMessageAvailable` be `null`?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
```diff
@@ -0,0 +1,210 @@
...
+                            || !hasMessageAvailable) {
+                        return CompletableFuture.completedFuture(null);
+                    } else {
+                        CompletableFuture<Message<byte[]>> opMessage = reader.readNextAsync();
+                        return opMessage.thenCompose(msg -> {
+                            byte[] value = msg.getValue();
+                            long newId = new BigInteger(value).longValue();
+                            log.debug("{} Read {} from {}", this, newId, msg.getMessageId());
```
BewareMyPower, line 73: Please add the `if (log.isDebugEnabled())` block.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
```diff
@@ -0,0 +1,210 @@
...
+        CompletableFuture<Reader<byte[]>> readerHandle = ensureReaderHandle();
+        final CompletableFuture<Void> newReadHandle =
+                readerHandle.thenCompose(this::readNextMessageIfAvailable);
+        currentReadHandle = newReadHandle;
+        return newReadHandle.whenComplete((a, b) -> {
+            endReadLoop(newReadHandle);
+            if (b != null) {
+                throw new CompletionException(b);
+            }
+        });
```
BewareMyPower, line 106:
```suggestion
        return newReadHandle.thenApply(result -> {
            endReadLoop(newReadHandle);
            return result;
        });
```

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
```diff
@@ -0,0 +1,210 @@
...
+            }
+            // if there is an ongoing read operation then complete it
+            return currentReadHandle;
+        }
+        // please note that the read operation is async,
+        // and it is not execute inside this synchronized block
+        CompletableFuture<Reader<byte[]>> readerHandle = ensureReaderHandle();
+        final CompletableFuture<Void> newReadHandle =
+                readerHandle.thenCompose(this::readNextMessageIfAvailable);
+        currentReadHandle = newReadHandle;
```
BewareMyPower, line 100: I see this method returns `currentReadHandle` when it's not null. Should here be following code?

```java
currentReadHandle = newReadHandle.thenApply(/* ... */);
return currentReadHandle;
```

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
```diff
@@ -0,0 +1,210 @@
...
+                                .key("") // always the same key, this way we can rely on compaction
+                                .value(serialized)
+                                .sendAsync()
+                                .thenApply((msgId) -> {
+                                    if (log.isDebugEnabled()) {
+                                        log.debug("{} written {} as {}", this, nextAvailableId, msgId);
+                                    }
+                                    nextId.set(nextAvailableId);
+                                    availableIdsLocally.addAll(block);
+                                    long result = availableIdsLocally.remove();
```
BewareMyPower, line 154: Should these two lines be composed into an atomic operation? I see `availableIdsLocally.poll()` could be called in another thread?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
```diff
@@ -0,0 +1,210 @@
...
+                                    if (log.isDebugEnabled()) {
+                                        log.debug("Returning allocated id {} for {}, new range: {}-{}",
+                                                result, topic, start, nextAvailableId);
+                                    }
+                                    return result;
+                                });
+                        return res;
```
BewareMyPower, line 161: We can just remove the `res` variable.

```java
return opProducer.newMessage()./* ... */
```

BewareMyPower:
I've left some comments, PTAL. In addition, I'm a little confused about `availableIdsLocally`.  It looks like the `addAll` operations can be called concurrently.

```java
                                    availableIdsLocally.addAll(block);
                                    long result = availableIdsLocally.remove();
```

How can we ensure the order of `addAll` operations? Or is the order not necessary?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
```diff
@@ -0,0 +1,210 @@
...
+    private CompletableFuture<Reader<byte[]>> reader;
+    private CompletableFuture<Void> currentReadHandle;
+    private final ConcurrentLinkedQueue<Long> availableIdsLocally = new ConcurrentLinkedQueue<>();
+
+    private synchronized CompletableFuture<Reader<byte[]>> ensureReaderHandle() {
+        if (reader == null) {
+            reader = pulsarClient.newReader()
+                    .topic(topic)
+                    .startMessageId(MessageId.earliest)
+                    .startMessageIdInclusive()
```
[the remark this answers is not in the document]
eolivelli: I can remove it. good catch

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
eolivelli: in theory it cannot be `null`, but we don't know (if there is a Pulsar client bug).
Unfortunately nullable Booleans are very nasty.
I would keep this for the sake of "defensive coding"

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
eolivelli: updated

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
eolivelli: sure

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
eolivelli: good idea.
I have updated the patch.
And I also now currentReadHandle will be the same CompletableFuture returned in case of non null `currentReadHandle`.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
eolivelli: in theory there can be only 1 Exclusive producer working on the topic (partition) at a time. Even in case of multiple executions in the same JVM.
So there should be no other thread accessing `availableIdsLocally` when we are here

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
eolivelli: sure (I had a problem with IDE auto-completion and that way IntelliJ worked well...)
I have simplified the code. thanks

eolivelli (commented):
@BewareMyPower I have addressed your comments.

Sorry for late reply

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/PulsarStorageProducerIdManagerImpl.java
BewareMyPower: I see when `currentReadHandle` is not null, the `ensureLatestData` returns `currentReadHandle`. Otherwise, it returns the `currentReadHandle.thenApply(/* ... */)`. Is it intentional?

I mean, the correct implementation might be

```java
        currentReadHandle = newReadHandle.thenApply((__) -> {
            endReadLoop(newReadHandle);
            return null;
        });
        return currentReadHandle;
```

eolivelli: I tried to do what you are suggesting 10 minutes ago, but if I do saw than tests are not passing anymore :) 

the `endReadLoop` is only a cleanup operation.

- When you are "reading" then it doesn't matter which "currentReadHandle" you pick
- When you are "writing" then we must be sure that we pick the latest "currentReadHandle" and we complete the read, because we are inside the Exclusive Producer guard and nobody else (even in the same JVM) can write.

BewareMyPower: Oh, sorry, I understand now.

Demogorgon314 (approved):

Issue #1134, opened 2022-03-04
[BUG] Failed to get owned topic list caused by getListOfNonPersistentTopics


BewareMyPower:
**Describe the bug**

There is something wrong at Pulsar side that `NamespaceService#getListOfNonPersistentTopics` might fail. It would lead to a consistent failure in `CacheInvalidator` and `TransactionStateRecover`.

```
2022-03-04T09:07:33,952+0000 [metadata-store-6-1] ERROR io.streamnative.pulsar.handlers.kop.KafkaProtocolHandler - Failed to get owned topic list for CacheInvalidator when triggering bundle ownership change pulsar/10.0.0.26:8080/0x00000000_0xffffffff.
java.util.concurrent.CompletionException: org.apache.pulsar.broker.web.RestException: Namespace not found
        at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:331) ~[?:?]
        at java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:1099) ~[?:?]
        at java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:2235) ~[?:?]
        at org.apache.pulsar.broker.namespace.NamespaceService.getListOfNonPersistentTopics(NamespaceService.java:1214) ~[io.streamnative-pulsar-broker-2.9.2.6.jar:2.9.2.6]
        at org.apache.pulsar.broker.namespace.NamespaceService.getFullListOfTopics(NamespaceService.java:1096) ~[io.streamnative-pulsar-broker-2.9.2.6.jar:2.9.2.6]
        at org.apache.pulsar.broker.namespace.NamespaceService.getOwnedTopicListForNamespaceBundle(NamespaceService.java:1103) ~[io.streamnative-pulsar-broker-2.9.2.6.jar:2.9.2.6]
        at io.streamnative.pulsar.handlers.kop.KafkaProtocolHandler$CacheInvalidator.invalidateBundleCache(KafkaProtocolHandler.java:155) ~[?:?]
        at io.streamnative.pulsar.handlers.kop.KafkaProtocolHandler$CacheInvalidator.onLoad(KafkaProtocolHandler.java:183) ~[?:?]
        at org.apache.pulsar.broker.namespace.NamespaceService.notifyNamespaceBundleOwnershipListener(NamespaceService.java:1077) ~[io.streamnative-pulsar-broker-2.9.2.6.jar:2.9.2.6]
[...]

PR #1136, opened 2022-03-05
Avoid listing non-persistent topics and optimize the bundle listener

BewareMyPower:
Fixes https://github.com/streamnative/kop/issues/1134

### Motivation

In KoP, when namespace bundle ownership changes, there's no need to get the owned non-persistent topics, while `getOwnedTopicListForNamespaceBundle` tries to get all owned topics. There is a bug when getting non-persistent topics at Pulsar side and exceptions might be thrown in the bundle listener's callback, which makes KoP not work well for the bundle load or unload events.

In addition, there are multiple implementations of `NamespaceBundleOwnershipListener` that are registered to the `NamespaceService`. Each listener calls the `getOwnedTopicListForNamespaceBundle` method, this is redundant.

### Modifications

- Only retain a common implementation of `NamespaceBundleOwnershipListener`. Implement a `getOwnedPersistentTopicList` to get only persistent topics of a bundle's owned topics.
- Allow `NamespaceBundleOwnershipListener` to register a `TopicOwnershipListener`, which processes a topic, for example, update the metadata cache, or call the immigration/emigration methods of group/transaction coordinator when the load or unload event happens. It also provides a `test` method to avoid iterating over the topics when the namespace doesn't match.
- Migrate the original implementations of `NamespaceBundleOwnershipListener` to `TopicOwnershipListener`.
- Remove the checks for the default topic namespace (`kafkaTopicNs`) when consumer group/transaction metadata topics are loaded or unloaded.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
@@ -41,7 +41,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
-import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
@@ -54,12 +53,11 @@ import org.apache.kafka.common.utils.Time;
 import org.apache.pulsar.broker.PulsarServerException;
 import org.apache.pulsar.broker.PulsarService;
 import org.apache.pulsar.broker.ServiceConfiguration;
-import org.apache.pulsar.broker.namespace.NamespaceBundleOwnershipListener;
+import org.apache.pulsar.broker.namespace.NamespaceService;
 import org.apache.pulsar.broker.protocol.ProtocolHandler;
 import org.apache.pulsar.broker.service.BrokerService;
 import org.apache.pulsar.client.admin.PulsarAdmin;
 import org.apache.pulsar.client.admin.PulsarAdminException;
-import org.apache.pulsar.common.naming.NamespaceBundle;
 import org.apache.pulsar.common.naming.NamespaceName;
 import org.apache.pulsar.common.naming.TopicName;
 import org.apache.pulsar.common.policies.data.ClusterData;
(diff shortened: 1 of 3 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
```diff
@@ -520,11 +283,34 @@ private TransactionCoordinator createAndBootTransactionCoordinator(String tenant
...
+                @Override
+                public String name() {
+                    return "TransactionStateRecover";
+                }
+
+                @Override
+                public boolean test(NamespaceName namespaceName) {
+                    return namespaceName.equals(kafkaMetaNs) || namespaceName.equals(kafkaTopicNs);
+                }
+            });
```
wenbingshen, line 313: In PR #964, we expect `metadata topics` to appear within the `metadata namespace`. For the `test` method, we do not need to judge the `kafkaNamespace`.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
```diff
@@ -550,10 +336,34 @@ private GroupCoordinator createAndBootGroupCoordinator(String tenant) {
...
+
+                @Override
+                public String name() {
+                    return "OffsetAndTopicListener";
+                }
+
+                @Override
+                public boolean test(NamespaceName namespaceName) {
+                    return namespaceName.equals(kafkaMetaNs) || namespaceName.equals(kafkaTopicNs);
+                }
```
wenbingshen, line 365: the same as above, for `group metadata topics`, we do not need to judge whether `kafkaTopicNs` is hit.

wenbingshen (commented):
Good work. Just leave a minor point. PTAL.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
BewareMyPower: Sounds reasonable. It looks like this redundant check has been here from the beginning when `OffsetAndTopicListener` and `TransactionStateRecover` were added.

Demogorgon314 (approved):

wenbingshen (approved):

Issue #1137, opened 2022-03-06
[BUG] NPE caused by asyncGetCursorByOffset


Demogorgon314:
**Describe the bug**
```
2022-03-04T08:46:34,187+0000 [BookKeeperClientWorker-OrderedExecutor-7-0] ERROR io.streamnative.pulsar.handlers.kop.MessageFetchContext - KafkaTopicConsumerManager.asyncGetCursorByOffset(362273) failed for topic test-topic-OjvIkSI-0000-22.
java.lang.NullPointerException: null
        at io.streamnative.pulsar.handlers.kop.KafkaTopicConsumerManager.lambda$asyncGetCursorByOffset$7(KafkaTopicConsumerManager.java:241) ~[?:?]
         at java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:642) ~[?:?]
         at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506) ~[?:?]
         at java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:2073) ~[?:?]
         at org.apache.bookkeeper.mledger.impl.ManagedLedgerImpl$15.findEntryFailed(ManagedLedgerImpl.java:1668) ~[io.streamnative-managed-ledger-2.9.2.6.jar:2.9.2.6]
         at org.apache.bookkeeper.mledger.impl.OpFindNewest.readEntryFailed(OpFindNewest.java:142) ~[io.streamnative-managed-ledger-2.9.2.6.jar:2.9.2.6]
         at org.apache.bookkeeper.mledger.impl.EntryCacheImpl.lambda$asyncReadEntry0$1(EntryCacheImpl.java:233) ~[io.streamnative-managed-ledger-2.9.2.6.jar:2.9.2.6]
         at java.util.concurrent.CompletableFuture.uniExceptionally(CompletableFuture.java:986) [?:?]
         at java.util.concurrent.CompletableFuture$UniExceptionally.tryFire(CompletableFuture.java:970) [?:?]
         at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506) [?:?]
         at java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2088) [?:?]
         at org.apache.bookkeeper.client.PendingReadOp.submitCallback(PendingReadOp.java:645) [org.apache.bookkeeper-bookkeeper-server-4.14.4.jar:4.14.4]
[...]

PR #1138, opened 2022-03-06
Fix NPE when managed ledger find entry failed

Demogorgon314:
Fixes: #1137 

### Motivation

The `ManagedLedgerImpl#asyncFindPosition` method might return future null, when we use it we should check position is null first.

### Modifications
Add NULL check.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicConsumerManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicConsumerManager.java
@@ -237,6 +237,9 @@ public class KafkaTopicConsumerManager implements Closeable {
         }
 
         return MessageMetadataUtils.asyncFindPosition(ledger, offset, skipMessagesWithoutIndex).thenApply(position -> {
+            if (position == null) {
+                return null;
+            }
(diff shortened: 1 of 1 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicConsumerManager.java
```diff
@@ -237,6 +237,9 @@ public void close() {
         }
 
         return MessageMetadataUtils.asyncFindPosition(ledger, offset, skipMessagesWithoutIndex).thenApply(position -> {
+            if (position == null) {
+                return null;
+            }
```
BewareMyPower, line 242: This change LGTM because `ManagedLedgerImpl#asyncFindPosition` might return a null completed future. But since this future will be put into `KafkaTopicConsumerManager#cursors`, could you double check if the null check is performed when processing the future's result from `cursors`?

Demogorgon314: I checked all the places where `cursors` was used, all path has null check.

BewareMyPower (approved):

PR #1140, opened 2022-03-07
Fix NOT_COORDINATOR error caused by wrong brokers in Metadata response

BewareMyPower:
### Motivation

After https://github.com/streamnative/kop/pull/1107 was cherry-picked into branch-2.8.2, some tests of `KafkaListenerNameTest` are failed. Because that PR forgot to migrate the following code:

https://github.com/streamnative/kop/blob/51563feb582424251c8a8345ed9a7606546cf77e/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java#L792-L796

After that when the advertised listener is different with the listener, the brokers field in `MetadataResponse` doesn't contain the advertised listener, while the leader and ISR in `MetadataResponse.PartitionMetadata` use the advertised listener. Then Kafka client won't be able to find the advertised listener in brokers field.

### Modifications

Add the partition leader's node to the brokers field if the brokers field doesn't contain the node.


BewareMyPower:
It looks like in master branch, the leader and ISR fields both use the listener (0.0.0.0:15003) but not the advertised listener (localhost:15003).

image

Demogorgon314 (approved):

Issue #1142, opened 2022-03-08
Confirm the behavior change for METADATA response since 2.9 and later


BewareMyPower:
See details in https://github.com/streamnative/kop/pull/1140#issuecomment-1060844730.

BewareMyPower:
The description is wrong. The `advertisedListeners` config has been useless in KoP since #864, so some tests in `KafkaListenerNameTests` are outdated. We should update these tests.

PR #1144, opened 2022-03-08
Update tests with the correct way to configure multiple listeners

BewareMyPower:
Fixes https://github.com/streamnative/kop/issues/1142

### Motivation

Currently, `testConnectListenerNotExist` and `testMultipleListenerName` both use the legacy way to configure multiple listeners. Though KoP still supports the legacy way, see [listeners configuration in branch-2.8.2](https://github.com/streamnative/kop/blob/branch-2.8.2/docs/configuration.md#listeners), the advertised listener in METADATA response is not as expected. We should make it clear in tests.

### Modifications
- Use the correct way to configure multiple listeners in `testConnectListenerNotExist` and verify the brokers cache.
- Rename `testMultipleListenerName` to `testLegacyMultipleListenerName` and emphasize the brokers cache doesn't use the host name in `advertisedListeners`.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
@@ -77,6 +77,7 @@ public class KafkaProtocolHandler implements ProtocolHandler, TenantContextManag
     private PrometheusMetricsProvider statsProvider;
     @Getter
     private KopBrokerLookupManager kopBrokerLookupManager;
+    @Getter
     private AdminManager adminManager = null;
     private SystemTopicClient txnTopicClient;
     private DelayedOperationPurgatory<DelayedOperation> producePurgatory;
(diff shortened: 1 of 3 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
```diff
@@ -77,6 +77,7 @@
     private PrometheusMetricsProvider statsProvider;
     @Getter
     private KopBrokerLookupManager kopBrokerLookupManager;
+    @Getter
```
Demogorgon314, line 80: Should add a `@VisibleForTesting` annotation.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KopProtocolHandlerTestBase.java
```diff
@@ -782,9 +784,20 @@ protected Properties newKafkaAdminClientProperties() {
...
+    }
+
+    public static <T> T getFirst(Set<T> set) {
+        Assert.assertNotNull(set);
+        final Iterator<T> iterator = set.iterator();
+        Assert.assertTrue(iterator.hasNext());
+        return iterator.next();
+    }
+
     public KafkaChannelInitializer getFirstChannelInitializer() {
```
Demogorgon314, line 798: The `getFirstChannelInitializer` method never used, can be removed now.

Issue #1147, opened 2022-03-09
[BUG] NPE when handle produce request


Demogorgon314:
**Describe the bug**
```
2022-03-09T04:25:01,041+0000 [pulsar-io-4-10] ERROR io.streamnative.pulsar.handlers.kop.storage.PartitionLog - Failed to handle produce request for public/default/perf-t1-1
java.lang.NullPointerException: null
	at io.streamnative.pulsar.handlers.kop.format.EncodeResult.updateProducerStats(EncodeResult.java:93) ~[?:?]
	at io.streamnative.pulsar.handlers.kop.storage.PartitionLog.publishMessages(PartitionLog.java:265) ~[?:?]
	at io.streamnative.pulsar.handlers.kop.storage.PartitionLog.lambda$appendRecords$2(PartitionLog.java:218) ~[?:?]
	at io.streamnative.pulsar.handlers.kop.storage.PartitionLog.appendRecords(PartitionLog.java:226) ~[?:?]
	at io.streamnative.pulsar.handlers.kop.storage.ReplicaManager.lambda$appendRecords$5(ReplicaManager.java:110) ~[?:?]
	at java.util.concurrent.ConcurrentHashMap.forEach(ConcurrentHashMap.java:1603) ~[?:?]
	at io.streamnative.pulsar.handlers.kop.storage.ReplicaManager.appendRecords(ReplicaManager.java:101) ~[?:?]
	at io.streamnative.pulsar.handlers.kop.KafkaRequestHandler.lambda$handleProduceRequest$34(KafkaRequestHandler.java:783) ~[?:?]
	at io.streamnative.pulsar.handlers.kop.KafkaRequestHandler.lambda$null$35(KafkaRequestHandler.java:832) ~[?:?]
	at java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:859) ~[?:?]
	at java.util.concurrent.CompletableFuture.uniWhenCompleteStage(CompletableFuture.java:883) ~[?:?]
	at java.util.concurrent.CompletableFuture.whenComplete(CompletableFuture.java:2251) ~[?:?]
	at io.streamnative.pulsar.handlers.kop.KafkaRequestHandler.lambda$handleProduceRequest$36(KafkaRequestHandler.java:816) ~[?:?]
	at java.util.HashMap.forEach(HashMap.java:1337) [?:?]
[...]

PR #1148, opened 2022-03-09
Fix npe when update producer states

Demogorgon314:
Fixes: #1147

### Motivation

See: #1147

The NPE happened when `KafkaTopicManager#references` removed the producer before `getReferenceProducer`. 

### Modifications

Use `computeIfAbsent` to make sure we can get a producer instance before we use it.

### Documentation

Need to update docs? 

- [x] `no-need-doc` 
  
This is a bug fix no need for docs.


BewareMyPower (commented):
`KafkaTopicManager#getProducer` is unused now and can be removed.

BewareMyPower (approved):

wenbingshen (approved):

PR #1155, opened 2022-03-10
[Test] Change JUnit to TestNG

Demogorgon314:
### Motivation

We should use TestNG instead of JUnit in the KoP units test.

### Modifications

Change all Junit to TestNG

### Documentation

- [x] `no-need-doc`

### DIFF
--- a/kafka-impl/src/test/java/io/streamnative/pulsar/handlers/kop/coordinator/group/MemberMetadataTest.java
+++ b/kafka-impl/src/test/java/io/streamnative/pulsar/handlers/kop/coordinator/group/MemberMetadataTest.java
@@ -13,17 +13,17 @@
  */
 package io.streamnative.pulsar.handlers.kop.coordinator.group;
 
-import static org.junit.Assert.assertArrayEquals;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.fail;
+import static org.testng.internal.junit.ArrayAsserts.assertArrayEquals;
 
 import com.google.common.collect.Sets;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
-import org.junit.Test;
+import org.testng.annotations.Test;
 import org.testng.collections.Maps;
 
 /**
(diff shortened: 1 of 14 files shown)


kafka-impl/src/test/java/io/streamnative/pulsar/handlers/kop/coordinator/group/MemberMetadataTest.java
```diff
@@ -13,17 +13,17 @@
...
-import static org.junit.Assert.assertArrayEquals;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.fail;
+import static org.testng.internal.junit.ArrayAsserts.assertArrayEquals;
```
BewareMyPower, line 20: We can use `assertEquals` for `testMetadata` to eliminate the usage of internal JUnit package.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/BasicEndToEndTestBase.java
```diff
@@ -311,7 +311,7 @@ protected void kafkaConsumeCommitMessage(KConsumer kConsumer,
             ConsumerRecords<Integer, String> records = kConsumer.getConsumer().poll(Duration.ofSeconds(1));
             for (ConsumerRecord<Integer, String> record : records) {
                 Integer key = record.key();
-                assertEquals(messageStrPrefix + key.toString(), record.value());
+                assertEquals(record.value(), messageStrPrefix + key.toString());
```
BewareMyPower, line 314: It looks like there is no need to swap these arguments here.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/utils/timer/TimerTaskListTest.java
```diff
@@ -59,50 +59,50 @@ public void testAll() {
...
+        assertEquals(size(list3), 6);
 
         list3.forEach(TimerTask::cancel);
-        assertEquals(0, size(list1));
-        assertEquals(0, size(list2));
-        assertEquals(0, size(list3));
+        assertEquals(size(list1), 0);
+        assertEquals(size(list2), 0);
+        assertEquals(size(list3), 0);
     }
```
BewareMyPower, line 106: Could you avoid swapping the two arguments of `assertEquals`?

tests/src/test/java/io/streamnative/pulsar/handlers/kop/BasicEndToEndTestBase.java
Demogorgon314: TestNG‘s `assertEquals`  method define is `assertEquals(String actual, String expected)`, is different from Junit, we should swap the arguments.

BewareMyPower: Would it make any difference? I mean, if the two overloads have any difference, we must be careful to use `assertEquals` in tests.

Demogorgon314: Yes, the failed message is different.

Example:
```java
@Test
public void test() {
    assertEquals("expected message in junit","actual message in junit");
}
```
In Junit the message should be：
```
org.junit.ComparisonFailure: 
Expected :expected message in junit
Actual   :actual message in junit
```
If we don't swap the arguments, the failed message is wrong.

```
java.lang.AssertionError: 
Expected :actual message in junit
Actual   :expected message in junit
```

BewareMyPower (approved):

PR #1156, opened 2022-03-10
Fix when store group id failed the find coordinator future never complete

Demogorgon314:
### Motivation

Currently, when store group id failed, the `resultFuture` will never complete, it will cause client request timeout, we should continue the subsequent operations whether store group id is successful or not. Since the group id is only used to collect consumer metrics.

### Modifications

1. Continue the subsequent operations whether store group id is successful or not.
2. Add units test to verify it.

### Documentation

- [x] `no-need-doc`

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -889,10 +889,10 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
                         log.warn("Store groupId failed, the groupId might already stored.", ex);
                     }
                     findBroker(TopicName.get(pulsarTopicName))
-                            .thenAccept(node -> {
-                                if (node.error() != Errors.NONE) {
+                            .whenComplete((node, throwable) -> {
+                                if (node.error() != Errors.NONE || throwable != null) {
                                     log.error("[{}] Request {}: Error while find coordinator.",
-                                            ctx.channel(), findCoordinator.getHeader());
+                                            ctx.channel(), findCoordinator.getHeader(), throwable);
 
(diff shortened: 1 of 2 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -882,29 +882,30 @@ protected void handleFindCoordinatorRequest(KafkaHeaderAndRequest findCoordinato
...
-                                resultFuture.complete(KafkaResponseUtils
-                                        .newFindCoordinator(Errors.LEADER_NOT_AVAILABLE));
-                                return;
-                            }
+                .whenComplete((__, ex) -> {
+                    if (ex != null) {
+                        log.warn("Store groupId failed, the groupId might already stored.", ex);
+                    }
+                    findBroker(TopicName.get(pulsarTopicName))
+                            .thenAccept(node -> {
```
eolivelli, line 892: what happens if findBroker fails ?
we are not going to complete `resultFuture` even in that case

Demogorgon314: The `findBroker` method will never return a failed future, I think we don't need to handle it.

eolivelli: "currently" it does not return a failure future, you we cannot know (especially for internal errors)
I believe it is better to add a handler for exceptions even in that case.
(just replace thenAccept with whenComplete)

Demogorgon314: Yes, good suggestion. I have addressed.

eolivelli (approved):
LGTM

BewareMyPower (approved):

PR #1167, opened 2022-03-11
add send queue config for system client

### LINKED ISSUE #1146: [BUG] Write offset topic fail due to producer sending queue full
**Describe the bug**
During performance test, client received server error response, casue by write offset topic fail due to producer sending queue full.

**Expected behavior**
1. This kind of log should be ERROR level, not DEBUG level
2. Current system producer is using default setting, which send queue size is 1000. Change producer setting to `blockIfQueueIsFull` and/or `memoryLimit` to avoid this happen.

wangjialing218:
Fixes #1146

### Motivation

See #1146

### Modifications

1. Add `blockIfQueueFull` when create system producer
2. Add `maxPendingMessages` configuration for system producer. I suggest change the default value from 1000 to 10000, since offset topic's payload is small (less than 200 bytes), it won't take mush memory and could avoid send queue full happen.

### Documentation
  
- [x] `no-need-doc`

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
@@ -184,6 +184,12 @@ public class KafkaServiceConfiguration extends ServiceConfiguration {
 
     @FieldContext(
             category = CATEGORY_KOP,
+            doc = "send queue size of system client to produce system topic."
+    )
+    private int maxPendingMessages = 10000;
+
+    @FieldContext(
+            category = CATEGORY_KOP,
             doc = "Zookeeper path for storing kop consumer group"
     )
     private String groupIdZooKeeperPath = "/client_group_id";
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
@@ -186,7 +186,7 @@ public class KafkaServiceConfiguration extends ServiceConfiguration {
             category = CATEGORY_KOP,
             doc = "send queue size of system client to produce system topic."
     )
-    private int maxPendingMessages = 10000;
+    private int kafkaMetaMaxPendingMessages = 10000;
 
     @FieldContext(
             category = CATEGORY_KOP,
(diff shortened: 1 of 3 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
```diff
@@ -182,6 +182,12 @@
     )
     private long offsetsRetentionCheckIntervalMs = OffsetConfig.DefaultOffsetsRetentionCheckIntervalMs;
 
+    @FieldContext(
+            category = CATEGORY_KOP,
+            doc = "send queue size of system client to produce system topic."
+    )
+    private int maxPendingMessages = 10000;
```
eolivelli, line 189: what about adding "kafka" prefix ?

eolivelli (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
wenbingshen: +1 to this. Just left a minor suggestion, you can leave it out. Can we use for example `kafkaMetaMaxPendingMessages`? Otherwise I can easily think of `io.streamnative.pulsar.handlers.kop.KafkaServiceConfiguration#maxReadEntriesNum`, which is a configuration related to client-side.

BewareMyPower: I suggest adding a config whose value is a config file like

```properties
private String kopMetaProducerConfigFile = null;
```

If it's not null, the producer of group coordinator will use the config file to initialize some configurations. This way could be more flexible.

wangjialing218: @eolivelli @wenbingshen done, changed to kafkaMetaMaxPendingMessages
@BewareMyPower We could add the client config file in another pr. This pr aim to change the default value for send queue size.

eolivelli (approved):
LGTM

can you please update the docs ?

Demogorgon314 (approved):
LGTM. Can you please update the docs? You can update it in next PR.

wenbingshen (approved):
LGTM. Thanks.

BewareMyPower (approved):

BewareMyPower:
Add the `doc-required` label. We should add related configuration docs into https://github.com/streamnative/kop/blob/master/docs/configuration.md of all active branches.

PR #1174, opened 2022-03-13
Fix message disorder issue

Demogorgon314:
### Motivation

Currently, the produced message might be disorder, because `topicFuture ` might already done, so we have not queued the produce future.  this future finish might disorder.

See:
https://github.com/streamnative/kop/blob/c03cd9ec2c8da8078b45151f598469b8af4dfb6d/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java#L224-L231

We can add a sleep in `publishMessages` method, it will increase the reproduction probability.
```java
// For control message we don't need to check deduplication.
if (appendInfo.isControlBatch()) {
    offsetFuture = publishControlMessage(persistentTopic, byteBuf, numMessages);
} else {
    offsetFuture = publishNormalMessage(persistentTopic, byteBuf, appendInfo);
}
try {
    TimeUnit.MILLISECONDS.sleep(400);
} catch (InterruptedException e) {
    e.printStackTrace();
}
```

The second problem is we need to use `Errors.OUT_OF_ORDER_SEQUENCE_NUMBER ` instead of `Errors.DUPLICATE_SEQUENCE_NUMBER `, the `DUPLICATE_SEQUENCE_NUMBER` is used to client-side exception.

### Modifications

1. Make sure all produce future are queued
2. Use `Errors.OUT_OF_ORDER_SEQUENCE_NUMBER` instead of `Errors.DUPLICATE_SEQUENCE_NUMBER`
3. Handle `MessageDeduplication.MessageDupUnknownException` exception

### Documentation

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
- [ ] `doc`

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessagePublishContext.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/MessagePublishContext.java
@@ -27,6 +27,7 @@ import org.apache.kafka.common.protocol.Errors;
 import org.apache.pulsar.broker.service.BrokerServiceException;
 import org.apache.pulsar.broker.service.Topic;
 import org.apache.pulsar.broker.service.Topic.PublishContext;
+import org.apache.pulsar.broker.service.persistent.MessageDeduplication;
 
 /**
  * Implementation for PublishContext.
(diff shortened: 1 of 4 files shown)


Demogorgon314:
After applying this patch, I run the CI many times, disorder issue did not reproduce.

eolivelli (approved):
great catch !

PR #1180, opened 2022-03-15
fix client reconnect fail after bundle unload

### LINKED ISSUE #1111: [BUG] Fail to reconnect after bundle unload
**To Reproduce**
Broker version: 2.9.2
KoP version:2.9.2.4

Use benchmark to do performance test.
Kafka producer received NOT_LEADER_OR_FOLLOWER when bundle unload happed and failed to reconnect.

I used multi-listener in k8s, I'm not sure if this is related with multi-listener.

[benchmark.log](https://github.com/streamnative/kop/files/8131479/benchmark.log)
[pulsar.log](https://github.com/streamnative/kop/files/8131480/pulsar.log)
[kop.log](https://github.com/streamnative/kop/files/8131481/kop.log)

wangjialing218:
Fixes #1111

### Motivation
refer to #1111 
When there are 3 brokers or more in cluster, one broker may not receive the bundle load and unload event if the bundle is not owned by the broker, in this case the broker will not able to invalidate the look up cache for the topics in this bundle.

### Modifications
We should not cache all topics' look up result, only to cache those topics' look up result which are owned by the broker.

### Verifying this change
I have checked in my local clusters with 3 brokers.
We need to add the third broker in `DistributedClusterTest` to verify this issue in unit test later.

### Documentation
  
- [x] `no-need-doc`

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopBrokerLookupManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopBrokerLookupManager.java
@@ -141,8 +141,8 @@ public class KopBrokerLookupManager {
             log.error("No node for broker {} under loadBalance", internalListenerAddress);
             return null;
         }
-        if (serviceLookupData.get().getProtocol(KafkaProtocolHandler.PROTOCOL_NAME).isPresent() &&
-                serviceLookupData.get().getProtocol(KafkaProtocolHandler.PROTOCOL_NAME).get()
+        if (serviceLookupData.get().getProtocol(KafkaProtocolHandler.PROTOCOL_NAME).isPresent()
+                && serviceLookupData.get().getProtocol(KafkaProtocolHandler.PROTOCOL_NAME).get()
                         .equals(selfAdvertisedListeners)){
             // the topic is owned by this broker, cache the look up result
             LOOKUP_CACHE.put(topic, CompletableFuture.completedFuture(internalListenerAddress));
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopBrokerLookupManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopBrokerLookupManager.java
@@ -40,6 +40,7 @@ public class KopBrokerLookupManager {
 
     private final LookupClient lookupClient;
     private final MetadataStoreCacheLoader metadataStoreCacheLoader;
+    private final String selfAdvertisedListeners;
 
     private final AtomicBoolean closed = new AtomicBoolean(false);
 
@@ -50,6 +51,7 @@ public class KopBrokerLookupManager {
         this.lookupClient = KafkaProtocolHandler.getLookupClient(pulsarService);
         this.metadataStoreCacheLoader = new MetadataStoreCacheLoader(pulsarService.getPulsarResources(),
                 conf.getBrokerLookupTimeoutMs());
+        this.selfAdvertisedListeners = conf.getKafkaAdvertisedListeners();
     }
 
     public CompletableFuture<Optional<InetSocketAddress>> findBroker(String topic,
@@ -98,7 +100,11 @@ public class KopBrokerLookupManager {
         if (log.isDebugEnabled()) {
             log.debug("Handle Lookup for topic {}", topicName);
         }
-        return LOOKUP_CACHE.computeIfAbsent(topicName, this::lookupBroker);
+
+        if (LOOKUP_CACHE.containsKey(topicName)) {
+            return LOOKUP_CACHE.get(topicName);
+        }
+        return lookupBroker(topicName);
     }
 
     private CompletableFuture<InetSocketAddress> lookupBroker(final String topic) {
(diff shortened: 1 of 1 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopBrokerLookupManager.java
```diff
@@ -98,7 +100,11 @@ public KopBrokerLookupManager(KafkaServiceConfiguration conf, PulsarService puls
         if (log.isDebugEnabled()) {
             log.debug("Handle Lookup for topic {}", topicName);
         }
-        return LOOKUP_CACHE.computeIfAbsent(topicName, this::lookupBroker);
```
eolivelli, line 101: I can't understand this change.
can you please explain ?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopBrokerLookupManager.java
```diff
@@ -135,6 +141,12 @@ private String getAdvertisedListener(InetSocketAddress internalListenerAddress,
             log.error("No node for broker {} under loadBalance", internalListenerAddress);
             return null;
         }
+        if (serviceLookupData.get().getProtocol(KafkaProtocolHandler.PROTOCOL_NAME).isPresent()
+                && serviceLookupData.get().getProtocol(KafkaProtocolHandler.PROTOCOL_NAME).get()
+                        .equals(selfAdvertisedListeners)){
```
eolivelli, line 146: this may be a list, I am not sure we should compare it this way

eolivelli (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopBrokerLookupManager.java
wangjialing218: It's the return value of `KafkaProtocolHandler.getProtocolDataToAdvertise()`, which value is `kafkaConfig.getKafkaAdvertisedListeners()`, I compare here to check if the look up result is the broker itself, in this case we cache the look up result since the topic is owned by this broker.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopBrokerLookupManager.java
wangjialing218: If we cache all topics' lookup result, we will meet the issue here 
https://github.com/streamnative/kop/issues/1111#issuecomment-1066583891

eolivelli: but `computeIfAbsent` is like 
```
if (map.containsKey()) {
 return map.get();
}
Result r = compute();
put(r)
return r;
```

so it is basically the same code, but without concurrency guarantees

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopBrokerLookupManager.java
eolivelli: makes sense

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopBrokerLookupManager.java
```diff
@@ -98,7 +100,11 @@ public KopBrokerLookupManager(KafkaServiceConfiguration conf, PulsarService puls
         if (log.isDebugEnabled()) {
             log.debug("Handle Lookup for topic {}", topicName);
         }
-        return LOOKUP_CACHE.computeIfAbsent(topicName, this::lookupBroker);
+
+        if (LOOKUP_CACHE.containsKey(topicName)) {
+            return LOOKUP_CACHE.get(topicName);
+        }
+        return lookupBroker(topicName);
```
BewareMyPower, line 107: The composite operation of `containsKey` and `get` is not thread safe. We can use a single call of `get` here.

```suggestion
        final CompletableFuture<InetSocketAddress> future = LOOKUP_CACHE.get(topicName);
        return (future != null) ? future : lookupBroker(topicName);
```

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopBrokerLookupManager.java
```diff
@@ -135,6 +141,12 @@ private String getAdvertisedListener(InetSocketAddress internalListenerAddress,
...
             return null;
         }
+        if (serviceLookupData.get().getProtocol(KafkaProtocolHandler.PROTOCOL_NAME).isPresent()
+                && serviceLookupData.get().getProtocol(KafkaProtocolHandler.PROTOCOL_NAME).get()
+                        .equals(selfAdvertisedListeners)){
+            // the topic is owned by this broker, cache the look up result
+            LOOKUP_CACHE.put(topic, CompletableFuture.completedFuture(internalListenerAddress));
+        }
 
         return serviceLookupData.get().getProtocol(KafkaProtocolHandler.PROTOCOL_NAME).map(kafkaAdvertisedListeners ->
```
BewareMyPower, line 151: I think we can simplify the code by

```java
        return serviceLookupData.get().getProtocol(KafkaProtocolHandler.PROTOCOL_NAME).map(kafkaAdvertisedListeners -> {
            if (kafkaAdvertisedListeners.equals(selfAdvertisedListeners)) {
                // the topic is owned by this broker, cache the look up result
                LOOKUP_CACHE.put(topic, CompletableFuture.completedFuture(internalListenerAddress));
            }
            return Optional.ofNullable(advertisedEndPoint)
                    .map(endPoint -> EndPoint.findListener(kafkaAdvertisedListeners, endPoint.getListenerName()))
                    .orElse(EndPoint.findFirstListener(kafkaAdvertisedListeners));
        }).orElseThrow(() -> new IllegalStateException(
                "No kafkaAdvertisedListeners found in broker " + internalListenerAddress));
```

There is no need to call `serviceLookupData.get().getProtocol(KafkaProtocolHandler.PROTOCOL_NAME)` for multiple times.

wangjialing218: Done

BewareMyPower (approved):

BewareMyPower:
I've opened issue #1185 to track the following test.

BewareMyPower:
Remove the `release/2.8.2.x` label at the moment since the `LOOKUP_CACHE` is different in branch-2.8.2.

PR #1187, opened 2022-03-16
Add basic IN/OUT bytes metrics

eolivelli:
### Motivation

We need a way to measure the network traffic originating directly from KOP clients

### Modifications

Add two simple metrics KOP_TOTAL_BYTES_IN and KOP_TOTAL_BYTES_OUT

### Documentation

- [x] `doc-required`

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
@@ -184,6 +184,7 @@ public abstract class KafkaCommandDecoder extends ChannelInboundHandlerAdapter {
     public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
         // Get a buffer that contains the full frame
         ByteBuf buffer = (ByteBuf) msg;
+        requestStats.getTotalBytesIn().add(buffer.readableBytes());
 
         // Update parse request latency metrics
         final BiConsumer<Long, Throwable> registerRequestParseLatency = (timeBeforeParse, throwable) -> {
@@ -432,6 +433,7 @@ public abstract class KafkaCommandDecoder extends ChannelInboundHandlerAdapter {
 
                     final ByteBuf result = responseToByteBuf(response, request);
                     channel.writeAndFlush(result).addListener(future -> {
+                        requestStats.getTotalBytesOut().add(result.readableBytes());
                         if (response instanceof ResponseCallbackWrapper) {
                             ((ResponseCallbackWrapper) response).responseComplete();
                         }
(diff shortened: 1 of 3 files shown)


eolivelli:
I will be happy to add docs/tests if you agree (@BewareMyPower @Demogorgon314 ) on these two new metrics

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
```diff
@@ -432,6 +433,7 @@ protected void writeAndFlushResponseToClient(Channel channel) {
 
                     final ByteBuf result = responseToByteBuf(response, request);
                     channel.writeAndFlush(result).addListener(future -> {
+                        requestStats.getNetworkTotalBytesOut().add(result.readableBytes());
```
BewareMyPower, line 436: Should we check `future.isSuccess()`?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopServerStats.java
```diff
@@ -92,4 +92,10 @@
     String KOP_EVENT_QUEUE_SIZE = "KOP_EVENT_QUEUE_SIZE";
     String KOP_EVENT_QUEUED_LATENCY = "KOP_EVENT_QUEUED_LATENCY";
     String KOP_EVENT_LATENCY = "KOP_EVENT_LATENCY";
+
+    /**
+     * Network stats.
+     */
+    String KOP_NETWORK_TOTAL_BYTES_IN = "KOP_NETWORK_TOTAL_BYTES_IN";
+    String KOP_NETWORK_TOTAL_BYTES_OUT = "KOP_NETWORK_TOTAL_BYTES_OUT";
```
BewareMyPower, line 100: Maybe we can remove the `KOP_` prefix in the string. Because in prometheus's metric string, the field already starts with `kop_server_` prefix. In addition, please also add the related verifications to `testMetricsProvider`.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
eolivelli: I did not check `future.isSuccess()` because actually it is not so important.
in case of future failure we don't know if anything has been sent to the client or not.

I can add "if future.isSuccess()" if you prefer, not a big deal.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KopServerStats.java
eolivelli: you are right !

My eyes weren't seeing the prefix !

I have updated the name, and I am adding tests, thanks

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
eolivelli: done

wenbingshen (approved):

eolivelli:
@wenbingshen good catch
I updated the help

BewareMyPower (commented):
I think we still need to add two assertions in `MetricsProviderTest#testMetricsProvider`.

```java
        // network stats
        Assert.assertTrue(sb.toString().contains("NETWORK_TOTAL_BYTES_IN"));
        Assert.assertTrue(sb.toString().contains("NETWORK_TOTAL_BYTES_OUT"));
```

eolivelli:
@BewareMyPower I have added the assertions, PTAL

BewareMyPower (approved):

PR #1190, opened 2022-03-17
Metrics: expose per-tenant metrics

eolivelli:
### Motivation

We want to track metrics per tenant, in order to be able to debug and inspect the resource usage by each tenant, and also possibly troubleshot problems.

### Modifications

In case of "authentication enabled" and "kafkaEnableMultiTenantMetadata enabled" when a connection  (a `KafkaRequestHandler`) completes authentication with success, we start pushing metrics with a "tenant=pulsar-tenant" label.

The Tenant is the username used for authentication

### Verifying this change

TODO (I will add tests when we agree on this feature)

### Documentation

- [x ] `doc-required`

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
@@ -62,7 +62,7 @@ public abstract class KafkaCommandDecoder extends ChannelInboundHandlerAdapter {
     private final LinkedBlockingQueue<ResponseAndRequest> requestQueue;
     @Getter
     @Setter
-    protected RequestStats requestStats;
+    protected volatile RequestStats requestStats;
     @Getter
     protected final KafkaServiceConfiguration kafkaConfig;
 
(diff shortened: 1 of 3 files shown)


eolivelli:
@BewareMyPower @Demogorgon314 this is kind of a draft.
I wanted to share with you this idea and see your feedback.

I will be happy to add tests and documentation if we agree on this feature

eolivelli:
@Demogorgon314 @BewareMyPower any comment on this patch ?

BewareMyPower:
It makes sense to me.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
```diff
@@ -60,8 +60,9 @@
     protected AtomicBoolean isActive = new AtomicBoolean(false);
     // Queue to make response get responseFuture in order and limit the max request size
     private final LinkedBlockingQueue<ResponseAndRequest> requestQueue;
-
-    protected final RequestStats requestStats;
+    @Getter
+    @Setter
+    protected RequestStats requestStats;
```
BewareMyPower, line 65: Should we use a `Map` here?

eolivelli: there is no need (maybe I can make this 'volatile')
because one KafkaRequestHandler serves only one tenant:

- before auth -> default RequestStats
- after auth -> the RequestStats for the logged tenant

Demogorgon314:
Makes sense to me.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
BewareMyPower: Got it. And yeah, it should be `volatile`.

eolivelli: @BewareMyPower @Demogorgon314 I have updated the patch
PTAL when you have time

eolivelli:
I will be happy to enhance docs when this feature lands

We have been using this feature in production (in clusters with many tenants) and it works very well and it is very useful

BewareMyPower (approved):

PR #1213, opened 2022-03-28
[security] Upgrade jackson-databind to get rid of CVE-2020-36518

nicoloboschi:
See the related pull in Pulsar https://github.com/apache/pulsar/pull/14871

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,7 +38,8 @@
     <redirectTestOutputToFile>true</redirectTestOutputToFile>
 
     <!-- dependencies -->
-    <jackson.version>2.12.1</jackson.version>
+    <jackson.version>2.12.6</jackson.version>
+    <jackson-databind.version>2.12.6.1</jackson-databind.version>
     <kafka.version>2.0.0</kafka.version>
     <lombok.version>1.18.4</lombok.version>
     <mockito.version>2.22.0</mockito.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -38,7 +38,8 @@
     <redirectTestOutputToFile>true</redirectTestOutputToFile>
 
     <!-- dependencies -->
-    <jackson.version>2.12.1</jackson.version>
+    <jackson.version>2.12.6</jackson.version>
+    <jackson-databind.version>2.12.6.1</jackson-databind.version>
```
BewareMyPower, line 42:
```suggestion
    <jackson.version>2.13.2</jackson.version>
    <jackson-databind.version>2.13.2.1</jackson-databind.version>
```

I think we need to upgrade the version to 2.13.2 or higher, see https://nvd.nist.gov/vuln/detail/CVE-2020-36518 and https://github.com/apache/pulsar/pull/14871.

BewareMyPower (changes requested):

pom.xml
nicoloboschi: @BewareMyPower 2.12.6.1 also contains the fix, see [here](https://github.com/FasterXML/jackson-databind/issues/2816#issuecomment-1081046753)

I thought it was better to not upgrade to 2.13.x as it is a more conservative choice. Let me know if you think it's worth to upgrade to 2.13

BewareMyPower: IMO, it's better to keep it consistent with Pulsar's dependency.

BewareMyPower (approved):

PR #1230, opened 2022-04-12
Fix authorization failed when using oauth2

Demogorgon314:
### Motivation

Currently, when KoP uses the oauth2 authentication, the authorization will fail, because we have customized negotiated properties in `SaslServer` but the oauth2 server is not implemented yet. See: https://github.com/streamnative/kop/blob/b6344dc06ec7b8e4f2e64e43ec1227476f70e8fb/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/PlainSaslServer.java#L143-L154

Error log:
```
2022-04-12T02:36:14,551+0000 [pulsar-io-4-1] ERROR io.streamnative.pulsar.handlers.kop.KafkaCommandDecoder - error while handle command:
java.lang.NullPointerException: null
        at io.streamnative.pulsar.broker.AuthUtils.getToken(AuthUtils.java:29) ~[pulsar-broker-auth-oauth2-2.9.2.7.jar:?]
        at io.streamnative.pulsar.broker.authorization.AuthorizationProviderOAuth.isSuperUser(AuthorizationProviderOAuth.java:87) ~[pulsar-broker-auth-oauth2-2.9.2.7.jar:?]
        at org.apache.pulsar.broker.authorization.AuthorizationService.canLookupAsync(AuthorizationService.java:301) ~[io.streamnative-pulsar-broker-common-2.9.2.7.jar:2.9.2.7]
        at io.streamnative.pulsar.handlers.kop.security.auth.SimpleAclAuthorizer.canLookupAsync(SimpleAclAuthorizer.java:145) ~[_iJkmHjzCXmJbuMTkIuXqA/:?]
        at io.streamnative.pulsar.handlers.kop.KafkaRequestHandler.authorize(KafkaRequestHandler.java:2456) ~[_iJkmHjzCXmJbuMTkIuXqA/:?]
        at io.streamnative.pulsar.handlers.kop.KafkaRequestHandler.authorize(KafkaRequestHandler.java:2435) ~[_iJkmHjzCXmJbuMTkIuXqA/:?]
```

### Modifications

Support customized negotiated properties in KopOAuthBearerSaslServer 

### Verifying this change

Modify the SaslOauthKopHandlersTest, and set super users with empty.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
@@ -439,7 +439,7 @@ public class KafkaServiceConfiguration extends ServiceConfiguration {
             doc = "The fully qualified name of a SASL server callback handler class that implements the "
                     + "AuthenticateCallbackHandler interface, which is used for OAuth2 authentication. "
                     + "If it's not set, the class will be Kafka's default server callback handler for "
-                    + "OAUTHBEARER mechanism: OAuthBearerUnsecuredValidatorCallbackHandler."
+                    + "OAUTHBEARER mechanism: KopOAuthBearerUnsecuredValidatorCallbackHandler."
     )
     private String kopOauth2AuthenticateCallbackHandler;
 
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
@@ -22,6 +22,7 @@ import io.netty.buffer.Unpooled;
 import io.netty.channel.ChannelHandlerContext;
 import io.streamnative.pulsar.handlers.kop.KafkaServiceConfiguration;
 import io.streamnative.pulsar.handlers.kop.security.oauth.KopOAuthBearerSaslServer;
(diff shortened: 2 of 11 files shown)


BewareMyPower (commented):
I think we still need to adopt more secure coding practices. The key point is here:

https://github.com/streamnative/kop/blob/b6344dc06ec7b8e4f2e64e43ec1227476f70e8fb/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java#L424-L427

Unlike `getAuthorizationId`, which should return a non-null value if the `SaslServer#evaluateResponse` has been called and no exception has been thrown, we cannot guarantee `getNegotiatedProperty` for a specific input returns a non-null value.

So we must do some null check for it. For example, a fail-fast solution could be

```java
    /**
     * The exception to indicate that the SaslServer doesn't have the expected property
     */
    public static class NoExpectedPropertyException extends AuthenticationException {

        public NoExpectedPropertyException(String propertyName, String msg) {
            super("No expected property for " + propertyName + ": " + msg);
        }
    }

    private static <T> T safeGetProperty(final SaslServer saslServer, final String propertyName) {
        try {
            final Object property = saslServer.getNegotiatedProperty(propertyName);
            if (property == null) {
                throw new NoExpectedPropertyException(propertyName, "property not found");
            }
            return (T) property;
        } catch (ClassCastException e) {
            throw new NoExpectedPropertyException(propertyName, e.getMessage());
        }
    }
```

BewareMyPower (approved):

PR #1237, opened 2022-04-15
Support print full thread dump when the units test timeout

Demogorgon314:
### Motivation

Currently, the KoP some units test might timeout, but we hard to find the reason without the thread dump. 

After researching the TestNG, I find we can implement a listener to dump the thread when a timeout happens. 

See: https://testng.org/doc/documentation-main.html#testng-listeners

### Modifications

Add a TimeOutTestListener to handle the timeout test and print the full thread dump into surefire-reports. 

Now we can easily find the deadlock that happened in the units test.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  

BewareMyPower:
`ThreadDumpUtil` was firstly introduced in Pulsar 2.10.0, so I removed the 2.9.2 and 2.8.3 release labels.

BewareMyPower:
`ThreadDumpUtil` was firstly introduced in Pulsar 2.10.0, so I removed the 2.9.2 and 2.8.3 release labels.

BewareMyPower (approved):

PR #1244, opened 2022-04-20
Add authorization tests for OAuth2 authentication based on Hydra

BewareMyPower:
### Motivation

#1235 introduced Hydra as the OAuth2 server for tests. However, there are some flaws on current test. The creation of the super user role is performed by the `hydra` CLI tool. There is no test for roles other than the super user role. There is also no test to verify if `grantPermissionOnNamespace` or `revokePermissionOnNamespace` works.

### Modifications

- Use Hydra official Java SDK to create clients dynamically on the Hydra server and save the credential files into the resource directory.
- Remove old OAuth tests based on a real world server. Instead, create another role on OAuth2 server to test `grantPermissionOnNamespace` or `revokePermissionOnNamespace` with specific permissions.
- In addition, inherit the built-in `PulsarAuthorizationProvider` instead of a mocked authorization provider because we configure another authorization provider only to verify KoP works without `superUserRoles` config.
- Add `testWrongSecret` to test a wrong client credential.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  

tests/src/test/java/io/streamnative/pulsar/handlers/kop/HydraHttpClient.java
```diff
@@ -0,0 +1,51 @@
...
+import java.io.Closeable;
+import java.io.IOException;
+import javax.ws.rs.ClientErrorException;
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.MediaType;
+
+public class HydraHttpClient implements Closeable {
```
Demogorgon314, line 25: Can we use hydra SDK to create the client? See: https://www.ory.sh/docs/hydra/sdk

BewareMyPower: Yes. I'll replace the HTTP client with the Hydra Java SDK.

BewareMyPower: Done. PTAL again.

Demogorgon314 (approved):

PR #1256, opened 2022-04-25
[Bugfix] Delete group id stored in metadata storage when consumer close

### LINKED ISSUE #769: [BUG] Group name store in zookeeper for metrics never delete
## Describe the bug

Currently, KoP must store group name in zookeeper when KoP handle fetch request, because in current implementation, to monitor consumer stats must have group name when consumer fetch datas. 

But this data never delete in zookeeper. As time goes by, the amount of data becomes larger and larger. Then may cause cluster crash.

We should find a way to delete this group name, when group is not active.

Demogorgon314:
Fixes #769 

### Motivation

Currently, KoP must store group name in zookeeper when KoP handle fetch request because in the current implementation, to monitor consumer stats must have group name when consumer fetch data.

But this data is never deleted in zookeeper. As time goes by, the amount of data becomes larger and larger. Then may cause the cluster crash.

We should find a way to delete this group name when the group is not active.

### Modifications

Delete the group id path when the consumer close.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -376,7 +376,7 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
                 metadataStore.delete(path, Optional.empty())
                         .whenComplete((__, ex) -> {
                             if (ex != null) {
-                                if (ex instanceof MetadataStoreException.NotFoundException) {
+                                if (ex.getCause() instanceof MetadataStoreException.NotFoundException) {
                                     return;
                                 }
                                 log.error("Delete groupId failed. Path: [{}]", path, ex);
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -377,6 +377,9 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
                         .whenComplete((__, ex) -> {
                             if (ex != null) {
                                 if (ex.getCause() instanceof MetadataStoreException.NotFoundException) {
+                                    if (log.isDebugEnabled()) {
+                                        log.debug("The groupId store path doesn't exist. Path: [{}]", path);
+                                    }
                                     return;
                                 }
                                 log.error("Delete groupId failed. Path: [{}]", path, ex);
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -68,6 +68,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentSkipListSet;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
@@ -204,6 +205,7 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
     private final int failedAuthenticationDelayMs;
     // store the group name for current connected client.
     private final ConcurrentHashMap<String, CompletableFuture<String>> currentConnectedGroup;
+    private final ConcurrentSkipListSet<String> currentConnectedClientId;
     private final String groupIdStoredPath;
 
     @Getter
@@ -318,6 +320,7 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
         this.maxReadEntriesNum = kafkaConfig.getMaxReadEntriesNum();
         this.entryFormatter = EntryFormatterFactory.create(kafkaConfig);
         this.currentConnectedGroup = new ConcurrentHashMap<>();
+        this.currentConnectedClientId = new ConcurrentSkipListSet<>();
         this.groupIdStoredPath = kafkaConfig.getGroupIdZooKeeperPath();
         this.maxPendingBytes = kafkaConfig.getMaxMessagePublishBufferSizeInMB() * 1024L * 1024L;
         this.resumeThresholdPendingBytes = this.maxPendingBytes / 2;
@@ -361,6 +364,25 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
                 log.info("currentConnectedGroup remove {}", clientHost);
                 currentConnectedGroup.remove(clientHost);
             }
+
+            // Try to remove all stored groupID on the metadata store.
+            if (log.isDebugEnabled()) {
+                log.debug("Try to remove all stored groupID on the metadata store. Current connected clientIds: {}",
+                        currentConnectedClientId);
+            }
+            currentConnectedClientId.forEach(clientId -> {
+                String path = groupIdStoredPath + GroupIdUtils.groupIdPathFormat(clientHost, clientId);
+                metadataStore.deleteRecursive(path)
+                        .whenComplete((__, ex) -> {
+                            if (ex != null) {
+                                log.error("Delete groupId failed. Path: [{}]", path, ex);
+                                return;
+                            }
+                            if (log.isDebugEnabled()) {
+                                log.debug("Delete groupId success. Path: [{}]", path);
+                            }
+                        });
+            });
 
             // update alive channel count stat
             RequestStats.ALIVE_CHANNEL_COUNT_INSTANCE.decrementAndGet();
@@ -902,6 +924,7 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
         String groupId = request.coordinatorKey();
         String groupIdPath = GroupIdUtils.groupIdPathFormat(findCoordinator.getClientHost(),
                 findCoordinator.getHeader().clientId());
+        currentConnectedClientId.add(findCoordinator.getHeader().clientId());
 
         // Store group name to metadata store for current client, use to collect consumer metrics.
         storeGroupId(groupId, groupIdPath)
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -172,6 +172,7 @@ import org.apache.pulsar.common.naming.TopicName;
 import org.apache.pulsar.common.schema.KeyValue;
 import org.apache.pulsar.common.util.FutureUtil;
 import org.apache.pulsar.common.util.Murmur3_32Hash;
+import org.apache.pulsar.metadata.api.MetadataStoreException;
 import org.apache.pulsar.metadata.api.extended.MetadataStoreExtended;
 
 /**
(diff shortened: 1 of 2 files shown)


tests/src/test/java/io/streamnative/pulsar/handlers/kop/MetricsProviderTest.java
```diff
@@ -230,36 +234,65 @@ public void testMetricsProvider() throws Exception {
...
+            List<String> children1 = mockZooKeeper.getChildren(conf.getGroupIdZooKeeperPath(), false);
+            Assert.assertEquals(children1.size(), 0);
+        });
     }
 
-    private void tryConsume(final String topic, final String clientId, final String groupId) {
+    private void tryConsume(final String topic,
+                            final String clientId,
+                            final String groupId,
+                            SafeRunnable runBeforeClose) {
```
BewareMyPower, line 284: We should not use `SafeRunnable` because it swallows all exceptions. In addition, since checked in a `Runnable` must be caught, using a `Runnable` here is too complicated because we need a try-catch block.

I think 

```java
    public void testUpdateGroupId() throws Exception {
        // ...
        tryConsume(topic, clientId, group1);
        List<String> children = mockZooKeeper.getChildren(conf.getGroupIdZooKeeperPath(), false);
        Assert.assertEquals(children.size(), 1);
        Assert.assertEquals(children.get(0), "127.0.0.1-" + clientId);
        byte[] data = mockZooKeeper
                .getData(conf.getGroupIdZooKeeperPath() + "/" + children.get(0), false, null);
        Assert.assertEquals(new String(data, StandardCharsets.UTF_8), group1);
```

looks more clear than

```java
    public void testUpdateGroupId() {
        // ...
        tryConsume(topic, clientId, group1, () -> {
            try {
                List<String> children = mockZooKeeper.getChildren(conf.getGroupIdZooKeeperPath(), false);
                Assert.assertEquals(children.size(), 1);
                Assert.assertEquals(children.get(0), "127.0.0.1-" + clientId);
                byte[] data = mockZooKeeper
                        .getData(conf.getGroupIdZooKeeperPath() + "/" + children.get(0), false, null);
                Assert.assertEquals(new String(data, StandardCharsets.UTF_8), group1);
            } catch (Exception ex) {
```

BewareMyPower, line 284: We should not use `SafeRunnable` because it swallows all exceptions. In addition, since checked in a `Runnable` must be caught, using a `Runnable` here is too complicated because we need a try-catch block.

I think 

```java
    public void testUpdateGroupId() throws Exception {
        // ...
        tryConsume(topic, clientId, group1);
        List<String> children = mockZooKeeper.getChildren(conf.getGroupIdZooKeeperPath(), false);
        Assert.assertEquals(children.size(), 1);
        Assert.assertEquals(children.get(0), "127.0.0.1-" + clientId);
        byte[] data = mockZooKeeper
                .getData(conf.getGroupIdZooKeeperPath() + "/" + children.get(0), false, null);
        Assert.assertEquals(new String(data, StandardCharsets.UTF_8), group1);
```

looks more clear than

```java
    public void testUpdateGroupId() {
        // ...
        tryConsume(topic, clientId, group1, () -> {
            try {
                List<String> children = mockZooKeeper.getChildren(conf.getGroupIdZooKeeperPath(), false);
                Assert.assertEquals(children.size(), 1);
                Assert.assertEquals(children.get(0), "127.0.0.1-" + clientId);
                byte[] data = mockZooKeeper
                        .getData(conf.getGroupIdZooKeeperPath() + "/" + children.get(0), false, null);
                Assert.assertEquals(new String(data, StandardCharsets.UTF_8), group1);
            } catch (Exception ex) {
```

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -362,6 +365,25 @@ protected void close() {
...
             }
 
+            // Try to remove all stored groupID on the metadata store.
+            if (log.isDebugEnabled()) {
+                log.debug("Try to remove all stored groupID on the metadata store. Current connected clientIds: {}",
+                        currentConnectedClientId);
+            }
+            currentConnectedClientId.forEach(clientId -> {
+                String path = groupIdStoredPath + GroupIdUtils.groupIdPathFormat(clientHost, clientId);
+                metadataStore.deleteRecursive(path)
```
BewareMyPower, line 375: IMO, it's better to use `delete` here because the path should not contain any child.

BTW, I think it would be better if we could separate all these z-node create/get/delete operations from `KafkaRequestHandler` and `MessageFetchContext` to a single class like `GroupIdStorage`.

BewareMyPower, line 375: IMO, it's better to use `delete` here because the path should not contain any child.

I think it would be better if we could separate all these z-node create/get/delete operations from `KafkaRequestHandler` and `MessageFetchContext` to a single class like `GroupIdStorage`.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -362,6 +365,25 @@ protected void close() {
...
+                        .whenComplete((__, ex) -> {
+                            if (ex != null) {
+                                log.error("Delete groupId failed. Path: [{}]", path, ex);
+                                return;
+                            }
```
BewareMyPower, line 380: We should ignore the `MetadataStoreException$NotFoundException`. Here are some logs when I test this patch in a standalone.

```
2022-04-28T15:06:13,143+0800 [metadata-store-22-1] ERROR io.streamnative.pulsar.handlers.kop.KafkaRequestHandler - Delete groupId failed. Path: [/client_group_id/127.0.0.1-consumer-my-group-3]
java.util.concurrent.CompletionException: org.apache.pulsar.metadata.api.MetadataStoreException$NotFoundException: org.apache.zookeeper.KeeperException$NoNodeException: KeeperErrorCode = NoNode for /client_group_id/127.0.0.1-consumer-my-group-3
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:292) ~[?:1.8.0_332]
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:308) ~[?:1.8.0_332]
	at java.util.concurrent.CompletableFuture.uniRun(CompletableFuture.java:714) ~[?:1.8.0_332]
	at java.util.concurrent.CompletableFuture$UniRun.tryFire(CompletableFuture.java:701) ~[?:1.8.0_332]
	at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:488) ~[?:1.8.0_332]
	at java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:1990) ~[?:1.8.0_332]
	at org.apache.pulsar.metadata.impl.ZKMetadataStore.lambda$null$10(ZKMetadataStore.java:341) ~[org.apache.pulsar-pulsar-metadata-2.10.0.jar:2.10.0]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_332]
	at java.util.concurrent.FutureTask.run(FutureTask.java:266) [?:1.8.0_332]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180) [?:1.8.0_332]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293) [?:1.8.0_332]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:1.8.0_332]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [?:1.8.0_332]
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) [io.netty-netty-common-4.1.74.Final.jar:4.1.74.Final]
	at java.lang.Thread.run(Thread.java:750) [?:1.8.0_332]
Caused by: org.apache.pulsar.metadata.api.MetadataStoreException$NotFoundException: org.apache.zookeeper.KeeperException$NoNodeException: KeeperErrorCode = NoNode for /client_group_id/127.0.0.1-consumer-my-group-3
	at org.apache.pulsar.metadata.impl.ZKMetadataStore.getException(ZKMetadataStore.java:428) ~[org.apache.pulsar-pulsar-metadata-2.10.0.jar:2.10.0]
	... 9 more
Caused by: org.apache.zookeeper.KeeperException$NoNodeException: KeeperErrorCode = NoNode for /client_group_id/127.0.0.1-consumer-my-group-3
	at org.apache.zookeeper.KeeperException.create(KeeperException.java:118) ~[org.apache.zookeeper-zookeeper-3.6.3.jar:3.6.3]
	at org.apache.zookeeper.KeeperException.create(KeeperException.java:54) ~[org.apache.zookeeper-zookeeper-3.6.3.jar:3.6.3]
	at org.apache.pulsar.metadata.impl.ZKMetadataStore.getException(ZKMetadataStore.java:422) ~[org.apache.pulsar-pulsar-metadata-2.10.0.jar:2.10.0]
	... 9 more
```

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
Demogorgon314: Yes, `delete` is better. We can open other PR to refactor to `GroupIdStorage `.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/MetricsProviderTest.java
Demogorgon314: We need to check if the path exists before the client close. Can you explain it in detail?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -362,6 +366,28 @@ protected void close() {
...
+            if (log.isDebugEnabled()) {
+                log.debug("Try to remove all stored groupID on the metadata store. Current connected clientIds: {}",
+                        currentConnectedClientId);
+            }
+            currentConnectedClientId.forEach(clientId -> {
+                String path = groupIdStoredPath + GroupIdUtils.groupIdPathFormat(clientHost, clientId);
+                metadataStore.delete(path, Optional.empty())
+                        .whenComplete((__, ex) -> {
+                            if (ex != null) {
+                                if (ex instanceof MetadataStoreException.NotFoundException) {
```
BewareMyPower, line 379: `ex` is a `CompletionException`, we should check `ex.getCause()` rather than `ex` itself.

BewareMyPower (changes requested):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/MetricsProviderTest.java
BewareMyPower: Sorry I missed this point.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
BewareMyPower: In addition, we can add a debug log when a `NotFoundException`.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -362,6 +366,28 @@ protected void close() {
...
             }
 
+            // Try to remove all stored groupID on the metadata store.
+            if (log.isDebugEnabled()) {
+                log.debug("Try to remove all stored groupID on the metadata store. Current connected clientIds: {}",
+                        currentConnectedClientId);
+            }
+            currentConnectedClientId.forEach(clientId -> {
+                String path = groupIdStoredPath + GroupIdUtils.groupIdPathFormat(clientHost, clientId);
+                metadataStore.delete(path, Optional.empty())
```
eolivelli, line 376: so when we close a client connection we do the deletion.
I am +1 to this as a work around.

We should this store data somewhere else, like we are doing now with the ProducerId

I don't have time these days, but I will be happy to work on a alternative solution in the mid term

eolivelli (approved):
LGTM

BewareMyPower (approved):

PR #1268, opened 2022-05-05
[transaction] Support write transaction marker when using oauth2 authentication

Demogorgon314:
### Motivation

Currently, when we use OAuth2 authentication and use the transaction, the broker writes transaction marker will fail, the error log sees below: 
```
io.streamnative.pulsar.handlers.kop.KafkaCommandDecoder - Failed authentication with [/10.52.8.87:45106] (SASL mechanism 'PLAIN' requested by client is not supported)
```
We need to support OAuth2 authentication when writing transaction markers.

### Modifications

* Support write transaction marker when using oauth2 authentication.
* Add new TransactionWithAuthTest units test to cover this case.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionMarkerChannelHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionMarkerChannelHandler.java
@@ -322,8 +322,11 @@ public class TransactionMarkerChannelHandler extends ChannelInboundHandlerAdapte
                         + "\u0000" + authUsername
                         + "\u0000" + authPassword;
                 saslAuthBytes = usernamePassword.getBytes(UTF_8);
-            } else {
+            } else if (authentication instanceof AuthenticationOAuth2) {
                 saslAuthBytes = new OAuthBearerClientInitialResponse(commandData).toBytes();
+            } else {
+                log.error("Unknown authentication : {}", authentication);
+                saslAuthBytes = new byte[0];
             }
             SaslAuthenticateRequest request = new SaslAuthenticateRequest
                     .Builder(ByteBuffer.wrap(saslAuthBytes))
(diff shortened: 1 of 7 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionMarkerChannelHandler.java
```diff
@@ -286,33 +301,52 @@ private CompletableFuture<ChannelHandlerContext> authenticate(final ChannelHandl
...
+                String usernamePassword = prefix
+                        + "\u0000" + authUsername
+                        + "\u0000" + authPassword;
+                saslAuthBytes = usernamePassword.getBytes(UTF_8);
+            } else if (authentication instanceof AuthenticationOAuth2) {
+                saslAuthBytes = new OAuthBearerClientInitialResponse(commandData).toBytes();
+            } else {
+                log.error("Unknown authentication : {}", authentication);
+                saslAuthBytes = new byte[0];
+            }
```
BewareMyPower, line 330:
```java
            switch (mechanism) {
                case PlainSaslServer.PLAIN_MECHANISM:
                    // TODO:
                    break;
                case OAuthBearerLoginModule.OAUTHBEARER_MECHANISM:
                    // TODO:
                    break;
                default:
                    break;
            }
```

It's better to use the mechanism string to differ how to handle the `commandData` and other information (username and password).

tests/src/test/java/io/streamnative/pulsar/handlers/kop/TransactionWithAuthTest.java
```diff
@@ -0,0 +1,157 @@
...
+import org.apache.pulsar.client.impl.auth.oauth2.AuthenticationFactoryOAuth2;
+import org.apache.pulsar.client.impl.auth.oauth2.AuthenticationOAuth2;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import sh.ory.hydra.ApiException;
+import sh.ory.hydra.api.AdminApi;
+import sh.ory.hydra.model.OAuth2Client;
+
+@Slf4j
+public class TransactionWithAuthTest extends TransactionTest {
```
BewareMyPower, line 38: This test can be renamed with `TransactionWithOAuthBearerAuthTest`.

But I think a better test is configuring both `plain` and `oauthbearer` mechanisms so that we can verify transaction could work for all existing supported mechanisms.

Demogorgon314: Yes, in the current code, we have one test already covering `PLATN` mechanisms, see `SaslPlainTestBase#transactionsReadCommittedTest`.

BewareMyPower (approved):

Issue #1272, opened 2022-05-06
[BUG] Strange write frequency to the offset topic


BewareMyPower:
**Describe the bug**

Following screenshots all represent the publish rate to the offset topic (`__consumer_offsets`), which represents the frequency to commit the offsets.

<img width="854" alt="image" src="https://user-images.githubusercontent.com/18204803/167169903-9fd32697-ef1a-4a60-871e-9045816e71f1.png">

We can see there was a wave nearly every 6 hours.

----

<img width="834" alt="image" src="https://user-images.githubusercontent.com/18204803/167170285-2f196e9c-41c6-496f-b5db-ec63efaa86bb.png">

Usually, the publish rate is stable.

----

<img width="824" alt="image" src="https://user-images.githubusercontent.com/18204803/167171012-cf7dcd81-0753-4664-9e9c-1b70730622cb.png">

During the wave, the publish rate increased with a constant speed.

----

<img width="866" alt="image" src="https://user-images.githubusercontent.com/18204803/167171269-0c6444c8-a4b1-416e-8948-9e45702bc04e.png">

Each local wave range is about **2 minutes.**

**To Reproduce**

The same consumer subscribes the topic with the same group name, call `commitSync` method for each message. Consumers are closed and reopened **every 2 minutes**.

**Expected behavior**

We need to figure out whether the phenomenon is reasonable.

**Analysis**

It might due to the **topic compaction**. Currently, KoP only sets the compaction size with 100MB. But the compaction time is not configured.

```java
@FieldContext(
    category = CATEGORY_SERVER,
    doc = "Interval between checks to see if topics with compaction policies need to be compacted"
)
private int brokerServiceCompactionMonitorIntervalInSeconds = 60;
```

BewareMyPower:
The bug is that the metadata namespace's compaction threshold is never set after #448. Then every 60 seconds (`brokerServiceCompactionMonitorIntervalInSeconds`) the broker will perform compaction on this namespace.

BewareMyPower:
With the previous implementation, the compaction is disabled when the compaction threshold is not set. So we need to investigate more about this issue.

PR #1273, opened 2022-05-06
Fix compaction threshold not configured in metadata namespace

BewareMyPower:
Fixes https://github.com/streamnative/kop/issues/1272

### Motivation

https://github.com/streamnative/kop/pull/448 introduced a bug that the
compaction threshold was not set in metadata namespace.

```java
// NOTE: compactionThreshold is always null unless it has been configured
Long compactionThreshold = namespaces.getCompactionThreshold(kafkaMetadataNamespace);
if (compactionThreshold != null && compactionThreshold != MAX_COMPACTION_THRESHOLD) {
    // it never gets here unless compaction threshold has been configured
    namespaces.setCompactionThreshold(kafkaMetadataNamespace, MAX_COMPACTION_THRESHOLD);
}
```

In addition, currently there is no test that ensures the metadata has
been initialized correctly.

### Modifications

- Configure the compaction threshold when the metadata namespace is
  created. Since the policies could only be configured when the
  namespace didn't exist before, there is no need to check the existing
  policy.
- Add `MetadataInitTest` to test the metadata namespace has been
  configured correctly. It also verifies the
  `kafkaManageSystemNamespaces` config, which disables the
  initialization of the metadata namespace.
- Refactor the `internalSetup` method to avoid the redundant start of
  broker.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MetadataUtils.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MetadataUtils.java
@@ -204,28 +204,12 @@ public class MetadataUtils {
 
             // set namespace config only when offset metadata namespace first create
             if (isMetadataNamespace) {
-                int retentionMinutes = (int) conf.getOffsetsRetentionMinutes();
-                int retentionSizeInMB = conf.getSystemTopicRetentionSizeInMB();
-                RetentionPolicies retentionPolicies = namespaces.getRetention(kafkaNamespace);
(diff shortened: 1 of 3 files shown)


BewareMyPower:
This PR relies on the changes of #1274.

wenbingshen (approved):
LGTM.
Good find! :)

PR #1274, opened 2022-05-07
[Tests] Configure a different metadata namespace with the default namespace

BewareMyPower:
### Motivation

Currently in KoP tests, the metadata namespace configured by
`kafkaMetadataNamespace` is "default", which is same with the default
namespace of Pulsar and KoP (`kafkaNamespace`).

https://github.com/streamnative/kop/blob/64001937c92d80ed4f652becd8d6b9768918b875/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KopProtocolHandlerTestBase.java#L176-L177

It's a terrible config because some topic policies might be changed with
the metadata namespace and the offset topic is created during starting.
We should avoid the namespace for normal topics being affected by these
implicit operations.

### Modifications

Remove the configs for `kafkaMetadataTenant` and
`kafkaMetadataNamespace`, then fix the following tests:

- `MultiLedgerTest`: set the infinite retention to avoid rollover
  ledgers being removed automatically.
- `KafkaTopicConsumerManagerTest`: trigger the topic lookup before
  creating `PersistentTopic` objects in `BrokerService` so that broker
  can acquire the ownership of namespace bundles.
- `CacheInvalidatorTest`: unload the metadata namespace instead of the
  default namespace.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KopProtocolHandlerTestBase.java
+++ b/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KopProtocolHandlerTestBase.java
@@ -113,9 +113,6 @@ public abstract class KopProtocolHandlerTestBase {
     protected NonClosableMockBookKeeper mockBookKeeper;
     protected final String configClusterName = "test";
 
-    protected final String tenant = "public";
-    protected final String namespace = "default";
-
     private SameThreadOrderedSafeExecutor sameThreadOrderedSafeExecutor;
     private OrderedExecutor bkExecutor;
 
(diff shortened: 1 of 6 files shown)


Demogorgon314 (approved):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KopProtocolHandlerTestBase.java
```diff
@@ -173,9 +170,6 @@ protected void resetConfig() {
         kafkaConfig.setForceDeleteTenantAllowed(true);
         kafkaConfig.setForceDeleteNamespaceAllowed(true);
 
-        kafkaConfig.setKafkaMetadataTenant(tenant);
-        kafkaConfig.setKafkaMetadataNamespace(namespace);
-
```
wenbingshen, line 178: Good work.

PR #1276, opened 2022-05-09
Improve handle OAuth callback throw authentication exception behavior

Demogorgon314:
### Motivation

Currently, when the `OauthValidatorCallbackHandler#handleCallback` method throws an exception, the client will receive warn log below and it will always try to reconnect.
```
15:51:14.143 [kafka-consumer-2] WARN org.apache.kafka.clients.NetworkClient - [Consumer clientId=consumer-topic-1-1, groupId=topic-1] Connection to node 782843879 (localhost/127.0.0.1:9092) terminated during authentication. This may happen due to any of the following reasons: (1) Authentication failed due to invalid credentials with brokers older than 1.0.0, (2) Firewall blocking Kafka TLS traffic (eg it may only allow HTTPS traffic), (3) Transient network issue.
```

Broker side log:
```
2022-05-09T14:29:39,673+0800 [pulsar-io-18-10] ERROR io.streamnative.pulsar.handlers.kop.KafkaRequestHandler - Caught error in handler, closing channel
java.lang.IllegalStateException: Authentication exchange has not completed
	at io.streamnative.pulsar.handlers.kop.security.oauth.KopOAuthBearerSaslServer.getAuthorizationID(KopOAuthBearerSaslServer.java:94) ~[?:?]
	at io.streamnative.pulsar.handlers.kop.security.SaslAuthenticator.handleSaslToken(SaslAuthenticator.java:503) ~[?:?]
	at io.streamnative.pulsar.handlers.kop.security.SaslAuthenticator.authenticate(SaslAuthenticator.java:223) ~[?:?]
	at io.streamnative.pulsar.handlers.kop.KafkaRequestHandler.channelPrepare(KafkaRequestHandler.java:382) ~[?:?]
// ...
```

In this case, we should have `SaslAuthenticationException` instead of the warning log.

### Modifications

* Make sure to send the authentication error response to the client before closing the channel.
* Add check to make sure the `SaslServer` is complete before getting the authorization ID.
* Add units test to cover this case.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
@@ -448,7 +448,7 @@ public class SaslAuthenticator {
                 byte[] response = saslServer.evaluateResponse(clientToken);
                 if (response != null) {
                     ByteBuf byteBuf = sizePrefixed(ByteBuffer.wrap(response));
-                    Session newSession = null;
+                    final Session newSession;
                     if (saslServer.isComplete()) {
                         newSession = new Session(
                                 new KafkaPrincipal(KafkaPrincipal.USER_TYPE, saslServer.getAuthorizationID(),
@@ -458,14 +458,15 @@ public class SaslAuthenticator {
                         if (!tenantAccessValidationFunction.apply(newSession)) {
                             throw new AuthenticationException("User is not allowed to access this tenant");
                         }
+                    } else {
+                        newSession = null;
                     }
-                    final Session finalNewSession = newSession;
                     ctx.channel().writeAndFlush(byteBuf).addListener(future -> {
                         if (!future.isSuccess()) {
                             log.error("[{}] Failed to write {}", ctx.channel(), future.cause());
                         } else {
                             // This session is required for authorization.
-                            session = finalNewSession;
+                            session = newSession;
                             if (log.isDebugEnabled()) {
                                 log.debug("Send sasl response to SASL_HANDSHAKE v0 old client {} successfully, "
                                         + "session {}", ctx.channel(), session);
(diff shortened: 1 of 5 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
```diff
@@ -444,21 +447,25 @@ private void handleSaslToken(ChannelHandlerContext ctx,
...
+                        newSession = new Session(
+                                new KafkaPrincipal(KafkaPrincipal.USER_TYPE, saslServer.getAuthorizationID(),
+                                        safeGetProperty(saslServer, USER_NAME_PROP),
+                                        safeGetProperty(saslServer, AUTH_DATA_SOURCE_PROP)),
+                                "old-clientId");
+                        if (!tenantAccessValidationFunction.apply(newSession)) {
+                            throw new AuthenticationException("User is not allowed to access this tenant");
+                        }
+                    }
+                    final Session finalNewSession = newSession;
```
BewareMyPower, line 462: We can keep using a `final` variable here.

```java
                    final Session newSession;
                    if (saslServer.isComplete()) {
                        newSession = /* ... */;
                    } else {
                        newSession = null;
                    }
```

BewareMyPower, line 462: We can keep using a `final` variable here.

```java
                    final Session newSession;
                    if (saslServer.isComplete()) {
                        newSession = /* ... */;
                    } else {
                        newSession = null;
                    }
```

BewareMyPower (changes requested):
We'd better catch the exception thrown from `KopOAuthBearerSaslServer#getAuthorizationID` instead of checking `isComplete` before calling `getAuthorizationID`. Maybe we can throw a `SaslAuthentication` directly instead of an `IllegalStateException`.

In addition, I found `PlainSaslServer#getAuthorizationID` doesn't check `complete` field. We should make them consistent rather than check `isComplete` outside these methods.

BewareMyPower (approved):

BewareMyPower:
It's right that we should check `isComplete` outside the `SaslServer#getAuthorizationID` method. See

```java
    /**
     * Reports the authorization ID in effect for the client of this
     * session.
     * This method can only be called if isComplete() returns true.
     * @return The authorization ID of the client.
     * @exception IllegalStateException if this authentication session has not completed
     */
    public String getAuthorizationID();
```

Issue #1281, opened 2022-05-12
[BUG] OMB Draining Backlog test does not complete in case of topic unloads


eolivelli:
**Describe the bug**
Running OMB @dave2wave reported that the test does not complete.
The test is about using the "drain backlog" feature.
Using the default Kafka Driver shipped with OMB

**To Reproduce**
run OMB using this workload, on Pulsar Standalone, with Pulsar 2.10 and KOP

```
name: test

topics: 1
partitionsPerTopic: 1
messageSize: 100
#payloadFile: "payload/payload-100b.data"
useRandomizedPayloads: true
randomBytesRatio: 0.5
randomizedPayloadPoolSize: 1000

subscriptionsPerTopic: 1
consumerPerSubscription: 1
producersPerTopic: 1

# Discover max-sustainable rate
producerRate: 1000000
acceptableBacklog: 50000
producerPauseDuringDrain: 1
consumerBacklogSizeGB: 5
testDurationMinutes: 15
```

during the drain backlog phase unload all the partitions:
```
for i in $(pulsar-admin topics list public/default)
pulsar-admin topics unload $i
```

The Consumers are no more able to consume and the backlog is never fully drained 

**Expected behavior**
The test completes

**Additional context**

eolivelli:
I am working on this problem

eolivelli:
The bug is about topic unloading during a Fetch, I am still investigating

eolivelli:
the problem was due to a wrong configuration about retention.
So topic unloading triggers ledger trimming and you lose your data.

Closing this issue now

PR #1282, opened 2022-05-12
Clean up errors sent to the client during Topic Unload

eolivelli:
### Motivation

While investigating on #1281 I have found a couple of enhancements:
- we should add more logging to troubleshoot unknown errors during topic unload
- we can improve some logging
- in case of CursorAlreadyClosedException  we return UNKNOWN_SERVER_ERROR and NOT_LEADER_FOR_PARTITION and this generates a "received an unknown error" log line on the client, that is pretty scary

### Modifications

- improve the toString representation of the callback passed to asyncFindPosition
- return NOT_LEADER_FOR_PARTITION in case of CursorAlreadyClosedException and ManagedLedgerFencedException
- reduce log level of some parts of the code

### Verifying this change

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MessageMetadataUtils.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MessageMetadataUtils.java
@@ -13,10 +13,12 @@
  */
 package io.streamnative.pulsar.handlers.kop.utils;
 
+import com.google.common.base.Predicate;
 import io.netty.buffer.ByteBuf;
 import io.streamnative.pulsar.handlers.kop.exceptions.MetadataCorruptedException;
 import java.util.concurrent.CompletableFuture;
 import javax.annotation.Nullable;
+import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.bookkeeper.mledger.AsyncCallbacks;
 import org.apache.bookkeeper.mledger.Entry;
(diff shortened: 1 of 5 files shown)


BewareMyPower:
Could you give the correct issue number? I see you still mentioned this PR after the change.

eolivelli:
> Could you give the correct issue number? I see you still mentioned this PR after the change.

sorry, it was #1281. I closed the issue, it was a configuration problem. Btw I added these useful little enhancements that helped me debugging the problem and also we removed some scary logs on the client

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/MessageMetadataUtils.java
```diff
@@ -191,6 +204,12 @@ public static CompletableFuture<Position> asyncFindPosition(final ManagedLedger
             } finally {
                 entry.release();
             }
-        });
+        }
+
+        @Override
+        public String toString() {
+            return "FindEntryByOffset{ " + offset + "}";
+        }
```
BewareMyPower, line 212: Good find! I think we can add the info of the managed ledger like

```java
            return "FindEntryByOffset{ managed ledger: " + managedLedger.getName() + ", offset: " + offset + " }";
```

And we can only reference the name when constructing the predicate.

```java
        return managedLedger.asyncFindPosition(new FindEntryByOffset(managedLedger.getName(),
                offset, skipMessagesWithoutIndex));
    }

    @AllArgsConstructor
    private static class FindEntryByOffset implements Predicate<Entry> {

        private final String managedLedgerName;
```

eolivelli: IIRC the topic name is already reported in the logs of asyncFindPosition:

https://github.com/apache/pulsar/blob/master/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L1710

BewareMyPower (approved):

PR #1305, opened 2022-05-25
Add OAuth token refresh test

### LINKED ISSUE #1261: [FEATURE] Verify OauthLoginCallbackHandler works when token expired

#1260 implements `lifetimeMs` and `startTimeMs` methods. They are claimed "correctly" but not verified. We need to set up a test to simulate a quickly expired token and verify Kafka clients with the `oauth-client` plugin can refresh the token. Otherwise, we should think of a way to process the token expiration.

Configure a shorter timeout like 10s in Hydra, and see what will happen.

ZiyaoWei:
Fixes #1261.

### Motivation

Adds a test to make sure when OAuth tokens expire the Kafka client will automatically refresh the token.

### Modifications

Change the Hydra token expiration to 5 seconds, and add a test that sleeps for 5 seconds after getting OAuth tokens, and make sure the Kafka clients still works which means the tokens are refreshed automatically.

### Verifying this change

- [ ] Make sure that the change passes the CI checks.=

  - Run the test added which should pass;
  - If we change the OAuthBearerTokenImpl class so it returns a lifetimMs that's very far in the future, token refresh is in effect disabled and the test would then fail.

### Documentation

Need to update docs? 

- [x] `no-need-doc` 
  
  Only adding a test.

### DIFF
--- a/tests/src/test/java/io/streamnative/pulsar/handlers/kop/SaslOauthKopHandlersTest.java
+++ b/tests/src/test/java/io/streamnative/pulsar/handlers/kop/SaslOauthKopHandlersTest.java
@@ -16,7 +16,7 @@ package io.streamnative.pulsar.handlers.kop;
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.when;
-
+import static org.testng.Assert.assertEquals;
 import com.google.common.collect.Sets;
 import io.streamnative.pulsar.handlers.kop.security.oauth.OauthLoginCallbackHandler;
 import io.streamnative.pulsar.handlers.kop.security.oauth.OauthValidatorCallbackHandler;
@@ -34,10 +34,13 @@ import java.util.concurrent.ExecutionException;
 import javax.naming.AuthenticationException;
 import javax.security.auth.login.LoginException;
 import lombok.Cleanup;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.clients.producer.RecordMetadata;
 import org.apache.kafka.common.errors.SaslAuthenticationException;
 import org.apache.kafka.common.errors.TopicAuthorizationException;
 import org.apache.pulsar.broker.ServiceConfiguration;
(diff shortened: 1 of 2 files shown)


BewareMyPower:
Could you sign the license and rebase to master?

tests/src/test/java/io/streamnative/pulsar/handlers/kop/SaslOauthKopHandlersTest.java
```diff
@@ -16,7 +16,7 @@
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.when;
-
+import static org.testng.Assert.assertEquals;
```
BewareMyPower, line 19:
```suggestion
import static org.testng.Assert.assertEquals;

```

ZiyaoWei:
Thanks @BewareMyPower, will try that instead!

Also nit: maybe we should standardize how we spell OAuth and change all the `Oauth` usages to `OAuth` instead?

BewareMyPower:
> maybe we should standardize how we spell OAuth and change all the Oauth usages to OAuth instead?

Good suggestion. In Kafka, it uses `OAuth`, e.g. `OAuthBearerUnsecuredJws`.

But we should keep `OauthLoginCallbackHandler` and `OauthValidatorCallbackHandler` not changed for backward compatibility. Because at KoP side we need to configure `OauthValidatorCallbackHandler` as the `kopOauth2AuthenticateCallbackHandler` and at Kafka client side we need to configure `OauthLoginCallbackHandler` as the "sasl.login.callback.handler.class" property.

For the same reason, we should keep the KoP configs with `kopOauth2` prefixes not changed.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/SaslOauthKopHandlersTest.java
```diff
@@ -122,6 +134,33 @@ public void testSimpleProduceConsume() throws Exception {
...
+
+        // Check that new tokens have been generated
+        assertTrue(CustomOauthLoginCallbackHandler.tokens.size() > 1);
+
+        producer.send(new ProducerRecord<>(topic, message)).get();
+        RecordMetadata metadata = producer.send(new ProducerRecord<>(topic, message)).get();
+
+        log.info("Send to {}-partition-{}@{}", metadata.topic(), metadata.partition(), metadata.offset());
+        log.error("tokens: " + CustomOauthLoginCallbackHandler.tokens.stream().map(Object::toString)
+                .collect(Collectors.joining()));
```
BewareMyPower, line 161: Why do you print error logs here? I think we should verify the size of `tokens` is 2 and `tokens[1]` is different from `tokens[0]`.

BewareMyPower (changes requested):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/SaslOauthKopHandlersTest.java
ZiyaoWei: Oops, forgot to remove debug logs - I didn't verify it's 2 so that if the test environment somehow gets very slow and the token refreshed more than once there won't be a false positive (checking there's only one token in the beginning seems less risky), but I guess we can start with 2 and see how it goes.

BewareMyPower (approved):

Issue #1293, opened 2022-05-23
[FEATURE] Restore the support for Kafka clients 0.9 and 0.10


BewareMyPower:
**Is your feature request related to a problem? Please describe.**
#1060 updates the `kafka-clients` dependency to 2.1.1 but it drops the support for Kafka clients 0.9 and 0.10 because `ListOffsetRequest.offsetData()` method has removed.

**Describe the solution you'd like**

PR #1060, opened 2022-02-09
Update Kafka Client to 2.1.1 and Disable Kafka 0.9 and 0.10 tests

eolivelli:
This patch upgrade the Kafka Client dependency to 2.1.1.
The reason is that Kafka 2.0.0 dependency has some CVEs reported and we cannot ship this project to users if it contains CVEs.

The downside is that we are dropping compatibility with 0.9 and 0.10 clients because version 0 of ListOffSets is no more supported.

PR #1307, opened 2022-05-26
Restore the support for Kafka clients 0.9 and 0.10

Demogorgon314:
Fixes: #1293

### Motivation

#1060 updates the `kafka-clients` dependency to 2.1.1 but it drops the support for Kafka clients 0.9 and 0.10 because `ListOffsetRequest.offsetData()` method has removed.

### Modifications

Add the old version of `ListOffsetRequest` to be compatible with Kafka clients 0.9 and 0.10.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  

eolivelli:
IIRC the main problem that was solved by https://github.com/streamnative/kop/pull/1060 is that older Kafka Clients bring well known CVEs inside the KOP package.
For that reason we had to drop it, otherwise KOP is not usable by enterprises that require 0 CVEs to run something in production.

This PR defeats the purpose of 1060.
I am not sure we can do this.

@nicoloboschi can you please confirm ?

eolivelli:
an alternative approach is to copy/paste (or write from scratch) only the classes the we need

Demogorgon314:
@eolivelli Sure,  I'll try to copy the necessary classes.

eolivelli (approved):
LGTM

great work

BewareMyPower (approved):

woshihaoren:
@BewareMyPower why revert this pr ?
kop 2.10.6 can not work on 0.10.

BewareMyPower:
@woshihaoren Both this PR and https://github.com/streamnative/kop/pull/1060 are reverted. https://github.com/streamnative/kop/pull/1060 breaks the compatibility with 0.9 and 1.0 clients and this PR restores the compatibility. Since both them are reverted, 2.10.1.6 should still be compatible with 0.9 and 1.0 clients.

Could you open an issue to verify KoP 2.10.1.6 cannot work for Kafka 0.10 client?

PR #1313, opened 2022-05-29
Wrap the sasl response token body with the correct method

wenbingshen:
### Motivation

I found this problem by re-reading this part of the code today while preparing the KoP meetup ppt.

PR #697 In order to solve the problem described in issue #696, the method used is not accurate enough. If `SaslServer` is extended in the future, supporting V0 authentication will return an incorrect response.

`sizePrefixed` originally wanted to increase the message length before the message body, because KoP uses `LengthFieldPrepender` to extend the message length field before the message body, so here we can send the message body directly.

### Documentation
- [x] `no-need-doc`


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
```diff
@@ -420,15 +420,9 @@ private static ByteBuf buildKafkaResponse(RequestHeader header,
 
     @VisibleForTesting
     public static ByteBuf sizePrefixed(ByteBuffer buffer) {
```
Demogorgon314, line 422: I think we can remove this method and replace the usage with `Unpooled.wrappedBuffer(response)`.

BewareMyPower: +1

eolivelli (commented):
is this method used anywhere ?
we are changing the behaviour but integration tests are still passing.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
wenbingshen: Yes, this method can be removed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
```diff
@@ -447,7 +433,7 @@ private void handleSaslToken(ChannelHandlerContext ctx,
                 nioBuffer.get(clientToken, 0, clientToken.length);
                 byte[] response = saslServer.evaluateResponse(clientToken);
                 if (response != null) {
-                    ByteBuf byteBuf = sizePrefixed(ByteBuffer.wrap(response));
+                    ByteBuf byteBuf = Unpooled.wrappedBuffer(ByteBuffer.wrap(response));
```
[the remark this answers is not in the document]
Demogorgon314: Yes we can directly use `Unpooled.wrappedBuffer(response);`.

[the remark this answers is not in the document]
wenbingshen: We might be able to directly use `Unpooled.wrappedBuffer(response);` or `ctx.channel().writeAndFlush(response)`?

> btw, now we are changing the data that we are sending on the write.
why aren't test failing ?

Why does the test fail? I think the expression of the response may have changed, but its value will not change.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/SaslAuthenticator.java
wenbingshen: I have addressed your comments. PTAL

Demogorgon314 (approved):

eolivelli (approved):
Now I understand that previously the method didn't do anything, because we were calling "resetWriterIndex"

BewareMyPower (approved):

PR #1324, opened 2022-06-06
Add support for TLS only brokers - new feature kopTlsEnabledWithBroker

eolivelli:
Modifications:
- add new flag `kopTlsEnabledWithBroker` (named after tlsEnabledWithBroker in Pulsar proxy) to tell KOP to always use TLS while connecting to KOP brokers
- this applies to Transactions

Please note that using the broker advertised listeners and connect to the TLS endpoint is not doable:
- This is not doable on the Proxy because in the proxy cannot do discovery without ZooKeeper
- This won't work well on KOP because a broker may have multiple listeners and a flag to require to connect only by using TLS is needed in any case

### DIFF
--- a/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaSSLChannelTest.java
+++ b/tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaSSLChannelTest.java
@@ -270,6 +270,12 @@ public class KafkaSSLChannelTest extends KopProtocolHandlerTestBase {
             Thread.sleep(100);
         }
 
+        consumeTxData(kafkaServer, topicName, isolation,
+                totalTxnCount * messageCountPerTxn, lastMessage);
+    }
+
+    private void consumeTxData(String kafkaServer, String topicName, String isolation,
+                               int totalMessageCount, String lastMessage) throws Exception {
         Properties consumerProps = new Properties();
         consumerProps.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, kafkaServer);
         consumerProps.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, IntegerDeserializer.class.getName());
@@ -286,8 +292,6 @@ public class KafkaSSLChannelTest extends KopProtocolHandlerTestBase {
 
         // default is https, here need to set empty.
         consumerProps.put("ssl.endpoint.identification.algorithm", "");
-
-        final int totalMessageCount = totalTxnCount * messageCountPerTxn;
 
         @Cleanup
         KafkaConsumer<Integer, String> consumer = new KafkaConsumer<>(consumerProps);


BewareMyPower:
Could you fix the Codacy check?

eolivelli:
> Could you fix the Codacy check?

@BewareMyPower done

eolivelli:
have a few other changes related to TLS.
In particular I have some code to bootstrap TLS using the same Pulsar Broker TLS certificate, that is very convenient because this way you don't have to add special configurations for TLS for KOP and we are going to support the same features as Pulsar broker regarding TLS.

I also have the code to enable TLS in the SchemaRegistry

BewareMyPower:
Good. I'll review it soon.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/utils/ssl/SSLUtils.java
```diff
@@ -257,4 +257,83 @@ public static SSLEngine createSslEngine(SslContextFactory.Server sslContextFacto
...
 
+    public static SSLEngine createClientSslEngine(SslContextFactory.Client sslContextFactory) throws Exception {
+        sslContextFactory.start();
+        SSLEngine engine  = sslContextFactory.newSSLEngine();
+        engine.setUseClientMode(true);
+
+        return engine;
+    }
+
+    public static SslContextFactory.Client createClientSslContextFactory(
```
BewareMyPower, line 268: Could you reuse the code in `createSslContextFactory` instead of copying them?

eolivelli: this method is returning SslContextFactory.Client and not SslContextFactory.Server 
also there are entries that should not be set like KopSslKeystoreLocation (that would mean that we use the certificate of the KOP service for client auth).
because here we are building a "client".

There is way to not copy/paste

BewareMyPower: But they both use the same way to generate an `ImmutableMap`. I mean:

```java
    private static ImmutableMap<String, Object> buildSslConfigMap(KafkaServiceConfiguration kafkaServiceConfiguration) {
        /* ... */
    }
    public static SslContextFactory.Server createSslContextFactory(
            KafkaServiceConfiguration kafkaServiceConfiguration) {
        return createSslContextFactory(buildSslConfigMap(kafkaServiceConfiguration));
    }

    public static SslContextFactory.Client createClientSslContextFactory(
        return createClientSslContextFactory(buildSslConfigMap(kafkaServiceConfiguration));
    }
```

eolivelli:
TransactionMarkerChannelHandler is a "client" that connects to another broker using SSL endpoint in this case.
This is the only case of broker-to-broker communication in KOP using the Kafka protocol.
The new test covers this case.

Incoming TLS connections are already handled by KOP and TransactionMarkerChannelHandler(client) connects to the KOP endpoint.

This is why we have to set the Truststore for TransactionMarkerChannelHandler otherwise we cannot verify the certificate of the KOP broker.

enableTls  in TransactionMarkerChannelHandler will only always mean "connect using TLS", TransactionMarkerChannelHandler will never be a server

BewareMyPower:
Got it. Thanks.

BewareMyPower (approved):

PR #1391, opened 2022-07-07
Prevent incomplete future and add more logs for ProducerIdManager

BewareMyPower:
### Motivation

Currently, if the `ProducerIdManager` failed to generate the producer ID,
the future of epoch and `TxnMetadata` would never complete, see

https://github.com/streamnative/kop/blob/f63b191e72d2319f67f7a5ab7242e898c10ed9c5/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java#L248-L253

`epochAndTxnMetaFuture` would not complete in the case above.

In addition, since `epochAndTxnMetaFuture` never complete exceptionally,
the following code is redundant:

https://github.com/streamnative/kop/blob/f63b191e72d2319f67f7a5ab7242e898c10ed9c5/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java#L304-L309

### Modifications

When the `ProducerIdManager` failed to generate the producer ID,
complete `epochAndTxnMetaFuture` with null and handle the null value in
the callback.

Then add some debug and error logs for `ProducerIdManager`. Regarding
the `exceptionally` block of `epochAndTxnMetaFuture`, treat it as an
unexpected error caused by `thenAccept` block and return a
`UNKNOWN_SERVER_ERROR` as well. Because in this case
`Errors.forException(ex.getCause())` never return a valid error code.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java
@@ -221,8 +221,12 @@ public class TransactionCoordinator {
             producerIdManager.generateProducerId().whenComplete((pid, throwable) -> {
                 short producerEpoch = 0;
                 if (throwable != null) {
+                    log.error("Failed to generate producer id for idempotent producer", throwable);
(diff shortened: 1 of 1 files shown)


Demogorgon314 (approved):
LGTM.

PR #1392, opened 2022-07-08
Make ProducerIdManager configurable

BewareMyPower:
### Motivation

Currently KoP only supports two built-in implementations of
`ProducerIdManager` via the `kafkaTransactionProducerIdsStoredOnPulsar`
config. We can make it configurable to leverage other storage system for
the generation of producer ID.

### Modifications

- Add `kopProducerIdManagerClassName` config to specify the class name
  of customized `ProducerIdManager` implementation. If it's specified,
  the `kafkaTransactionProducerIdsStoredOnPulsar` will be ignored.
- Add `CustomProducerIdManagerTest` to verify the config works.

### Documentation

Check the box below.

Need to update docs? 

- [x] `doc-required` 
  
- [ ] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java
@@ -37,6 +37,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
+import java.util.function.Supplier;
 import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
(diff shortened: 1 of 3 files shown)


BewareMyPower:
We need a proposal for this. Mark it as draft.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java
```diff
@@ -128,12 +129,31 @@ public static TransactionCoordinator of(String tenant,
...
-            producerIdManager = new PulsarStorageProducerIdManagerImpl(
-                    transactionConfig.getTransactionProducerIdTopicName(), txnTopicClient.getPulsarClient());
+        final String producerIdManagerClassName = kafkaConfig.getKopProducerIdManagerClassName();
+        if (producerIdManagerClassName == null) {
+            producerIdManager = producerIdManagerSupplier.get();
         } else {
-            producerIdManager = new ProducerIdManagerImpl(transactionConfig.getBrokerId(), metadataStore);
+            try {
+                producerIdManager = (ProducerIdManager)
+                        Class.forName(producerIdManagerClassName).getDeclaredConstructor().newInstance();
```
eolivelli, line 147: I think that at least we have to pass the KafkaServiceConfiguration object,

the problem is that we need access to some env object, like the PulsarClient

BewareMyPower: I think we need to modify the `ProducerIdManager` interface as well. Ideally I think the `initialize` method should be:

```java
    // If we use `KafkaServiceConfiguration`, maybe we'd better put it in another module like `kop-common`
    CompletableFuture<Void> initialize(Properties props);
```

I'm not sure if it would be proper to pass a `BrokerService` object because the `PulsarClient` or `MetadataStore` object that can be retrieved from it. My initial purpose is to separate these objects from the dependency on `pulsar-broker` dependency, i.e. if you want to use ZK/Etcd/Pulsar as the pid storage, you must create the corresponding client according to your own client configs.

BTW, I think it should be carefully considered and it's not an urgent task so I mark it as drafted and might continue in future.

PR #1393, opened 2022-07-09
Fix linting errors in io.streamnative.pulsar.handlers.kop.coordinator.group 

ZiyaoWei:
### Motivation

Reading through the code in `io.streamnative.pulsar.handlers.kop.coordinator.group` and there are a lot of linting errors. These are annoying and could mask potential bugs, e.g., #1394.

### Modifications

Fix the linter warnings. Most changes should be trivial cleanup, but there is one real bug where `String.format()` is misused.

### Verifying this change

This change is already covered by existing tests, such as `GroupMetadataTest`.

### Documentation

Check the box below.

Need to update docs? 
  
- [x] `no-need-doc` 
  
 No client facing functionality changes


BewareMyPower (approved):

BewareMyPower:
LGTM. I think we need some strict check for it. Do you have any idea about adding the lint tool?

ZiyaoWei:
I'm just using the IntelliJ linter, but I have no idea what it actually is. Will look into something more strict; I think we already use spotbugs, so maybe something like errorprone?

BewareMyPower:
Good idea. Could you open an issue for it?

PR #1394, opened 2022-07-09
Use `.equals()` instead of `==` for equality checks in GroupMetadata

ZiyaoWei:
### Motivation

It seems during the Scala to Java translation some equality checks aren't ported correctly. This could cause potential bugs as demonstrated in the updated tests.

### Modifications

Use `.equals()` instead of `==`.

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

This change is already covered by existing tests, such as `GroupMetadataTest`.

### Documentation

Check the box below.

Need to update docs? 
  
- [x] `no-need-doc` 
  
  Minor bug fix

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupMetadata.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupMetadata.java
@@ -311,7 +311,7 @@ public class GroupMetadata {
             this.protocolType = Optional.of(member.protocolType());
         }
 
-        checkArgument(groupId == member.groupId());
+        checkArgument(groupId.equals(member.groupId()));
         checkArgument(Objects.equals(protocolType.orElse(null), member.protocolType()));
         checkArgument(supportsProtocols(member.protocols()));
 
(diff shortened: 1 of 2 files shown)


ZiyaoWei:
A couple of things:

- I am actually not 100% sure that using `==` wasn't intentional since it seems like it would have been a very obvious bug that we should be able to catch a while ago? I think most likely it's a bug, but if it isn't we should probably comment on it.
- I'm not sure the updated unit tests make any sense - it looks really weird to construct objects just for this test case. Maybe we should change all the objects to suppliers since we have a fair amount of code converted from Scala and there might be more cases like this, but in the long run I don't see the value in that. Maybe we should cover this with something like the continuous verification tests? cc: @utahkay

BewareMyPower:
> I am actually not 100% sure that using == wasn't intentional 

I think not. The group coordinator related code was added at the very beginning of KoP and has not been modified.

> it looks really weird to construct objects just for this test case.

`GroupMetadataTest` is only a unit test for `GroupMetadata`.

I'm also confused about why this bug was not exposed before, for example, `GroupMetadata#add` should be a common seen case:

```java
    public void add(MemberMetadata member) {
        if (members.isEmpty()) {
            this.protocolType = Optional.of(member.protocolType());
        }

        checkArgument(groupId == member.groupId());
```

The exception should be thrown if `groupId` and `member.groupId()` are not the same object. But this exception was never seen before. Could you look deeper into when it was called and why it didn't fail?

Let's talk back to this PR itself, great find! Please fix the Codacy check as well.

BewareMyPower (approved):

PR #1399, opened 2022-07-12
[Refactor] Introduce Scala-like Either and Option classes

BewareMyPower:
### Motivation

Currently we migrate the Scala code to Java for transaction related
code from Kafka. Java 8 doesn't support pattern matching, which is
widely used in Kafka's source code. And Scala's `Either` class is also
widely used in transaction related code.

Before this PR, `ErrorsAndData<T>` is an alternative `Either<Errors, T>`
in KoP. However, the design is not good for following reasons:
1. It's not immutable, and the setters can be avoided in the places they
   are used.
2. The semantic of `Either` should be OR, not AND. i.e. only one of
   these two fields should be not null. Since `Errors.NONE` and
  `Optional.empty()` are somehow null-like, the
  `ErrorsAndData<Optional<T>>` object could be ambiguous when `errors`
  is `Errors.NONE` and `data` is `Optional.empty()`.

### Modifications

First, add `Either` and `Option` classes, both of which have a `match`
method that simulates the pattern matching semantic. For example,

```java
final Either<Integer, String> either = Either.right("hello");
either.match(i -> System.out.println("left: " + i), s -> System.out.println("right: " + s));
```

is nearly equivalent with the following code in Scala.

```scala
val either: Either[Int, String] = Right("hello");
either match {
  case Left(i) => println(s"left: $i")
  case Right(s) => println(s"right: $s")
}
```

Then, replace all `ErrorsAndData` usages with `Either`. The `Option` is
only used in the returned value of `TransactionStateManager#getTransactionState`.

To avoid too many code changes, some code that can be simplified by
`match` method still uses `isLeft()` and `getLeft()` to keep the
original style.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/scala/Option.java
```diff
@@ -0,0 +1,87 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.streamnative.pulsar.handlers.kop.scala;
+
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+public class Option<T> {
```
tisonkun, line 20: This class seems a subset of `java.util.Optional`. What's the differentce?

BewareMyPower: The only difference is the `match` method that is only used in `TransactionCoordinator#handleInitProducerId`. The initial purpose is to use `Option#match` method everywhere. But I found it could lead to many code changes later.

I think for now, we can change it back to Java's `Optional`.

ZiyaoWei (approved):

Demogorgon314 (approved):

PR #1404, opened 2022-07-18
[Refactor] Unify the pending inter-broker transaction requests

BewareMyPower:
### Motivation

The pending inter-broker transaction requests, including
WriteTxnMarkers, SaslHandshake and SaslAuthenticate requests, are not
unified currently. There are two maps whose key is the correlation id,
but the values are different in `TransactionMarkerChannelHandler`. One
is `InFlightRequest` while the other is `PendingGenericRequest`.
- InFlightRequest: the WriteTxnMarkers request
- PendingGenericRequest: the SASL related requests

The methods for serialization are not unified as well. `InFlightRequest`
uses `KopRequestUtils` to serialize the request to a Netty `ByteBuf`,
while `PendingGenericRequest` uses Kafka's utilities to serialize the
request to a NIO `ByteBuffer`.

In addition, the callback for WriteTxnMarkers response only accepts a
`WriteTxnMarkersResponse` instance, while in Kafka it's a
`ClientResponse`, which contains some extra fields like the request
header and the remote address. Therefore, currently KoP prints a
meaningless log and it's not debug level:

> Received WriteTxnMarker response from node with correlation id $correlationId

### Modifications

- Add `PendingRequest` class as the abstraction of the pending request.
- Add `ResponseContext` as the callback's parameter so that the callback
  can print logs with enough information.
- Change how the correlation id is generated to make it consistent with
  Kafka.

### Documentation

Check the box below.

Need to update docs? 
  
- [x] `no-need-doc` 
  
  No functionality change


BewareMyPower:
Add `type/bug` label for this PR because it looks like to have fixed a bug accidentally. I will open an issue to describe how to reproduce it.

eolivelli (approved):
great work !

LGTM

Demogorgon314 (approved):

BewareMyPower:
This refactor might make the `SaslPlainTestBase#transactionsReadCommittedTest` fail. Revert it from branch-2.9.3.

@Demogorgon314 Could you take a look at this issue (not an urgent task) later?

Demogorgon314:
Sure. I'll take a look.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionMarkerChannelHandler.java
```diff
@@ -331,38 +223,46 @@ private CompletableFuture<ChannelHandlerContext> authenticateInternal(ChannelHan
...
-                    null
+            final PendingRequest pendingRequest = new PendingRequest(
+                    ApiKeys.SASL_AUTHENTICATE,
+                    correlationIdGenerator.next(),
+                    newSaslAuthenticate(saslAuthBytes),
+                    __ -> {}
             );
-            sendGenericRequestOnTheWire(channel, fullRequest, result);
-
+            enqueueRequest(channel, pendingRequest);
```
Demogorgon314, line 233: @BewareMyPower Not sure why, after cherry-picking this PR, branch-2.9 will lose this line. I have re-cherry-picked this PR and added this line. The test will not fail now.

https://github.com/streamnative/kop/blob/e298ccb803d1bc132c079e31799f7720a9616265/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionMarkerChannelHandler.java#L226-L234

PR #1405, opened 2022-07-18
[improve] Make the LengthFieldPrepender singleton

Demogorgon314:
### Motivation

The `LengthFieldPrepender` is a sharable handler, so we can make it singleton to avoid too many objects created.

### Modifications

Make the LengthFieldPrepender singleton

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaChannelInitializer.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaChannelInitializer.java
@@ -63,6 +63,8 @@ public class KafkaChannelInitializer extends ChannelInitializer<SocketChannel> {
     private final RequestStats requestStats;
     private final OrderedScheduler sendResponseScheduler;
 
+    private final LengthFieldPrepender lengthFieldPrepender;
+
     public KafkaChannelInitializer(PulsarService pulsarService,
                                    KafkaServiceConfiguration kafkaConfig,
                                    TenantContextManager tenantContextManager,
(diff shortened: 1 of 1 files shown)


BewareMyPower (commented):
The `LengthFieldPrepender` is created per channel, there should not be too many objects. As we can see, there are many fields of `KafkaRequestHandler` that are created per channel as well. Even in `KafkaChannelInitializer`, `SslHandler` and `LengthFieldBasedFrameDecoder` are created per channel as well.

In addition, I'm not sure if it's safe to share this object among all channels. Maybe yes, but I'm wondering making it sharable could really bring some improvements? I just took a look at `PulsarChannelInitializer#initChannel`, it also doesn't pass any singleton object to `addLast` method.

Demogorgon314:
@BewareMyPower 
You can see the `LengthFieldPrepender` has a `@Sharable` annotation, which means the same instance of the annotated `ChannelHandler` can be added to one or more `ChannelPipeline`s multiple times without a race condition.

```java
/**
 * Indicates that the same instance of the annotated {@link ChannelHandler}
 * can be added to one or more {@link ChannelPipeline}s multiple times
 * without a race condition.
 * <p>
 * If this annotation is not specified, you have to create a new handler
 * instance every time you add it to a pipeline because it has unshared
 * state such as member variables.
 * <p>
 * This annotation is provided for documentation purpose, just like
 * <a href="http://www.javaconcurrencyinpractice.com/annotations/doc/">the JCIP annotations</a>.
 */
@Inherited
@Documented
@Target(ElementType.TYPE)
@Retention(RetentionPolicy.RUNTIME)
@interface Sharable {
    // no value
}
```

```java
@Sharable
public class LengthFieldPrepender extends MessageToMessageEncoder<ByteBuf> {

```


Also, the Pulsar has a singleton handler names `ByteBufPairEncoder`.

BewareMyPower (approved):

PR #1423, opened 2022-07-27
[improve] Trigger pending fetches when producing messages

### LINKED ISSUE #1419: [FEATURE] Trigger pending fetches when producing messages
When the Kafka client issues a Fetch and sets a maxWait time we are already scheduling a DelayedFetch, but there is no way to trigger that Fetch and it is deemed to wait for the fully timeout.
This adds latency spikes on the Kafka Consumer.

Demogorgon314:
Fixes #1419 

### Motivation

When the Kafka client issues a Fetch and sets a max wait time we are already scheduling a DelayedFetch, but there is no way to trigger that Fetch and it is deemed to wait for the full timeout.
This adds latency spikes to the Kafka Consumer.

#### Logic

1. Refactor the existing fetch logic to make the delayed fetch easier to implement. 
2. Implement the DelayedFetch:
    The delayed fetch can be triggered by the produce operation but there are three cases:
    1. When trying to complete the operation triggered, get the last position failed(PositionImpl.EARLIEST), the delay fetch will complete immediately with the previous fetch result.
    2. When the previous last position is last then the current last position, try to adjust the max read entries num, and try to read the records again.
    3. On expiration happened, complete immediately with the previous fetch result.

#### Metrics

Add `WAITING_FETCHES_TRIGGERED` stats.

### Benchmark Result

Here is some benchmark result with the following workloads in standalone env:

#### Workloads
---

```
topics: 1
partitionsPerTopic: 10
messageSize: 100
#payloadFile: "payload/payload-100b.data"
useRandomizedPayloads: true
randomBytesRatio: 0.5
randomizedPayloadPoolSize: 1000

subscriptionsPerTopic: 1
consumerPerSubscription: 10
producersPerTopic: 10

# Discover max-sustainable rate
producerRate: 200000

consumerBacklogSizeGB: 0
testDurationMinutes: 5
```

#### End To End Latency 95pct:
---

End To End Latency 95pct

#### End To End Latency Avg:
---

End To End Latency Avg

#### End To End Latency Quantiles:
---

End To End Latency Quantiles

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [ ] `no-need-doc` 
  
  (Please explain why)
  
- [x] `doc` 
  

eolivelli (commented):
I am very interested in this change !
(Last time I tried it was a failure)

Demogorgon314:
Here is some benchmark result with the following workloads in standalone env:

Workloads
---

```
topics: 1
partitionsPerTopic: 10
messageSize: 100
#payloadFile: "payload/payload-100b.data"
useRandomizedPayloads: true
randomBytesRatio: 0.5
randomizedPayloadPoolSize: 1000

subscriptionsPerTopic: 1
consumerPerSubscription: 10
producersPerTopic: 10

# Discover max-sustainable rate
producerRate: 200000

consumerBacklogSizeGB: 0
testDurationMinutes: 5
```

End To End Latency 95pct:
---
End To End Latency 95pct

End To End Latency Avg:
---
End To End Latency Avg

End To End Latency Quantiles:
---
End To End Latency Quantiles

Demogorgon314:
@BewareMyPower @eolivelli This PR is ready to get a quick review for earlier feedback.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaApisTest.java
```diff
@@ -1023,4 +1032,60 @@ public void testIllegalManagedLedger() throws Exception {
...
+        ExecutorService executorService = Executors.newFixedThreadPool(1);
+
+        long startTime = System.currentTimeMillis();
+        AtomicLong endTime = new AtomicLong(0);
+        CountDownLatch latch = new CountDownLatch(1);
+        executorService.submit(() -> {
+            int totalRead = 0;
+            do {
+                ConsumerRecords<String, String> result = consumer1.poll(Duration.ofMillis(maxWaitMs));
+                result.forEach(record -> {
```
BewareMyPower, line 1068: This test is not accurate. There is a chance that `poll` is called after some messages are sent. Instead of testing a real world consumer, it would be better to test `KafkaRequestHandler#handleFetchRequest` so that we won't need to run a consumer in another thread.

BewareMyPower (changes requested):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaApisTest.java
Demogorgon314: Yes, I'll add more tests in `KafkaRequestHandler#handleFetchRequest`.

For the new delayed fetch design, do you have any comments?

BewareMyPower: No at the moment.

I just started reviewing this design. The first thing I'm concerned about is the test that shows the difference brought by this PR. Actually I wrote a similar unit test, before this PR it took about 1472 ms to receive all 10 messages. The time is far less than the max wait ms (3000 ms). So I found the test is somehow meaningless.

I'm still going to minimize the code changes. You don't have to address my comments before I complete the review. Maybe I'll open a new PR based on this PR, or just leave some comments or requested changes.

BewareMyPower: The test:

```java
    @Test
    public void testFetchMinBytesSingleConsumer() throws Exception {
        final String topic = "testMinBytesTopic";
        admin.topics().createPartitionedTopic(topic, 1);

        final int maxWaitMs = 3000;
        final int minBytes = 1;

        @Cleanup
        final KafkaConsumer<String, String> consumer1 = createKafkaConsumer(maxWaitMs, minBytes);
        consumer1.assign(Collections.singleton(new TopicPartition(topic, 0)));
        assertTrue(consumer1.poll(Duration.ofMillis(maxWaitMs)).isEmpty());

        final ExecutorService executorService = Executors.newSingleThreadExecutor();
        final CountDownLatch latch = new CountDownLatch(1);
        final int numMessages = 10;
        final long startTime = System.currentTimeMillis();
        final AtomicLong endTime = new AtomicLong(0);
        executorService.execute(() -> {
            for (int i = 0; i < numMessages; ) {
                final int previous = i;
                for (ConsumerRecord<String, String> record : consumer1.poll(Duration.ofMillis(maxWaitMs))) {
                    log.info("Received record: {}", record.value());
                    i++;
                }
                log.info("The consumer read {} records totalRead {}", i - previous, i);
            }
            endTime.set(System.currentTimeMillis());
            latch.countDown();
        });

        final KafkaProducer<String, String> producer = createKafkaProducer();
        produceData(producer, Collections.singletonList(new TopicPartition(topic, 0)), numMessages);
        latch.await();
        log.info("Received {} records for {} ms", numMessages, endTime.get() - startTime);
        executorService.shutdown();
    }
```

The log:

> 15:52:47.022 [main:io.streamnative.pulsar.handlers.kop.KafkaApisTest@1067] INFO  io.streamnative.pulsar.handlers.kop.KafkaApisTest - Received 10 records for 1272 ms

BewareMyPower: A more accurate test example:

```java
    @Test
    public void testFetchMinBytesSingleConsumer() throws Exception {
        final String topic = "testMinBytesTopic";
        final TopicPartition topicPartition = new TopicPartition(topic, 0);
        admin.topics().createPartitionedTopic(topic, 1);

        final int maxWaitMs = 3000;
        final int minBytes = 1;
        @Cleanup
        final KafkaHeaderAndRequest request = buildRequest(FetchRequest.Builder.forConsumer(maxWaitMs, minBytes,
                Collections.singletonMap(topicPartition, new FetchRequest.PartitionData(
                        0L, -1L, 1024 * 1024, Optional.empty()
                ))));
        final CompletableFuture<AbstractResponse> future = new CompletableFuture<>();
        final long startTime = System.currentTimeMillis();
        kafkaRequestHandler.handleFetchRequest(request, future);

        // Trigger the fetch
        final int numMessages = 10;
        final KafkaProducer<String, String> producer = createKafkaProducer();
        produceData(producer, Collections.singletonList(new TopicPartition(topic, 0)), numMessages);

        final FetchResponse<MemoryRecords> response = ((ResponseCallbackWrapper)
                future.get(maxWaitMs + 1000, TimeUnit.MILLISECONDS)).getResponse();
        final long endTime = System.currentTimeMillis();
        log.info("Take {} ms to process FETCH request, record count: {}",
                endTime - startTime, response.responseData().size());
    }
```

> 17:27:16.386 [main:io.streamnative.pulsar.handlers.kop.KafkaApisTest@1062] INFO  io.streamnative.pulsar.handlers.kop.KafkaApisTest - Take 3028 ms to process FETCH request, record count: 1

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -1540,18 +1543,110 @@ protected void handleFetchRequest(KafkaHeaderAndRequest fetch,
...
+                        partitions.putAll(erroneous);
+                        resultFuture.complete(new ResponseCallbackWrapper(new FetchResponse<>(
+                                Errors.NONE,
+                                partitions,
+                                ((Integer) THROTTLE_TIME_MS.defaultValue),
+                                request.metadata().sessionId()), () ->
+                                resultMap.forEach((__, readRecordsResult) -> {
+                                    if (readRecordsResult.decodeResult() != null) {
+                                        readRecordsResult.decodeResult().recycle();
+                                    }
```
BewareMyPower, line 1610: Could it be `null`? If there is a race condition, I think there's also a chance that `readRecordsResult.decodeResult()` is not null in line 1608 but it became `readRecordsResult.decodeResult()` is null in line 1609.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java
```diff
@@ -226,13 +327,334 @@ public CompletableFuture<Long> appendRecords(final MemoryRecords records,
...
+                context.getSharedState().getKafkaTopicConsumerManagerCache().removeAndCloseByTopic(fullPartitionName);
+                if (log.isDebugEnabled()) {
+                    log.debug("Fetch for {}: no tcm for topic {} return NOT_LEADER_FOR_PARTITION.",
+                            topicPartition, fullPartitionName);
+                }
+                future.complete(ReadRecordsResult.error(Errors.NOT_LEADER_FOR_PARTITION));
+                return;
+            }
+            if (checkOffsetOutOfRange(tcm, offset, topicPartition, startPrepareMetadataNanos)) {
+                future.complete(ReadRecordsResult.error(Errors.OFFSET_OUT_OF_RANGE));
```
BewareMyPower, line 373: Should we add `return;` here?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -1540,18 +1548,110 @@ protected void handleFetchRequest(KafkaHeaderAndRequest fetch,
...
+                        && kafkaConfig.isKafkaTransactionCoordinatorEnabled()) {
+                    transactionCoordinator = getTransactionCoordinator();
+                }
+                String namespacePrefix = currentNamespacePrefix();
+                int fetchMaxBytes = request.maxBytes();
+                int fetchMinBytes = Math.min(request.minBytes(), fetchMaxBytes);
+                if (interesting.isEmpty()) {
+                    if (log.isDebugEnabled()) {
+                        log.info("Fetch interesting is empty.");
+                    }
```
BewareMyPower, line 1583: See https://github.com/streamnative/kop/pull/1423#discussion_r945150604

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
Demogorgon314: The `readRecordsResult#decodeResult ` can be null because we will get an "Errors.NONE" error if there are no more entries to read.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java
```diff
@@ -102,6 +139,118 @@ public static class LogAppendInfo {
         private KopLogValidator.CompressionCodec targetCodec;
     }
 
+    @Data
+    @ToString
+    @Accessors(fluent = true)
+    @AllArgsConstructor
+    public static class ReadRecordsResult {
```
BewareMyPower, line 146: I'm wondering if this recyclable class is meaningful? This object will finally be converted to `FetchResponse.PartitionData<Records>`. The only concern is the `DecodeResult` object that need to be recycled.

We can simply use a `Pair<FetchResponse.PartitionData<Records>, DecodeResult>`. Or you can insist the recyclable way to wrap these two classes.

Demogorgon314: We also need `Position lastPosition;`, so I think we should keep it.

gaozhangmin:
@Demogorgon314  Do we have test can reproduce the error `Unexpected error code 7 while fetching data` before applying this pr?

Demogorgon314:
@gaozhangmin No, the error code 7 is caused by a timeout error when handling the fetch request. 

If you check the Kafka code, you can see when the fetch request timeout, it will return an empty recodes. And it will never return the timeout error response.

As you mention, adding a test to cover this case is better. I'll add it later.

eolivelli:
@Demogorgon314 @BewareMyPower 
what's the status of this patch ?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/DelayedFetch.java
```diff
@@ -13,38 +13,124 @@
...
+        if (this.callback.isDone()) {
+            return true;
+        }
+        for (Map.Entry<TopicPartition, PartitionLog.ReadRecordsResult> entry : readRecordsResult.entrySet()) {
+            TopicPartition tp = entry.getKey();
+            PartitionLog.ReadRecordsResult result = entry.getValue();
+            PartitionLog partitionLog = replicaManager.getPartitionLog(tp, context.getNamespacePrefix());
+            PositionImpl currLastPosition = (PositionImpl) partitionLog.getLastPosition(context.getTopicManager());
+            if (currLastPosition.compareTo(PositionImpl.EARLIEST) == 0) {
+                HAS_ERROR_UPDATER.compareAndSet(this, false, true);
```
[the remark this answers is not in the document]
Demogorgon314: Good point.

Demogorgon314:
@eolivelli This patch is still in review.

BewareMyPower:
@eolivelli @Demogorgon314 Sorry for the delayed review on this PR. I will review it ASAP.

eolivelli:
@Demogorgon314 FYI I and @dave2wave are going to run performance tests over this patch and we will provide feedback.
It seems that this patch has a good chance to solve p9X latency issues.

Demogorgon314:
@eolivelli @dave2wave Thanks you all for providing feedback!

BewareMyPower (approved):

BewareMyPower:
Overall LGTM.

BewareMyPower:
Could this PR be cherry-picked into release branches?

Demogorgon314:
Oh, this commit just fixed your review comments. https://github.com/streamnative/kop/pull/1423#discussion_r974220673

Demogorgon314:
> Could this PR be cherry-picked into release branches?

@BewareMyPower I'll try to cherry-pick it to 2.9.x and 2.10.x, but I think we can wait for the performance test feedback before cherry-picking.

eolivelli:
> @eolivelli Oh, this commit just fixed your review comments. [#1423 (comment)](https://github.com/streamnative/kop/pull/1423#discussion_r974220673)

thanks

Demogorgon314:
@eolivelli It looks like commit https://github.com/streamnative/kop/commit/b172d3e0855a5fe7acf68f9aedd00839047d8e04 triggered the codecov check, It looks like forced push,  but it only changed one line.


image

dave2wave:
I ran OMB testing of DataStax's fork of KoP including this patch today from a build provided by @eolivelli .

TL/DR - The results can be summarized as follows:
1. The end to end latency is remarkably improved. Excellent work!
2. The publishing latency is slightly slower, but that is not surprising considering the "delayed fetch" in the new code.
3. There seems to be a resource leak given that later runs had worse end to end latency.

The tests were run on pretty robust cluster that is capable of 1 million messages per second of 100 byte payloads. We used an enhanced OMB based on the DS fork which allows for the produceRate to be stepped up incrementally. All test are 10 topics with 3 partitions.

A. Incremental Pulsar OMB from 20,000 to 400,000 mps of 1K payloads.
B. Incremental KoP OMB from 20,000 to 400,000 mps of 1K payloads.
Screen Shot 2022-11-08 at 4 21 31 PM
Screen Shot 2022-11-08 at 4 21 20 PM
C. KoP OMB at 400,000 mps of 1K payloads. One would expect that the performance would look similar to the last 5 minutes of the incremental KoP, but the latency is worse across the board.
Screen Shot 2022-11-08 at 4 25 09 PM
Screen Shot 2022-11-08 at 4 25 21 PM
D. KoP OMB at 1 million mps of 100 byte payloads.

dave2wave:
Later I ran another test B which was not as good as the first. My theory is that there is a resource leak happening.
Here is some OMB output which should give you a comparison.

Demogorgon314:
@dave2wave Thanks for sharing the benchmark results! I'll check if there are any resource leaks.

eolivelli:
@Demogorgon314 @BewareMyPower after analysing the dumps of the JVMs of @dave2wave 's tests I have come to this small improvement
https://github.com/streamnative/kop/pull/1563

let's follow up there the discussion

Issue #1475, opened 2022-09-12
[FEATURE] Live migration tooling


ZiyaoWei:
**Is your feature request related to a problem? Please describe.**
A clear solution for migration from Kafka to KoP currently doesn't exist.

**Describe the solution you'd like**
Add a live migration system so users could migrate from Kafka to KoP.

**Describe alternatives you've considered**
We can provide offline migration tooling, which wouldn't be as user friendly.

**Additional context**
None

eolivelli:
I think that it depends on your application.

For instance if you are not keeping data for much time it is easy to attach your producers to the new cluster (Pulsar) and add new consumers on the new cluster.
When the data from the old cluster is consumed you can drop it.

But if you have higher requirements then you have to design something specific

ZiyaoWei:
Thanks @eolivelli! I have a design doc that we've gone through internally, but I'd love your (and anybody else in the community's) take on it as well! It's my oversight to not have shared it earlier here.

Link to the doc: https://docs.google.com/document/d/1MCpJEJEkIuMW-0IK7ubNOvE1x2hjATv52KmiAQdhgrE

PR #1476, opened 2022-09-12
Add migration handler

ZiyaoWei:
Master Issue: #1475

### Motivation

The migration service needs a set of API for the clients to talk to.

### Modifications

This adds the HTTP request handling boilerplate for the new live migration service. Note this is largely based on the Schema Registry HTTP handling code and listens to a separate port from the main Kafka listener, and in the future we should probably refactor the Schema Registry code to using this common boilerplate.

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

The migration service is WIP so no tests are written yet but we can make sure the endpoint works by visiting it.

### Documentation
  
- [x] `no-need-doc` 
  
  Doesn't need docs at this moment, will add doc after the live migration service is done.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/http/HttpRequestProcessor.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/http/HttpRequestProcessor.java
@@ -1,0 +1,174 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.streamnative.pulsar.handlers.kop.http;
+
+import static io.netty.handler.codec.http.HttpResponseStatus.INTERNAL_SERVER_ERROR;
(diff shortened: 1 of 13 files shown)


eolivelli:
Can you please share your plan ?
do you have some design document ?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/http/HttpRequestProcessor.java
```diff
@@ -0,0 +1,140 @@
...
+        try {
+            String body = MAPPER.writeValueAsString(new ErrorModel(httpStatusCode, err.getMessage()));
+            httpResponse = new DefaultFullHttpResponse(HTTP_1_1, error,
+                    Unpooled.copiedBuffer(body, CharsetUtil.UTF_8));
+            httpResponse.headers().set(HttpHeaderNames.CONTENT_TYPE, "application/vnd.schemaregistry.v1+json");
+            httpResponse.headers().set(HttpHeaderNames.CONTENT_LENGTH, body.length());
+            addCORSHeaders(httpResponse);
+        } catch (JsonProcessingException impossible) {
+            String body = "Error " + err;
+            httpResponse = new DefaultFullHttpResponse(HTTP_1_1, error,
```
lin-zhao, line 89: This is a duplicate of line 82. There should be a better way to organize this.

If there's no equivalent of `Try` class of scala, how about

```
String body = "Error " + err;
try {
    body = MAPPER.writeValueAsString(new ErrorModel(httpStatusCode, err.getMessage()));
} catch (JsonProcessingException impossible) {
} finally {
  httpResponse = ...
}
```

lin-zhao (changes requested):
Mostly styling questions. But Unit tests are missing in this PR. They should be added before merge.

lin-zhao (approved):
LGTM

BewareMyPower (approved):

PR #1501, opened 2022-09-19
[BUG#1500] Cannot submit offset correctly

### LINKED ISSUE #1500: [BUG] Cannot submit offset correctly
**Describe the bug**
81ad161cc951006f0395748d9fa2270

`                    long baseOffset = MessageMetadataUtils.getMockOffset(
                        lastMessageId.getLedgerId(),
                        lastMessageId.getEntryId()
                    );`
The baseOffset is not monotonically increasing, consumer cannot submit offset correctly.
The offset can be detected incorrectly through Kafka Exporter.

**To Reproduce**
Normal consumption messages can reproduce the bug.

lwclover:
Fixes #1500 

### Motivation

The baseOffset is not monotonically increasing, consumer cannot submit offset correctly.

### Modifications

When the offset is submitted, the increasing judgment mode of baseOffset is changed

### Verifying this change

This change is already covered by existing tests.

1.passed GroupMetadataManagerTest
2.passed GroupMetadataTest

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupMetadata.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupMetadata.java
@@ -52,7 +52,6 @@ import lombok.experimental.Accessors;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.bookkeeper.mledger.impl.PositionImpl;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.math.NumberUtils;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.pulsar.common.schema.KeyValue;
 
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupMetadata.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupMetadata.java
@@ -51,6 +51,7 @@ import lombok.Setter;
 import lombok.experimental.Accessors;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.math.NumberUtils;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.pulsar.common.schema.KeyValue;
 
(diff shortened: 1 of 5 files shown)


BewareMyPower:
Could you fill the PR description with the template [here](https://github.com/streamnative/kop/blob/master/.github/PULL_REQUEST_TEMPLATE.md)? You should have seen the template when you open a PR like

image

BewareMyPower (changes requested):
Could you add a test to avoid the regression? It would be better to add a test that shows how to reproduce the bug in #1500.

lwclover:
> Could you fill the PR description with the template [here](https://github.com/streamnative/kop/blob/master/.github/PULL_REQUEST_TEMPLATE.md)? You should have seen the template when you open a PR like
> 
> image
> 
> In addition, you should sign the CLA.

Signed

BewareMyPower:
The existing tests can only verify `getMockOffset`, but it cannot verify the case when consumer failed to commit offset without this patch.

lwclover:
> The existing tests can only verify `getMockOffset`, but it cannot verify the case when consumer failed to commit offset without this patch.

OK, I will add a test case to cover the situation.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupMetadata.java
```diff
@@ -166,11 +167,22 @@ public static class GroupSummary {
      */
     @Data
     static class CommitRecordMetadataAndOffset {
-        private final Optional<Long> appendedBatchOffset;
+        private final Optional<String> appendedBatchOffset;
```
Demogorgon314, line 170: I'm more prefer to use `org.apache.bookkeeper.mledger.impl.PositionImpl` in here. 

Then we can use `this.appendedBatchOffset.get().compareTo(that.appendedBatchOffset.get()) < 0` to compare this two position.

Demogorgon314 (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupMetadata.java
lwclover: You are right

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupMetadata.java
```diff
@@ -166,11 +167,11 @@ public static class GroupSummary {
      */
     @Data
     static class CommitRecordMetadataAndOffset {
-        private final Optional<Long> appendedBatchOffset;
+        private final Optional<PositionImpl> appendedBatchOffset;
```
BewareMyPower, line 170: It's better to rename it with `appendedBatchPosition` since it doesn't represent an offset anymore.

lwclover: renamed

Demogorgon314 (approved):

BewareMyPower (approved):

PR #1504, opened 2022-09-19
[refactor][group] Simplify code and optimize loading partitions management

### LINKED ISSUE #1502: [FEATURE] Refactor the loading and owned partitions cache management
When I revisited the `GroupMetadataManager` related code, I found the code is not consistent with the latest Kafka implementation since it was written in a very early time.

Follow the Kafka 3.0 source code to refactor the `GroupMetadataManager` to manage these two most important fields correctly:

```scala
  /* partitions of consumer groups that are being loaded, its lock should be always called BEFORE the group lock if needed */
  private val loadingPartitions: mutable.Set[Int] = mutable.Set()

  /* partitions of consumer groups that are assigned, using the same loading partition lock */
  private val ownedPartitions: mutable.Set[Int] = mutable.Set()
```

BewareMyPower:
Fixes #1502

### Motivation

Currently the `partitioner` in `GroupMetadataManager` can be customized
but it's only for test. The reason is that when
`testRequestHandlingWhileLoadingInProgress` was migrated from Kafka
source code, there is something wrong:
1. `otherGroupId` should be tested while it's actually `groupId`.
2. `otherGroupId` should be "myGroup", not "myGroupId", which cause the
   error of hash code.

Therefore, a customized partitioner is used in `GroupCoordinatorTest`.

In addition, after https://github.com/apache/kafka/pull/9441, the
partition of a group id won't be put into `loadPartitions` if it already
exists in `ownedPartitions`.

### Modifications

- Don't add partition into `loadPartitions` when it exists in
  `ownedPartitions`.
- Use the fixed partitioner to calculate the partition of a group id.
  Though the algorithm is still a little different from Kafka.
- Add the partition to `ownedPartitions` even if loading offsets failed.
- Add a separate method `removeGroupsAndOffsets` to remove caches and
  close the producer and reader of a partition. In this method, use
  try-finally block for lock instead of `inLock` to avoid usage of
  `AtomicInteger` in a lambda.
- Fix the wrong tests.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupMetadataManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupMetadataManager.java
@@ -50,7 +50,6 @@ import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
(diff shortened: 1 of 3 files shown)


BewareMyPower:
@Demogorgon314 PTAL

Demogorgon314 (approved):

PR #1505, opened 2022-09-19
Add migration handler and various code style fixes

ZiyaoWei:
Master Issue: #1475 

### Motivation

The migration service needs a set of API for the clients to talk to.

Also, there are a lot of code style warnings for these files in IntelliJ which makes it harder to find real issues.

### Modifications

This adds the HTTP request handling boilerplate for the new live migration service. Note this is largely based on the Schema Registry HTTP handling code and listens to a separate port from the main Kafka listener, and in the future we should probably refactor the Schema Registry code to using this common boilerplate. Also contains some IntelliJ auto code style fixes.

### Verifying this change

### Documentation

Check the box below.

Need to update docs? 
  
- [x] `no-need-doc` 
  
  Doesn't need docs at this moment, will add doc after the live migration service is done.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -18,7 +18,6 @@ import static com.google.common.base.Preconditions.checkState;
 import static io.streamnative.pulsar.handlers.kop.KafkaServiceConfiguration.TENANT_ALLNAMESPACES_PLACEHOLDER;
 import static io.streamnative.pulsar.handlers.kop.KafkaServiceConfiguration.TENANT_PLACEHOLDER;
 import static java.nio.charset.StandardCharsets.UTF_8;
-import static org.apache.kafka.common.protocol.CommonFields.THROTTLE_TIME_MS;
 import static org.apache.kafka.common.requests.CreateTopicsRequest.TopicDetails;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -1551,10 +1550,8 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
             log.debug("[{}] Request {} Fetch request. Size: {}. Each item: ",
                 ctx.channel(), fetch.getHeader(), request.fetchData().size());
 
-            request.fetchData().forEach((topic, data) -> {
-                log.debug("  Fetch request topic:{} data:{}.",
-                    topic, data.toString());
-            });
+            request.fetchData().forEach((topic, data) -> log.debug("  Fetch request topic:{} data:{}.",
+                topic, data.toString()));
         }
         TransactionCoordinator transactionCoordinator = null;
         if (request.isolationLevel().equals(IsolationLevel.READ_COMMITTED)
@@ -1577,7 +1574,6 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
 
         Map<String, byte[]> protocols = new HashMap<>();
         request.groupProtocols()
-            .stream()
             .forEach(protocol -> protocols.put(protocol.name(), Utils.toArray(protocol.metadata())));
         getGroupCoordinator().handleJoinGroup(
             request.groupId(),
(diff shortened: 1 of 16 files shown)


BewareMyPower:
It's better to run `mvn checkstyle:check` in your local env to ensure the code style can pass the checkstyle check.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -458,12 +458,10 @@ protected ApiVersionsResponse overloadDefaultApiVersionsResponse(boolean unsuppo
...
-                        case LIST_OFFSETS:
-                            // V0 is needed for librdkafka
-                            versionList.add(new ApiVersion((short) 2, (short) 0, apiKey.latestVersion()));
-                            break;
-                        default:
-                            versionList.add(new ApiVersion(apiKey));
+                        case LIST_OFFSETS ->
+                                // V0 is needed for librdkafka
+                                versionList.add(new ApiVersion((short) 2, (short) 0, apiKey.latestVersion()));
+                        default -> versionList.add(new ApiVersion(apiKey));
```
BewareMyPower, line 464: We'd better still keep the code compatible with Java 8 for existing code so that we can cherry-pick the PR into older branches.

BTW, it looks like the JavaDocs plugin still uses an older source as well.

```
Error:  /home/runner/work/kop/kop/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java:461: error: switch rules are not supported in -source 8
Error:                          case LIST_OFFSETS ->
Error:                                            ^
Error:    (use -source 14 or higher to enable switch rules)
```

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -1096,7 +1096,7 @@ private CompletableFuture<Pair<Errors, Long>> fetchOffset(String topicName, long
         CompletableFuture<Pair<Errors, Long>> partitionData = new CompletableFuture<>();
 
         topicManager.getTopic(topicName).thenAccept((perTopicOpt) -> {
-            if (!perTopicOpt.isPresent()) {
+            if (perTopicOpt.isEmpty()) {
```
Demogorgon314, line 1099: The `Optional#isEmpty` is introduced in JDK11. Do we need cherry-pick this PR to 2.9.3 or 2.10.1? Since they are using JDK1.8 to run the tests.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandlerTest.java
```diff
@@ -1165,7 +1155,7 @@ public void testBrokerHandleTopicMetadataRequestAllowAutoTopicCreation(boolean b
             String topicName = "kopBrokerHandleTopicMetadataRequest-" + brokerAllowAutoTopicCreation + "-"
                     + overrideNameSpaceAutoTopicCreation + "-"
                     + allowAutoTopicCreationInRequest;
-            List<String> kafkaTopics = Arrays.asList(topicName);
+            List<String> kafkaTopics = List.of(topicName);
```
BewareMyPower, line 1158: Please revert this change because this PR could be cherry-picked to older branches that are compatible with Java 8

BewareMyPower (changes requested):

tests/src/test/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandlerTest.java
BewareMyPower: Just found we can use `Collections.singletonList` here because the list can be immutable.

Demogorgon314: +1

BewareMyPower:
@ZiyaoWei Just found https://github.com/streamnative/kop/pull/1476 is merged into this branch, did you intend to merge to master?

BewareMyPower:
IMO, it's not a good way for development. The CI won't run if the PR is not merged into master. 

image

So after #1476 is merged, this PR cannot pass the code style check.

It's better to change the target branch of all your PRs to master, including the draft PR https://github.com/streamnative/kop/pull/1513. If one PR (B) relies on another PR (A), you can include the changes in B first, then after A is merged, you can adjust code in branch B and push the extra changes to B by force.

This could also help verify the PR dependency so reviewers can know which PR should be reviewed first.

BewareMyPower:
As you can see, if some other PRs (I see https://github.com/streamnative/kop/pull/1499 as well) are merged into this branch, this PR could become very huge and it's hard to see new changes comparing with the other PRs, which leads to the difficulty to review this PR.

BewareMyPower (approved):

PR #1506, opened 2022-09-19
Refactor KafkaTopicManager to extract out KafkaTopicLookupService

ZiyaoWei:
Master Issue: #1475 

### Motivation

The `KafkaTopicManager` class is really tightly coupled with Kafka request processing; it's constructor requires a `KafkaRequestHandler`, and it caches producer for the topics. However, the topic lookup functionality is also needed in other places, e.g., in the migration service where the migration metadata would be stored in the topic managed ledger properties, which means it should be decoupled from handling Kafka requests.

### Modifications

This extracts a `KafkaTopicLookupService` which is currently stateless and looks up the Pulsar topic for a Kafka topic. This will be needed by the migration service.

### Verifying this change

### Documentation

Check the box below.

Need to update docs?
  
- [x] `no-need-doc` 
  
  Internal refactor

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicLookupService.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicLookupService.java
@@ -1,0 +1,108 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.streamnative.pulsar.handlers.kop;
+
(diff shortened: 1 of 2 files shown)


lin-zhao (approved):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicLookupService.java
```diff
@@ -0,0 +1,108 @@
...
+@Slf4j
+public class KafkaTopicLookupService {
+    private final BrokerService brokerService;
+
+    KafkaTopicLookupService(BrokerService brokerService) {
+        this.brokerService = brokerService;
+     }
+
+    // A wrapper of `BrokerService#getTopic` that is to find the topic's associated `PersistentTopic` instance
+    public CompletableFuture<Optional<PersistentTopic>> getTopic(String topicName, Channel channel) {
```
Demogorgon314, line 40: Good Job! If we can add some unit tests to cover this method will be better!

BewareMyPower (dismissed):

Demogorgon314 (dismissed):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaTopicLookupService.java
Demogorgon314: Sure, you can add test in follow-up PR.

Demogorgon314 (approved):

PR #1563, opened 2022-11-10
[improvement] Release PartitionResponse data as soon as possible

eolivelli:
Motivation:
The response map for a Produce Response corresponding to a Produce operation with non zero "timeout" is retained in memory for the whole timeout time. Under load this looks like a memory leak (it isn't really a leak but it add much pressure on the JVM).

Modification:
Move the lambda expression to a named class and explicitly release the references to the main data structures as soon as possible.

Results:
Even with basic testing with Pulsar Standlone and Kafka perf tools the improvements for the producer is visible.

Steps:
- start Standalone with the Kafka PH
- run the producer, at max speed 

`bin/kafka-producer-perf-test.sh --topic test --record-size 1024 --print-metrics  --throughput -1 --num-records 10000000 --producer.config config/producer.properties`

- run a consumer

`bin/kafka-consumer-perf-test.sh --topic test --consumer.config config/consumer.properties --print-metrics --bootstrap-server localhost:9092 --messages 10000000`

Observe the thought-put of the Producer and heap dumps on the JVM

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/ReplicaManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/ReplicaManager.java
@@ -31,6 +31,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


eolivelli:
The alternative is to remove the DelayedProduceAndFetch task from the Timer, but it is actually more hard and it requires more changes.
This patch doesn't change the internals but it already provides a good enhancement.

PR #1587, opened 2022-11-23
[improve] Remove partition log when bundle unload

Demogorgon314:
### Motivation

Currently, the `PartitionLog` object from `PartitionLogManager` is never removed, even if the topic has been unloaded. Also, if the topic is removed, the object of `PartitionLog` will still in the memory.

### Modifications

* Remove the `PartitionLog` object from `PartitionLogManager`.
* Delete useless `ReplicaManager` by tenant map.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/NamespaceBundleOwnershipListenerImpl.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/NamespaceBundleOwnershipListenerImpl.java
@@ -59,7 +59,11 @@ public class NamespaceBundleOwnershipListenerImpl implements NamespaceBundleOwne
                     if (log.isDebugEnabled()) {
                         log.debug("[{}][{}] Trigger load callback for {}", brokerUrl, listener.name(), topic);
                     }
-                    listener.whenLoad(TopicName.get(topic));
+                    try {
+                        listener.whenLoad(TopicName.get(topic));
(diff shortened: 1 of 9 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/NamespaceBundleOwnershipListenerImpl.java
```diff
@@ -59,7 +59,11 @@ public void onLoad(NamespaceBundle bundle) {
                     if (log.isDebugEnabled()) {
                         log.debug("[{}][{}] Trigger load callback for {}", brokerUrl, listener.name(), topic);
                     }
-                    listener.whenLoad(TopicName.get(topic));
+                    try {
+                        listener.whenLoad(TopicName.get(topic));
+                    } catch (Exception ex) {
+                        log.error("[{}][{}] Failed to do load for {}", brokerUrl, listener.name(), topic, ex);
+                    }
```
BewareMyPower, line 66: We should handle the exception inside the `whenLoad` method. Or just catch the specific exception, e.g. if you worried about that `TopicName.get` might throw an exception.

BewareMyPower (changes requested):

BewareMyPower (approved):

PR #1602, opened 2022-12-02
[fix] Use thread-safe list in TransactionMarkerChannelManager

michaeljmarshall:
### Motivation

The `TransactionMarkerChannelManager#addTxnMarkersToBrokerQueue` method modifies an `ArrayList` from callbacks on futures, which could result in unsafe modification of the list.

### Modifications

* Replace `ArrayList` with `CopyOnWriteArrayList`. I chose `CopyOnWriteArrayList` since it is simple and the expected number of writes to the list is low.

### Verifying this change

### Documentation
  
- [x] `no-need-doc` 
  
This is a simple fix.


eolivelli (approved):
lgtm

Demogorgon314 (approved):
Good catch!

BewareMyPower (approved):

PR #1604, opened 2022-12-02
[fix] Use correct path for pulsar client TlsTrustCertsFilePath

michaeljmarshall:
### Motivation

This is a trivial fix. The `getTlsCertificateFilePath` is meant to be the tls certificate chain when sending to a peer, and the `getTlsTrustCertsFilePath` is meant to be the root CAs.

This PR could technically break installations. I'm unsure how we'd like to handle backwards compatibility here.

Note that this PR is the same as https://github.com/apache/pulsar/pull/18712.

### Modifications

* Fix an incorrect reference to trusted certificates.

### Verifying this change

This is a trivial change.

### Documentation

- [x] `no-need-doc` 
  
No docs are required, but a release note might be helpful for some users.


eolivelli (approved):
good catch

Demogorgon314 (approved):

BewareMyPower (approved):

PR #946, opened 2021-12-02
Fix pulsar-io-thread block forever

lordcheng10:
When the requestQueue is full, the pulsar-io thread will always be blocked in: requestQueue.put(), which will cause the requestQueue.remove method to not be called, because the writeAndFlushResponseToClient method is also executed by the pulsar-io thread:
`                ctx.channel().eventLoop().execute(() -> {
                     writeAndFlushResponseToClient(channel);
                 });`
This will cause a lot of close_wait on the server side；

related to:
https://github.com/streamnative/kop/issues/942

PR #1621, opened 2022-12-13
[fix] Use async back pressure to limit RIP

michaeljmarshall:
### Motivation

This is a work in progress to make the `KafkaCommandDecoder` use asynchronous backpressure to limit the number of requests in progress.

### Modifications

* Revert #946.
* Introduce `numQueuedRequestsInProgress` to track number of requests. This variable is only updated from the event loop thread, so it does not need to be synchronized in any way.
* Disable `autoRead` when the channel has reached its limit, and enable auto read when it has gone below its limit.

### Verifying this change

I'll work on adding tests tomorrow.

### Documentation
  
- [x] `no-need-doc` 

This is an internal optimization that does not need to be documented.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaChannelInitializer.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaChannelInitializer.java
@@ -31,7 +31,6 @@ import io.streamnative.pulsar.handlers.kop.utils.delayed.DelayedOperationPurgato
 import io.streamnative.pulsar.handlers.kop.utils.ssl.SSLUtils;
 import java.util.concurrent.TimeUnit;
 import lombok.Getter;
-import org.apache.bookkeeper.common.util.OrderedScheduler;
 import org.apache.pulsar.broker.PulsarService;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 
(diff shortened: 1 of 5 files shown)


lhotari:
The changes make sense. Some improvement areas:
* there's a need to consider the overall backpressure solution. 
* it could be beneficial to have a high-watermark / low-watermark type of solution for backpressure so that channel state doesn't get toggled after each incoming request, when the queue is full and there's a continuous flow of incoming requests. The reason for this seems to be that there's a cost involved in backpressuring the channel and by enabling some sort of batch with the high-watermark / low-watermark, this could improve overall efficiency. [This stackoverflow answer](https://stackoverflow.com/a/25282627/166062) references a blog post that seems to support this theory. 

The 2nd point isn't as important as the 1st point where the overall backpressure solution should be considered.

Some parts of the existing backpressure solution can be seen in https://github.com/streamnative/kop/blob/10d1d920b3c1874bfe8286a150631562e47a6081/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java#L779-L825 . This code was added in #488 .

If there are multiple independent backpressure solutions on the same channel, that would lead to conflicting behavior. 

Regarding the overall backpressure solution, it's a major gap in Pulsar and KOP that Netty's asynchronous backpressure for handling outbound traffic (channelWritabilityChanged events). It seems that the solution added in #488 somewhat covers that without using Netty's built-in solution for not overwhelming outbound buffers.

michaeljmarshall:
Thanks for your review @lhotari. I didn't notice that we were already using the auto read, so I'll re-review the solution there as well as at the low and high water mark.

PR #1642, opened 2022-12-22
Optimize performance of EncodeResult.updateProducerStats and DecodeResult.updateConsumerStats

lhotari:
- reduce unnecessary garbage created by adding some caching

This came up in some profiling

image
image

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete`

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/RequestStats.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/RequestStats.java
@@ -35,6 +35,7 @@ import static io.streamnative.pulsar.handlers.kop.KopServerStats.RESPONSE_BLOCKE
 import static io.streamnative.pulsar.handlers.kop.KopServerStats.SERVER_SCOPE;
 import static io.streamnative.pulsar.handlers.kop.KopServerStats.TOPIC_SCOPE;
 import static io.streamnative.pulsar.handlers.kop.KopServerStats.WAITING_FETCHES_TRIGGERED;
+
 import com.google.common.annotations.VisibleForTesting;
 import io.streamnative.pulsar.handlers.kop.stats.NullStatsLogger;
 import io.streamnative.pulsar.handlers.kop.stats.StatsLogger;
@@ -157,7 +158,8 @@ public class RequestStats {
     private final Map<ApiKeys, StatsLogger> apiKeysToStatsLogger = new ConcurrentHashMap<>();
 
     private final Map<TopicPartition, StatsLogger> cachedLoggersForTopicPartitions = new ConcurrentHashMap<>();
-    private final Map<Pair<TopicPartition, String>, StatsLogger> cachedLoggersForTopicPartitionsAndGroups = new ConcurrentHashMap<>();
+    private final Map<Pair<TopicPartition, String>, StatsLogger> cachedLoggersForTopicPartitionsAndGroups =
+            new ConcurrentHashMap<>();
 
     private final Map<String, RequestStats> cachedRequestStatsForTenants = new ConcurrentHashMap<>();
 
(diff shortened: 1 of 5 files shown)


BewareMyPower (commented):
LGTM. Could you fix the code style check?

lhotari:
> LGTM. Could you fix the code style check?

@BewareMyPower I fixed the code style now.

BewareMyPower (approved):

BewareMyPower:
I think we still need a periodical task to clean up the cache, though I think we can do it in another PR. Does it make sense to you?

lhotari:
> I think we still need a periodical task to clean up the cache, though I think we can do it in another PR. Does it make sense to you?

I doubt that the cache will ever grow boundlessly. the amount of data in the cache is relatively low. Therefore cleaning the cache isn't needed.

BewareMyPower:
There is a point that the size of `cachedLoggersForTopicPartitions` is related to the amount of topics (partitions), including those deleted topics. I think it could be verified by adding a test like:

```
for i = 1 to 1000000:
    topic = randomString()
    admin.create(topic)
    producer.send(msgs)
    admin.delete(topic)
```

lhotari:
> There is a point that the size of `cachedLoggersForTopicPartitions` is related to the amount of topics (partitions), including those deleted topics. I think it could be verified by adding a test like:
> 
> ```
> for i = 1 to 1000000:
>     topic = randomString()
>     admin.create(topic)
>     producer.send(msgs)
>     admin.delete(topic)
> ```

@BewareMyPower yes, valid for that type of use cases. Perhaps adding a bounded size cache with Caffeine could be done in a separate PR as you suggested?

BewareMyPower:
> Perhaps adding a bounded size cache with Caffeine could be done in a separate PR as you suggested?

LGTM. I just thought that we can start cleaning the cache if the size grows over than a limit. But just as you said, the Caffeine would be better.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/RequestStats.java
```diff
@@ -152,6 +157,12 @@ public class RequestStats {
 
     private final Map<ApiKeys, StatsLogger> apiKeysToStatsLogger = new ConcurrentHashMap<>();
 
+    private final Map<TopicPartition, StatsLogger> cachedLoggersForTopicPartitions = new ConcurrentHashMap<>();
```
michaeljmarshall, line 160: Can we keep the cache up to date by removing unloaded topics using `TopicOwnershipListener#whenUnload`?

michaeljmarshall (commented):
My primary concern with an unbounded cache is that load balancing moves topics to new brokers, and that could lead to unnecessary caching.

The optimizations look good to me.

lhotari:
> My primary concern with an unbounded cache is that load balancing moves topics to new brokers, and that could lead to unnecessary caching.
> 
> The optimizations look good to me.

That is planned to be handled in a separate PR as discussed above. The number of topics isn't really unbounded unless there's a malicious user in the system which exploits a unbounded cache to DoS the system. However in that case, many other parts of the system would fall down much earlier. There could be unnecessary items, but the cost of that isn't high.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/RequestStats.java
eolivelli: We definitely must clear it

michaeljmarshall:
@lhotari - I missed the reference to another PR. That makes sense to me.

eolivelli (changes requested):
Overall LGTM

I second Micheal's comments

BewareMyPower:
I will rerun the tests. There might be some flaky tests.

Demogorgon314 (approved):

PR #1645, opened 2022-12-23
[bugfix] Fix regression with the upgrade to 2.8.x: Count corrently the number of partitions in ProduceRequest

eolivelli:
### Motivation

With the upgrade to Kafka 2.8.x we introduced a regression in the handleProduceRequest method, we are not counting
correctly the number of partitions.
If the client sends data for more than one partition for the same topic, then the result is a REQUEST_TIMED_OUT because we are calling "completeOne.run()" once per partition, but we are using the number of topics as the expected number of "completeOne" calls.

### Modifications

Count correctly the number of partitions.

### Verifying this change

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -830,7 +830,12 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
         checkArgument(produceHar.getRequest() instanceof ProduceRequest);
         ProduceRequest produceRequest = (ProduceRequest) produceHar.getRequest();
 
-        final int numPartitions = produceRequest.data().topicData().size();
+        final int numPartitions = produceRequest
+                .data()
+                .topicData()
+                .stream()
+                .mapToInt(t -> t.partitionData().size())
+                .sum();
         if (numPartitions == 0) {
             resultFuture.complete(new ProduceResponse(Collections.emptyMap()));
             return;


BewareMyPower (approved):

BewareMyPower (changes requested):
Could you add a test to avoid the regression?

eolivelli:
Sure, I will.
I wanted to post this PR as soon as possible, because in my fork after the upgrade we found those TIME OUT errors that were without an explanation and it took days to find out the problem.

eolivelli:
I also have a couple of other follow up PRs regarding the upgrade of the kafka clients, in order to improve performances.
I will send new PRs new week

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
```diff
@@ -830,7 +830,12 @@ protected void handleProduceRequest(KafkaHeaderAndRequest produceHar,
         checkArgument(produceHar.getRequest() instanceof ProduceRequest);
         ProduceRequest produceRequest = (ProduceRequest) produceHar.getRequest();
 
-        final int numPartitions = produceRequest.data().topicData().size();
+        final int numPartitions = produceRequest
+                .data()
+                .topicData()
+                .stream()
```
michaeljmarshall, line 836: Streams can be less efficient than iterating directly and adding to a mutable variable. Given that this code is on the hot path, can we use a `for` loop?

michaeljmarshall (commented):
Nice catch @eolivelli.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
eolivelli: While a generally agree on this point, the new Java model for the wire protocol is using this pattern with multiple layers (topics > partitions).
Doing 2 'for' loops will require in any case a similar amount of resources, especially if the data structures are small. The gain probably is negligible.

Please note that we could create some utility method and refactor the codebase. I didn't do it because I wasn't sure that we would be axtuslly able to make the code more readable or efficient.

The code in this form is pretty compact and readable. 
I am leaning towards keeping it like this

BewareMyPower (approved):
I just took a look again. It only affects the time to send produce response and it might be a little hard to add the test. I will merge this fix first.

Issue #1652, opened 2022-12-29
[BUG] - KoP CUP usage is 10 times higher than Pulsar on the broker side


Hongten:
**Describe the bug**
KoP CUP usage is 10 times higher than Pulsar on the broker side

**To Reproduce**
1. Create a topic with 100 partitions.
2. Use a pulsar client to produce data on the topic, and record the broker CPU usage.
3. Use a Kafka client to produce data on the topic, and record the broker CPU usage.

**Expected behavior**
KoP CPU usage should be greater than or equal to the Pulsar itself. 
In this case, when I use the Pulsar client, the broker CPU usage is around 3. When I use Kafka client, the broker CPU usage is around 30. 
Expected result: When I use Kafka client, the broker CPU usage >=3, but not too high.

**Screenshots**
The broker(KoP) flame graph
![image](https://user-images.githubusercontent.com/3076703/209929906-9f801467-7c6c-4546-8e24-bc2133865165.png)
Pulsar client vs. Kafka client(CUP usage)
![image](https://user-images.githubusercontent.com/3076703/209929945-571dfb52-db8c-4773-a1bc-5e1c50609327.png)

**Additional context**

Hongten:
Please refer to the flame graph HTML file.
[broker_flame_graph.html.zip](https://github.com/streamnative/kop/files/10318595/broker_flame_graph.html.zip)

BewareMyPower:
What's your KoP version? It looks like the `sanitizeMetricName`, which converts a metric name to a valid name, took too much time.

![image](https://user-images.githubusercontent.com/18204803/209932050-0766887e-7dc0-4867-a9d4-ee0ca21725e2.png)

However, the `io.prometheus:simpleclient:0.16.0` dependency of KoP master does not call regex related APIs:

```java
    public static String sanitizeMetricName(String metricName) {
        int length = metricName.length();
        char[] sanitized = new char[length];

        for(int i = 0; i < length; ++i) {
            char ch = metricName.charAt(i);
            if (ch != ':' && (ch < 'a' || ch > 'z') && (ch < 'A' || ch > 'Z') && (i <= 0 || ch < '0' || ch > '9')) {
                sanitized[i] = '_';
            } else {
                sanitized[i] = ch;
            }
        }

        return new String(sanitized);
    }
```

Hongten:
KOP- 2.10.1.11

BewareMyPower:
[...]

PR #1654, opened 2022-12-29
[branch-2.10] Use a more efficient implementation of sanitizeMetricName

BewareMyPower:
Fixes https://github.com/streamnative/kop/issues/1652

### Motivation

Pulsar 2.10 or earlier depends on io.prometheus:simpleclient:0.5.0, whose implementation of `Collectors.sanitizeMetricName` uses regex, which is inefficient and costs much CPU.

### Modifications

Migrate the implementation of `sanitizeMetricName` from io:prometheus:simpleclient:0.16.0 to branch-2.10 or earlier.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/stats/PrometheusStatsLogger.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/stats/PrometheusStatsLogger.java
@@ -14,7 +14,6 @@
 package io.streamnative.pulsar.handlers.kop.stats;
 
 import com.google.common.base.Joiner;
-import io.prometheus.client.Collector;
 import java.util.Map;
 import java.util.TreeMap;
 import org.apache.bookkeeper.stats.Counter;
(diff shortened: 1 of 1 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/stats/PrometheusStatsLogger.java
```diff
@@ -78,6 +77,22 @@ private ScopeContext scopeContext(String name) {
     }
 
     private String completeName(String name) {
-        return Collector.sanitizeMetricName(scope.isEmpty() ? name : Joiner.on('_').join(scope, name));
```
Demogorgon314, line 81: Can we upgrade the `prometheus:simpleclient ` version?

BewareMyPower: The dependency is from Pulsar. If we choose to upgrade the `simpleclient`, we have to import the dependency specially. Not sure if there are API compatibility issues.

Regarding the Pulsar side, the dependency cannot be cherry-picked into 2.10.x or earlier, see https://github.com/apache/pulsar/pull/13785#issuecomment-1235189724

Demogorgon314 (approved):

PR #1695, opened 2023-01-16
[oauthclient] Create a zero-dependencies jar:

eolivelli:
- remove Async HTTP client and use the standard JDK Http client
- shade and relocate Jackson Databind, used for JSON

### Motivation

Currently the OAuth client requires Async HTTP Client and Reactive Streams.
This is pretty awkward for users that are not using a dependency management system (like simply dropping the .jar file into the libs directory in order to run CLI tools)

### Modifications

- Remove the dependency from Async HTTP client and use the standard JDK URLConnection
- Shade the Jackson Mapper dependency
- Keep only the necessary classes into the jar (we are inheriting SNAKEYAML and other useless stuff from the parent pom) 

### Verifying this change

This change is already covered by existing tests

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 

This change is not visibile to most of the users  
  
- [ ] `doc` 
  
### DIFF
--- a/oauth-client/pom.xml
+++ b/oauth-client/pom.xml
@@ -29,11 +29,6 @@
   <name>StreamNative :: Pulsar Protocol Handler :: OAuth 2.0 Client</name>
   <description>OAuth 2.0 login callback handler for Kafka client</description>
 
-  <properties>
-    <async-http-client.version>2.12.1</async-http-client.version>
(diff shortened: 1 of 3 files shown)


BewareMyPower:
Replaced by https://github.com/streamnative/kop/pull/1979

PR #1747, opened 2023-03-01
fix: ListOffset interface returns an incorrect result

### LINKED ISSUE #1737: [BUG]The commit offset and partition data expire at the same time, and the ListOffset interface returns an incorrect result
**Describe the bug**
The client requests the commit offset of all partitions according to offsetFetch. If the commit offset of the partition expires, the broker returns the offset of the partition as null, and the client will auto.offset.reset according to the configured policy, and then set the offset of the missing partition to the latest topic Or the earliest offset, get this offset and then pull the data.

For example, the latest offset of a partition in topicA is 10, the oldest offset is 0, and the consumption offset of this partition is also 10. After a period of time, if the commit offset data of __consumer_offset expires and the client cannot get the consumption offset according to the OffsetFetch request, it will request the earliest offset of the partition according to auto.offset.reset=earliest. If the partition data is also It has expired, and the earliest offset returned becomes 9 (one smaller than the latest offset), which should actually be 10

**To Reproduce**
1. Send data to topicA
2. Consumed all the data of topicA
3. Record the partition offset and consumption offset at this time
4. After a period of time (make sure all data has expired)...
5. Use kafka client to query partition offset and consumption offset

**Expected behavior**
In this scenario, the earliest and latest offsets should be the same

**Screenshots**

**Additional context**

keyboardbobo:
The commit offset and partition data expire at the same time, and the ListOffset interface returns an incorrect result

*(If this PR fixes a github issue, please add `Fixes #<xyz>`.)*

Fixes #1737

### Motivation

The commit offset and partition data expire at the same time, and the ListOffset interface returns an incorrect result

### Modifications

If all the data expires, the entry of lac becomes -1, and the actual offset will be 1 smaller than LATEST,  add 1 back

### Verifying this change

*(Please pick either of the following options)*

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

*(example:)*
  - *Added integration tests for end-to-end deployment with large payloads (10MB)*
  - *Extended integration test for recovery after broker failure*

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -1129,7 +1129,10 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
                     return;
                 }
                 if (position.compareTo(lac) > 0) {
-                    partitionData.complete(Pair.of(Errors.NONE, latestOffset));
+                    // If all the data expires, the entry of lac becomes -1,
+                    // and the actual offset will be 1 smaller than LATEST,
+                    // here is special treatment, add 1 back
+                    partitionData.complete(Pair.of(Errors.NONE, latestOffset + 1));
                 } else {
                     MessageMetadataUtils.getOffsetOfPosition(managedLedger, position, false,
                             timestamp, skipMessagesWithoutIndex)


eolivelli (commented):
This is an interesting finding.
I am not sure that we can build a test.

How did you test that this fix actually fixed the problem ?

keyboardbobo:
> This is an interesting finding. I am not sure that we can build a test.
> 
> How did you test that this fix actually fixed the problem ?

@eolivelli 
This is indeed difficult to test, but the problem does exist. I tested the following scenarios, and the producer and consumer offsets all need to meet expectations:
1) __consumer_offset data has not expired, topic data has not expired
2) __consumer_offset data is expired, topic data is not expired
3) __consumer_offset data expires, topic data expires
At present, only the consumer offset queried in the third case is wrong, The third case corresponds to `if (position.compareTo(lac) > 0)`. I don't know if you have a better test method?

PR #1797, opened 2023-04-12
[fix][schema] Return JSON error message for failed request

BewareMyPower:
### Motivation

Sometimes when a schema REST request failed, `buildErrorResponse` might be called and a HTTP response whose content type is `text/plain` will be sent to the client. However, Confluent's Schema Registry client expects the response content is a JSON representation of `ErrorMessage` class:

```java
ErrorMessage errorMessage;
try {
  errorMessage = (ErrorMessage)jsonDeserializer.readValue(es, ErrorMessage.class);
} catch (JsonProcessingException e) {
  errorMessage = new ErrorMessage(JSON_PARSE_ERROR_CODE, e.getMessage());
}
```

https://github.com/confluentinc/schema-registry/blob/7bc2a042f794416f28f2e79598a6abb972243741/client/src/main/java/io/confluent/kafka/schemaregistry/client/rest/RestService.java#L199

For example, when a 404 Not Found error is returned, the error message at client side might look like:

> Unrecognized token ‘Not’: was expecting ‘null’, ‘true’, ‘false’ or NaN

### Modifications

- Migrate `ErrorMessage` from Confluent's Schema Registry, then remove the `ErrorModel` class.
- In `buildErrorResponse`, serialize the code and message into `ErrorMessage`'s JSON representation.
- Reduce the duplicated code by calling `buildErrorResponse` in `buildJsonResponse` and `buildJsonErrorResponse`.
- Add `SchemaRegistryTest#testGetLatestSchemaMetadata` to verify a correct `RestClientException` will be thrown.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/SchemaRegistryManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/SchemaRegistryManager.java
@@ -142,7 +142,7 @@ public class SchemaRegistryManager {
                                 username, role, tenant, topicName);
                         throw new SchemaStorageException("Role " + role + " cannot access topic " + topicName + " "
                                 + "tenant " + tenant + " does not exist (wrong username?)",
-                                HttpResponseStatus.FORBIDDEN.code());
+                                HttpResponseStatus.FORBIDDEN);
                     }
                     Boolean hasPermission = authorizer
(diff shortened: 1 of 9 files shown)


eolivelli (approved):
nice !

PR #1805, opened 2023-04-21
[Schema Registry] Fix the delete operation not work

BewareMyPower:
### Motivation

When I tried to delete a subject, the subject could still be queried even after I restarted the broker.

```bash
$ curl -L http://localhost:8001/subjects
[ test ]
$ curl -X DELETE -L http://localhost:8001/subjects/test
$ curl -L http://localhost:8001/subjects
[ test ]
```

The reason is that the `SchemaStatus.DELETED` is never processed.

### Modifications

In `PulsarSchemaStorage#applyOpToLocalMemory`, when the status is `DELETED`, remove the entry with the same schema id from the `schemas` cache.

Add `SchemaRestApiTest#testDeleteSubject` to verify this change. To verify modifications also work after the restart, expose the schema storage for tests and add `resetSchemaStorage` to clear the storage.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/schema-registry/src/main/java/io/streamnative/pulsar/handlers/kop/schemaregistry/model/impl/PulsarSchemaStorage.java
+++ b/schema-registry/src/main/java/io/streamnative/pulsar/handlers/kop/schemaregistry/model/impl/PulsarSchemaStorage.java
@@ -331,11 +331,15 @@ public class PulsarSchemaStorage implements SchemaStorage, Closeable {
             try {
                 compatibility.put(op.subject, CompatibilityChecker.Mode.valueOf(op.compatibilityMode));
             } catch (IllegalArgumentException err) {
-                log.error("Unrecognized mode, skip op", op);
+                log.error("Unrecognized mode, skip op", err);
             }
         } else {
-            SchemaEntry schemaEntry = op.toSchemaEntry();
-            schemas.put(schemaEntry.id, schemaEntry);
+            if (op.status == SchemaStatus.DELETED) {
+                schemas.remove(op.schemaId);
+            } else {
+                SchemaEntry schemaEntry = op.toSchemaEntry();
+                schemas.put(schemaEntry.id, schemaEntry);
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


schema-registry/src/main/java/io/streamnative/pulsar/handlers/kop/schemaregistry/model/impl/PulsarSchemaStorage.java
```diff
@@ -331,11 +331,15 @@ private void applyOpToLocalMemory(Op op) {
             try {
                 compatibility.put(op.subject, CompatibilityChecker.Mode.valueOf(op.compatibilityMode));
             } catch (IllegalArgumentException err) {
-                log.error("Unrecognized mode, skip op", op);
+                log.error("Unrecognized mode, skip op", err);
             }
         } else {
-            SchemaEntry schemaEntry = op.toSchemaEntry();
-            schemas.put(schemaEntry.id, schemaEntry);
+            if (op.status == SchemaStatus.DELETED) {
```
Demogorgon314, line 337: Will the deleted schema delete on the topic?

BewareMyPower: Yes. Deleting a subject will delete all the schemas that were registered into the subject. See https://docs.confluent.io/platform/current/schema-registry/schema-deletion-guidelines.html#soft-delete-a-schema

And currently KoP does not support deleting a specific version.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/SchemaRestApiTest.java
```diff
@@ -0,0 +1,111 @@
...
+        final var subject = "my-subject";
+        sendHttpRequest("POST", "/subjects/" + subject + "/versions",
+                MAPPER.writeValueAsString(createSchemaRequest));
+
+        assertEquals(getSubjects(), Collections.singletonList(subject));
+        resetSchemaStorage();
+        assertEquals(getSubjects(), Collections.singletonList(subject));
+
+        sendHttpRequest("DELETE", "/subjects/" + subject, null);
+
```
Demogorgon314, line 73:
```suggestion
        assertTrue(getSubjects().isEmpty());
```
We can assert first before resetting schema storage.

Demogorgon314 (dismissed):

Demogorgon314 (approved):

PR #1807, opened 2023-04-23
[Schema Registry] Force authorization when kafkaEnableMultiTenantMetadata is false

BewareMyPower:
### Motivation

Currently when the schema registry and authorization are both enabled, the authorization is only performed when `kafkaEnableMultiTenantMetadata` is false. It's not reasonable because the role of the token must have the write permission to the schema registry topic.

### Modifications

- In `performAuthorizationValidation`, do not check `kafkaEnableMultiTenantMetadata`.
- Add the `testSchemaWrongAuth` to verify that a wrong role cannot be used to create an Avro produce.
- Separate the default namespace and the default schema namespace in `KafkaAuthorizationTestBase` so that the authorization requirements will be clear.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  

Demogorgon314 (dismissed):

Demogorgon314:
@BewareMyPower Can you help resolve conflicts?

BewareMyPower:
Okay, I forgot this PR for some time.

PR #1824, opened 2023-04-27
Fix usage of Pulsar's AuthenticationState object

eolivelli:
The original patch was written by @michaeljmarshall [here](https://github.com/datastax/starlight-for-kafka/pull/85)

### Motivation

There is a breaking API change in upcoming Pulsar versions:
https://github.com/apache/pulsar/pull/19295

This patch allows KOP to run on newer Pulsar versions.

### Modifications

Use the authenticate() method instead of authenticateAsync() 

### Verifying this change

This change is already covered by existing tests, such as *(please describe tests)*.

### Documentation

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/SchemaRegistryManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/SchemaRegistryManager.java
@@ -109,7 +109,7 @@ public class SchemaRegistryManager {
                 AuthData authData = AuthData.of(password.getBytes(StandardCharsets.UTF_8));
                 final AuthenticationState authState = authenticationProvider
                         .newAuthState(authData, null, null);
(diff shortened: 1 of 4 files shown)


BewareMyPower:
I think we don't need to migrate this fix from https://github.com/datastax/starlight-for-kafka/pull/85. From the description:

>  Then when the luna streaming release containing the asynchronous authentication changes gets released, we will upgrade S4K via https://github.com/datastax/starlight-for-kafka/pull/84.

KoP depends on the 3.0.0 release candidate, which already contains the asynchronous authentication changes. This issue seems only related to the dependency of s4k.

eolivelli:
@BewareMyPower you are right.
S4K is still based on Pulsar 2.10.x, we don't need this here.

Issue #1830, opened 2023-04-28
Flaky-test: KafkaListenerNameTest.testConnectListenerNotExist


BewareMyPower:
KafkaListenerNameTest is flaky. The testConnectListenerNotExist test method fails sporadically.

[example failure](https://github.com/streamnative/kop/actions/runs/4827121766/jobs/8600906936)

Attempt 1:

```
Error:  Failures: 
Error:    KafkaListenerNameTest.testConnectListenerNotExist:214->KopProtocolHandlerTestBase.internalCleanup:318->KopProtocolHandlerTestBase.stopBroker:354->KopProtocolHandlerTestBase.stopBroker:350 » ThreadTimeout
```

Attempt 2:

```
Error:  Failures: 
Error:    KafkaListenerNameTest.testConnectListenerNotExist:214->KopProtocolHandlerTestBase.internalCleanup:318->KopProtocolHandlerTestBase.stopBroker:354->KopProtocolHandlerTestBase.stopBroker:350 » ThreadTimeout
```

Attempt 3:

```
Error:  Failures: 
Error:    KafkaListenerNameTest.testConnectListenerNotExist:214->KopProtocolHandlerTestBase.internalCleanup:318->KopProtocolHandlerTestBase.stopBroker:354->KopProtocolHandlerTestBase.stopBroker:350 » ThreadTimeout
Error:    KafkaListenerNameTest.testIpv6ListenerName:236->KopProtocolHandlerTestBase.internalSetup:262->KopProtocolHandlerTestBase.internalSetup:297->KopProtocolHandlerTestBase.startBroker:358->KopProtocolHandlerTestBase.startBroker:367 » PulsarServer
```

PR #1831, opened 2023-04-28
Increase timeout for KafkaListenerNameTest

BewareMyPower:
Fixes https://github.com/streamnative/kop/issues/1830

### Motivation

Different from other tests, the `KafkaListenerNameTest` sets up and cleans up the broker service during the test method, which will take much more time than other tests. `testConnectListenerNotExist` is more flaky because the timeout is 20 seconds while other methods are 30.

### Modifications

Increase all timeout of `KafkaListenerNameTest` to 1 minute.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  

Demogorgon314 (approved):

BewareMyPower:
It can be reproduced in my local env:

image

image

It seems that the synchronous close of `pulsarClient` took much time.

PR #1847, opened 2023-05-12
Fix list offsets for times failure when ledgers are removed by a roll…

zhouxy0809:
…over operation

### Motivation
When lac ledger is expired or removed，KafkaRequestHandler.fetchOffsetByTimestamp call failed and `NonRecoverableLedgerException` will be thrown.
- When lac ledger is expired or removed，the first legerId in managedLedger is greater than the lac ledgerId , then the `asyncFindNewestMatching.startPosition` is lac'sLedgerId:0. 
- `OpFindNewestEntry.find` will read `startPosition`'s entry, then managedLedger will thrown `NonRecoverableLedgerException`

### Modifications

- Handle the NonRecoverableLedgerException in findEntryFailed callback. In this case, return the LEO as the offset because we can treat the topic as empty.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -1231,9 +1231,16 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
             @Override
             public void findEntryFailed(ManagedLedgerException exception,
                                         Optional<Position> position, Object ctx) {
-                log.warn("Unable to find position for topic {} time {}. Exception:",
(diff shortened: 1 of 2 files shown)


BewareMyPower (changes requested):
Could you change the target branch to `master` so that we can cherry-pick it to all release branches?

zhouxy0809:
> Could you change the target branch to `master` so that we can cherry-pick it to all release branches?

I have changed target branch to `master`，and should I change source branch to `master`?

Demogorgon314:
Can you push a new PR base on the master branch? Look like you are using branch-2.9.4 to push the PR.

zhouxy0809:
> Can you push a new PR base on the master branch? Look like you are using branch-2.9.4 to push the PR.

Thanks for your advice

PR #1848, opened 2023-05-12
Use brokerClientTlsEnabled to configure pulsar client

michaeljmarshall:
### Motivation

The `createPulsarClient` method is meant to model the PulsarServer method named `getClient`. However, it uses a different config to determine when to enable TLS. This leads to divergent behavior within the broker. Here is the pulsar code:

https://github.com/apache/pulsar/blob/092819b5c4a68dbc39d9a650d5370af9455e805d/pulsar-broker/src/main/java/org/apache/pulsar/broker/PulsarService.java#L1554-L1555

### Modifications

* Replace `isTlsEnabled` with `isBrokerClientTlsEnabled` when creating pulsar client

### Verifying this change

### Documentation
  
- [x] `no-need-doc`


eolivelli (approved):
Lgtm

BewareMyPower (approved):

Issue #1852, opened 2023-05-17
Flaky-test: KafkaNonPartitionedTopicTest.testNonPartitionedTopic


BewareMyPower:
`KafkaNonPartitionedTopicTest` is flaky. The `testNonPartitionedTopic` test method fails sporadically.

[example failure](https://github.com/streamnative/kop/actions/runs/4998819574/jobs/8961726389?pr=1851)

```
Error:  testNonPartitionedTopic(io.streamnative.pulsar.handlers.kop.KafkaNonPartitionedTopicTest)  Time elapsed: 1.464 s  <<< FAILURE!
org.apache.pulsar.client.admin.PulsarAdminException$PreconditionFailedException: Topic has 1 connected producers/consumers
	at org.apache.pulsar.client.admin.internal.BaseResource.getApiException(BaseResource.java:236)
	at org.apache.pulsar.client.admin.internal.BaseResource$3.failed(BaseResource.java:184)
	at org.glassfish.jersey.client.JerseyInvocation$1.failed(JerseyInvocation.java:882)
	at org.glassfish.jersey.client.JerseyInvocation$1.completed(JerseyInvocation.java:863)
	at org.glassfish.jersey.client.ClientRuntime.processResponse(ClientRuntime.java:[229](https://github.com/streamnative/kop/actions/runs/4998819574/jobs/8961726389?pr=1851#step:13:230))
	at org.glassfish.jersey.client.ClientRuntime.access$200(ClientRuntime.java:62)
	at org.glassfish.jersey.client.ClientRuntime$2.lambda$response$0(ClientRuntime.java:173)
	at org.glassfish.jersey.internal.Errors$1.call(Errors.java:248)
	at org.glassfish.jersey.internal.Errors$1.call(Errors.java:244)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:292)
```

BewareMyPower:
For branch-2.9.5, I pushed a commit (fd67cca) directly to ignore this test for now.

PR #1853, opened 2023-05-17
Fix flaky-test: KafkaNonPartitionedTopicTest.testNonPartitionedTopic

Demogorgon314:
Fix: #1852

### Motivation

When the topic has an active producer or consumer, then we are not able to delete the topic by using `admin.topics().delete(topic)`, we should force delete it.

### Modifications

Use the force option to delete the topic in the test.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  

BewareMyPower (approved):

BewareMyPower (commented):
I just have a question that in which case could the topic have active producers or consumers? Should we fix it by closing the producers and consumers?

Demogorgon314:
@BewareMyPower 

I checked https://github.com/apache/pulsar/blob/master/pulsar-broker/src/main/java/org/apache/pulsar/broker/service/persistent/PersistentTopic.java#L1311, but currently, I do not find the reasons why there is active producer/consumer, maybe we can check later for this issue.

PR #1854, opened 2023-05-18
Fix list offsets for times failure when ledgers are removed by a rollover operation

zhouxy0809:
### Motivation
When lac ledger is expired or removed，KafkaRequestHandler.fetchOffsetByTimestamp call failed and `NonRecoverableLedgerException` will be thrown.
- When lac ledger is expired or removed，the first legerId in managedLedger is greater than the lac ledgerId , then the `asyncFindNewestMatching.startPosition` is lac'sLedgerId:0. 
- `OpFindNewestEntry.find` will read `startPosition`'s entry, then managedLedger will thrown `NonRecoverableLedgerException`

### Modifications

- Handle the NonRecoverableLedgerException in findEntryFailed callback. In this case, return the LEO as the offset because we can treat the topic as empty.

### Verifying this change

  - Extended integration test for `MultiLedgerTest.testListOffsetForEmptyRolloverLedger` 

### Documentation
  
- [x] `no-need-doc`

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaRequestHandler.java
@@ -1252,9 +1252,16 @@ public class KafkaRequestHandler extends KafkaCommandDecoder {
             @Override
             public void findEntryFailed(ManagedLedgerException exception,
                                         Optional<Position> position, Object ctx) {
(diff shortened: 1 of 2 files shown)


zhouxy0809:
@BewareMyPower @Demogorgon314 This pr is build successfully in my fork `master` branch, but failed unit test cases do not appear in my fork `master` branch and this failed cases is not in my committed code.

BewareMyPower:
It might be a flaky test. I will review this PR later.

BewareMyPower (approved):

PR #1871, opened 2023-05-31
[improvement] Use ByteBuf for Request serialization

Demogorgon314:
### Motivation

Use ByteBuf for Request serialization

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
@@ -655,12 +655,6 @@ public abstract class KafkaCommandDecoder extends ChannelInboundHandlerAdapter {
                 this.header, this.request, this.remoteAddress);
         }
 
-        public void bufferReleased() {
-            if (!released.compareAndSet(false, true)) {
-                throw new IllegalStateException("Already released");
-            }
-        }
-
         public void close() {
            if (!released.compareAndSet(false, true)) {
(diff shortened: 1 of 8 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
```diff
@@ -650,8 +655,17 @@ public String toString() {
                 this.header, this.request, this.remoteAddress);
         }
 
+        public void bufferReleased() {
+            if (!released.compareAndSet(false, true)) {
```
gaoran10, line 659: I'm not sure about this method, there is no release operation but changing the released state.

gaoran10 (commented):
I found there are some new changes, could you take a look at this commit? https://github.com/datastax/starlight-for-kafka/commit/a61c98a75967e2ff846b865882e0374669d6c6c4

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaCommandDecoder.java
Demogorgon314: Good point, I find we don't need to use this method. They are using this method on the proxy module.

gaoran10 (approved):

BewareMyPower (approved):

PR #1887, opened 2023-06-06
[refactor] Decoupling the offset topic I/O from GroupMetadataManager

BewareMyPower:
### Motivation

Currently the offset topic I/O operations are deeply coupled with GroupMetadataManager so that the code is hard to maintain.

### Modifications

Add a `CompactedPartitionedTopic` class as the abstraction to read or write a compacted partitioned topic with the following public APIs:
- sendAsync: write a key-value as well as the timestamp to a partition
- readToLatest: read until the latest message of a partition
- remove: remove the cached producer and reader of a partition
- close: close the resources

Then, use this class to replace the existing offset topic I/O logic in `GroupMetadataManager`.

Add a `CompactedPartitionedTopicTest` to simulate sending messages and reading message before or after the compaction.

### Important Changes

Before this PR, when the `GroupMetadataManager` reads the offset topic, it writes an empty `ByteBuffer` first, then get the MessageId to read until the latest message's MessageId reaches this empty message.

After this PR, it reads in a loop until no messages are available. To fix the compatibility issue, the empty messages will be ignored.

Another important change is that this PR don't cache futures of producers and readers and then call `thenXxx` APIs to send messages. Because the `CompletableFuture` stores pending callbacks in a stack.

```java
producerFuture.thenApply(producer -> producer.sendAsync(msg0));
producerFuture.thenApply(producer -> producer.sendAsync(msg1));
producerFuture.thenApply(producer -> producer.sendAsync(msg2));
```

For example, the code above might send `msg2`, `msg1`, `msg0` in order, which is unexpected.
### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/tests/src/test/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupMetadataManagerTest.java
+++ b/tests/src/test/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupMetadataManagerTest.java
@@ -37,6 +37,7 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import io.streamnative.pulsar.handlers.kop.KafkaProtocolHandler;
 import io.streamnative.pulsar.handlers.kop.KopProtocolHandlerTestBase;
+import io.streamnative.pulsar.handlers.kop.SystemTopicClient;
 import io.streamnative.pulsar.handlers.kop.coordinator.group.GroupMetadata.CommitRecordMetadataAndOffset;
 import io.streamnative.pulsar.handlers.kop.coordinator.group.GroupMetadataManager.BaseKey;
 import io.streamnative.pulsar.handlers.kop.coordinator.group.GroupMetadataManager.GroupMetadataKey;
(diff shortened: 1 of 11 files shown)


tests/src/test/java/io/streamnative/pulsar/handlers/kop/coordinator/group/GroupMetadataManagerTest.java
```diff
@@ -558,12 +539,7 @@ public void testGroupLoadedWithPendingCommits() throws Exception {
         byte[] key = groupMetadataKey(groupId);
 
 
-        Producer<ByteBuffer> producer = groupMetadataManager.getOffsetsTopicProducer(groupPartitionId).get();
-        producer.newMessage()
-            .keyBytes(key)
-            .value(buffer)
-            .eventTime(Time.SYSTEM.milliseconds())
-            .send();
+        groupMetadataManager.storeOffsetMessage(groupPartitionId, key, buffer);
```
Demogorgon314, line 542: Do we need to wait for it send success?

Demogorgon314 (approved):

PR #1906, opened 2023-06-16
[transaction] Producer state manager snapshot recovery - Part-1: Add snapshot I/O buffer

Demogorgon314:
This PR is a part of the Producer state manager snapshot recovery implementation, it cherry-picks from https://github.com/datastax/starlight-for-kafka/pull/29.

### Motivation

Introduce the `ProducerStateManagerSnapshotBuffer`. 
It can write and read the latest snapshot from a specified topic.

This PR introduces two different `ProducerStateManagerSnapshotBuffer` implementations:
* `PulsarTopicProducerStateManagerSnapshotBuffer`: Store the snapshot into the topic.
* `PulsarPartitionedTopicProducerStateManagerSnapshotBuffer` : Store the snapshot in a partitioned topic.

### Modifications

* Introduce `ProducerStateManagerSnapshotBuffer` interface.
* Provide two diffident implementations for `ProducerStateManagerSnapshotBuffer`.
* Add units test to cover the `Buffer`.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  

gaoran10 (approved):
LGTM

Demogorgon314:
@BewareMyPower Can you help take a look at this PR?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PulsarTopicProducerStateManagerSnapshotBuffer.java
```diff
@@ -0,0 +1,342 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.apache.kafka.common.errors.NotLeaderOrFollowerException;
+import org.apache.pulsar.client.api.Message;
+import org.apache.pulsar.client.api.MessageId;
+import org.apache.pulsar.client.api.Producer;
+import org.apache.pulsar.client.api.Reader;
+import org.apache.pulsar.common.util.FutureUtil;
+
+@Slf4j
+public class PulsarTopicProducerStateManagerSnapshotBuffer implements ProducerStateManagerSnapshotBuffer {
```
BewareMyPower, line 43: Is it possible to reuse `CompactedPartitionedTopic` for it? We can avoid much duplicated code.

Demogorgon314: Maybe we can try to reuse it, but I think we can follow the same implementation as s4k right now, and do some refactoring after this feature is done.

BewareMyPower: Okay.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/MemoryProducerStateManagerSnapshotBuffer.java
```diff
@@ -0,0 +1,43 @@
...
+package io.streamnative.pulsar.handlers.kop.storage;
+
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Use memory to store the latest snapshot.
+ */
+public class MemoryProducerStateManagerSnapshotBuffer implements ProducerStateManagerSnapshotBuffer {
```
BewareMyPower, line 23: We should remove this class and related tests. Because it is only used in tests, which is meaningless.

BewareMyPower (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/MemoryProducerStateManagerSnapshotBuffer.java
BewareMyPower: Here is a similar example: https://github.com/streamnative/kop/pull/1805

There is no bug with `MemorySchemaStorage` because it's protected by tests. However, it is never used out of tests. `PulsarSchemaStorage` is actually and always used and it's buggy.

eolivelli: +1

Demogorgon314: Good point, addressed.

BewareMyPower (approved):

PR #1917, opened 2023-06-25
[transaction] Recover from failures during endTransaction after writing to storage

gaoran10:
(cherry picked from commit 7f59a829bfdfb5cffd7f4eda0dc8d549048c9e59) refer to https://github.com/datastax/starlight-for-kafka/commit/7f59a829bfdfb5cffd7f4eda0dc8d549048c9e59

### Motivation

Recover from failures during endTransaction after writing to storage.

### Modifications

Reset the transaction pending state from failures during endTransaction.  

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java
@@ -335,8 +335,10 @@ public class TransactionCoordinator {
                     false,
                     errors -> {
                         if (errors != Errors.NONE) {
+                            log.error("Cannot initProducer {} due to {} error", transactionalId, errors);
                             responseCallback.accept(initTransactionError(errors));
                         } else {
+                            // reply to client and let it backoff and retry
                             responseCallback.accept(initTransactionError(Errors.CONCURRENT_TRANSACTIONS));
                         }
                     });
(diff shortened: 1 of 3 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java
```diff
@@ -613,6 +627,10 @@ public void fail(Errors errors) {
                             if (epochAndMetadata.getCoordinatorEpoch() == coordinatorEpoch) {
                                     // This was attempted epoch fence that failed, so mark this state on the metadata
                                 epochAndMetadata.getTransactionMetadata().setHasFailedEpochFence(true);
+
+                                // this line is not present in Kafka code base ?
+                                epochAndMetadata.getTransactionMetadata().setPendingState(Optional.empty());
```
gaoran10, line 632: This is the key point.

Demogorgon314: Okay, but why do we need to set the pending state to empty?

gaoran10: If don't reset the pending state, it seems that the transaction coordinator will throw a concurrent exception when re-end the transaction, refer to [here](https://github.com/streamnative/kop/blob/master/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionCoordinator.java#L669), but there is no test to cover this case yet.

Demogorgon314:
It looks like the test failed. Can you check the failed test?

Demogorgon314 (approved):

PR #1926, opened 2023-06-28
[improve] Pass group ID to authorizer when using OAuth

Demogorgon314:
### Motivation
Currently, the KoP only supports checking the topic permission when doing the consume authorization, but Kafka support checking the topic and group ID permission.

Before introducing this change, let's understand why KoP can't check group ID permission.

When Kafka does the consume authorization check, it will first check the group ID permission in `handleJoinGroupRequest` method, then it will check the topic permission in the `handleFetchRequest` method.

However, Pulsar is using another way to check consume permission. See the `org.apache.pulsar.broker.authorization.AuthorizationService#canConsumeAsync` method, it requires passing the topic name and subscription to check both permissions in the same place, and the topic name is required param.
```
public CompletableFuture<Boolean> canConsumeAsync(TopicName topicName, String role, AuthenticationDataSource authenticationData,String subscription)
```

If we follow the Kafka way to check the consumer permission, we can't get the topic name when joining a group since the join group request does not contain the topic name. So we have to authorize permission in the fetch request. 

However, to authorization consume permission in the fetch request, we can only get the topic name in this request. In this case, we can't authorize the group ID.

### Modifications
Since we can't get group ID when handling fetch requests, we need to find a way to pass through group ID when doing the authentication.

In OAuth, we have a `credentials_file.json` file that needs to be config, for example:
```
{
	"client_id": "my-id",
	"client_secret": "my-secret",
	"tenant": "my-tenant"
}
```
Here we can add a new parameter into the config:
```
{
	"client_id": "my-id",
	"client_secret": "my-secret",
	"tenant": "my-tenant",
        "group_id": "my-group-id"
}
```
Then we can add these parameters to `SaslExtensions`, and send it to the broker.

### Documentation

Check the box below.

Need to update docs? 

- [x] `doc-required` 
  
- [ ] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaServiceConfiguration.java
```diff
@@ -113,6 +113,13 @@ public class KafkaServiceConfiguration extends ServiceConfiguration {
     )
     private boolean kafkaEnableMultiTenantMetadata = true;
 
+    @FieldContext(
+        category = CATEGORY_KOP,
+        required = true,
+        doc = "Use to enable/disable Kafka authorization force groupId check."
+    )
+    private boolean kafkaEnableAuthorizationForceGroupIdCheck = false;
```
Demogorgon314, line 121: @BewareMyPower Do you have suggestions about this configuration?

BewareMyPower: No, we can improve the docs about this config later.

BewareMyPower (changes requested):
We need to prevent consumers from subscribing group-B if they have configured the permission for group-A.

tests/src/test/java/io/streamnative/pulsar/handlers/kop/SaslOAuthKopHandlersWithGroupId.java
```diff
@@ -0,0 +1,202 @@
...
+import org.apache.pulsar.common.policies.data.AuthAction;
+import org.apache.pulsar.common.policies.data.TenantInfo;
+import org.testng.Assert;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+import sh.ory.hydra.ApiException;
+
+@Slf4j
+public class SaslOAuthKopHandlersWithGroupId extends SaslOAuthBearerTestBase {
```
BewareMyPower, line 46: Please rename it with `SaslOAuthKopHandlersWithGroupIdTest`.

oauth-client/src/main/java/io/streamnative/pulsar/handlers/kop/security/oauth/ClientConfig.java
```diff
@@ -58,8 +76,43 @@ public ClientConfig(Map<String, String> configs) {
...
+    }
+
+    @Getter
+    @ToString
+    @EqualsAndHashCode
+    @NoArgsConstructor
+    @AllArgsConstructor
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    public static class ClientInfo {
```
BewareMyPower, line 104: I think it would be better to move `ClientInfo` out of `ClientConfig` because `ClientInfo` is also used in `ClientCredentialsFlow`.

oauth-client/src/main/java/io/streamnative/pulsar/handlers/kop/security/oauth/ClientConfig.java
```diff
@@ -58,8 +76,43 @@ public ClientConfig(Map<String, String> configs) {
...
+            this.clientInfo = loadPrivateKey();
+        } catch (IOException e) {
+            throw new IllegalArgumentException(String.format(
+                    "failed to load client credentials from %s: %s", credentialsUrlString, e.getMessage()));
+        }
         this.audience = configs.getOrDefault(OAUTH_AUDIENCE, null);
         this.scope = configs.getOrDefault(OAUTH_SCOPE, null);
     }
+
+    @VisibleForTesting
```
BewareMyPower, line 89: I didn't find `loadPrivateKey` is used in tests.

oauth-client/src/main/java/io/streamnative/pulsar/handlers/kop/security/oauth/OauthLoginCallbackHandler.java
```diff
@@ -68,6 +74,8 @@ public void handle(Callback[] callbacks) throws IOException, UnsupportedCallback
                 } catch (KafkaException e) {
                     throw new IOException(e.getMessage(), e);
                 }
+            } else if (callback instanceof SaslExtensionsCallback) {
```
BewareMyPower, line 77: In which case could `callback` be `SaslExtensionsCallback`? It seems to leverage the [KIP-342](https://cwiki.apache.org/confluence/display/KAFKA/KIP-342%3A+Add+support+for+Custom+SASL+extensions+in+OAuthBearer+authentication) feature, which is only available for Kafka client 2.1 or later.

If so, we need to document that the latest `oauth-client` module could only be used for Kafka client 2.1 or later.

Demogorgon314: Yes, we need to use Kafka client 2.1 or later.

BewareMyPower:
@Demogorgon314 Could you resolve the conflicts?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/oauth/KopOAuthBearerSaslServer.java
```diff
@@ -28,9 +30,21 @@
...
 import org.apache.kafka.common.security.oauthbearer.internals.OAuthBearerClientInitialResponse;
+import org.apache.kafka.common.utils.Utils;
 
+/**
+ * {@code SaslServer} implementation for SASL/OAUTHBEARER in Kafka. An instance
+ * of {@link OAuthBearerToken} is available upon successful authentication via
+ * the negotiated property "{@code OAUTHBEARER.token}"; the token could be used
+ * in a custom authorizer (to authorize based on JWT claims rather than ACLs,
+ * for example).
+ */
```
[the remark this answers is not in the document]
Demogorgon314: Updated.

Demogorgon314:
@BewareMyPower Resolved.

oauth-client/src/test/java/io/streamnative/pulsar/handlers/kop/security/oauth/ClientCredentialsFlowTest.java
```diff
@@ -35,23 +35,23 @@ public class ClientCredentialsFlowTest {
...
                 "http://localhost:4444",
                 Objects.requireNonNull(
                         getClass().getClassLoader().getResource("private_key.json")).toString()
-        ));
-        final ClientCredentialsFlow.ClientInfo clientInfo = flow.loadPrivateKey();
+        );
+        ClientInfo clientInfo = clientConfig.getClientInfo();
         Assert.assertEquals(clientInfo.getId(), "my-id");
         Assert.assertEquals(clientInfo.getSecret(), "my-secret");
-        Assert.assertEquals(clientInfo.getTenant(), "my-tenant");
```
BewareMyPower, line 54: Why did you remove this line?

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/security/auth/SimpleAclAuthorizer.java
```diff
@@ -152,8 +158,26 @@ public CompletableFuture<Boolean> canProduceAsync(KafkaPrincipal principal, Reso
...
+            if (StringUtils.isBlank(principal.getGroupId())) {
+                return CompletableFuture.completedFuture(false);
+            }
+            boolean isSameGroup = Objects.equals(principal.getGroupId(), resource.getName());
+            if (log.isDebugEnabled()) {
+                log.debug("Principal [{}] for resource [{}] isSameGroup [{}]", principal, resource, isSameGroup);
+            }
+            return CompletableFuture.completedFuture(isSameGroup);
+        }
+        return CompletableFuture.completedFuture(true);
```
BewareMyPower, line 180: Not a request change, but it would be better if you use guard clause.

```java
        if (!forceCheckGroupId) {
            return CompletableFuture.completedFuture(true);
        }
        // ...
```

oauth-client/src/test/java/io/streamnative/pulsar/handlers/kop/security/oauth/ClientCredentialsFlowTest.java
Demogorgon314: Added back.

BewareMyPower (approved):

PR #1945, opened 2023-07-05
[branch-2.10.4][improve] Pass group ID to authorizer when using OAuth (#1926)

Demogorgon314:
### Motivation
Currently, the KoP only supports checking the topic permission when doing the consume authorization, but Kafka support checking the topic and group ID permission.

Before introducing this change, let's understand why KoP can't check group ID permission.

When Kafka does the consume authorization check, it will first check the group ID permission in `handleJoinGroupRequest` method, then it will check the topic permission in the `handleFetchRequest` method.

However, Pulsar is using another way to check consume permission. See the
`org.apache.pulsar.broker.authorization.AuthorizationService#canConsumeAsync` method, it requires passing the topic name and subscription to check both permissions in the same place, and the topic name is required param.
```
public CompletableFuture<Boolean> canConsumeAsync(TopicName topicName, String role, AuthenticationDataSource authenticationData,String subscription)
```

If we follow the Kafka way to check the consumer permission, we can't get the topic name when joining a group since the join group request does not contain the topic name. So we have to authorize permission in the fetch request.

However, to authorization consume permission in the fetch request, we can only get the topic name in this request. In this case, we can't authorize the group ID.

### Modifications
Since we can't get group ID when handling fetch requests, we need to find a way to pass through group ID when doing the authentication.

In OAuth, we have a `credentials_file.json` file that needs to be config, for example:
```
{
	"client_id": "my-id",
	"client_secret": "my-secret",
	"tenant": "my-tenant"
}
```
Here we can add a new parameter into the config:
```
{
	"client_id": "my-id",
	"client_secret": "my-secret",
	"tenant": "my-tenant",
        "group_id": "my-group-id"
}
```
Then we can add these parameters to `SaslExtensions`, and send it to the broker.

(cherry picked from commit e108e44621decef8221b44f37322c59f8b674be5)

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/tests/src/test/java/io/streamnative/pulsar/handlers/kop/SaslOAuthKopHandlersWithGroupIdTest.java
+++ b/tests/src/test/java/io/streamnative/pulsar/handlers/kop/SaslOAuthKopHandlersWithGroupIdTest.java
@@ -1,0 +1,203 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.streamnative.pulsar.handlers.kop;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
+
+import com.google.common.collect.Sets;
+import io.streamnative.pulsar.handlers.kop.security.oauth.OauthLoginCallbackHandler;
+import io.streamnative.pulsar.handlers.kop.security.oauth.OauthValidatorCallbackHandler;
+import java.io.IOException;
+import java.net.URL;
+import java.time.Duration;
+import java.util.Collections;
+import java.util.Properties;
+import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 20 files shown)


tests/src/test/java/io/streamnative/pulsar/handlers/kop/SaslOAuthKopHandlersWithGroupIdTest.java
```diff
@@ -0,0 +1,203 @@
...
+        admin.namespaces().grantPermissionOnNamespace(namespace, role, Collections.singleton(AuthAction.produce));
+
+        final Properties consumerProps = newKafkaConsumerProperties();
+        internalConfigureOAuth2(consumerProps, clientCredentialPath);
+        final KafkaConsumer<String, String> consumer = new KafkaConsumer<>(consumerProps);
+        consumer.subscribe(Collections.singleton(topic));
+
+        admin.namespaces().grantPermissionOnNamespace(namespace, role, Sets.newHashSet(AuthAction.consume));
+        // Only have consume permission, can't consume without subscription permission.
+        Assert.assertThrows(GroupAuthorizationException.class, () -> consumer.poll(Duration.ofSeconds(5)));
```
BewareMyPower, line 131: Could you verify why the `TopicAuthorizationException` is thrown here?

```
 SaslOAuthKopHandlersWithGroupIdTest.testGrantAndRevokePermissionWithGroupId:131 Expected GroupAuthorizationException to be thrown, but TopicAuthorizationException was thrown
```

Demogorgon314: Fixed

BewareMyPower (approved):

PR #1960, opened 2023-07-13
[feature][transaction] Support turn off txn state recovery

gaoran10:
### Motivation

Support turn off txn state recovery.

### Modifications

Add a configuration to control whether kafkaTransactionStateProducer recovery.
Use recovery executor to complete ProducerStateManagerSnapshot reads.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
@@ -123,7 +123,7 @@ public class KafkaProtocolHandler implements ProtocolHandler, TenantContextManag
     private final Map<String, GroupCoordinator> groupCoordinatorsByTenant = new ConcurrentHashMap<>();
     private final Map<String, TransactionCoordinator> transactionCoordinatorByTenant = new ConcurrentHashMap<>();
 
-
+    @Getter
     private OrderedExecutor recoveryExecutor;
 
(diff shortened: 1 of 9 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/KafkaProtocolHandler.java
```diff
@@ -123,7 +123,7 @@ public class KafkaProtocolHandler implements ProtocolHandler, TenantContextManag
     private final Map<String, GroupCoordinator> groupCoordinatorsByTenant = new ConcurrentHashMap<>();
     private final Map<String, TransactionCoordinator> transactionCoordinatorByTenant = new ConcurrentHashMap<>();
 
-
+    @Getter
```
Demogorgon314, line 126: Does the `get` method only use for tests? Should we add a `VisibleForTesting` annotation?

gaoran10: Fixed

BewareMyPower (changes requested):
When should we turn off the txn state recovery?

gaoran10:
This switch is used for troubleshooting, we don't need it now, so close this PR.

PR #1977, opened 2023-07-24
[fix][transaction] Handle some failures in transaction

gaoran10:
### Modifications

Handle some failures while writing transaction logs and markers.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionMarkerChannelManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionMarkerChannelManager.java
@@ -191,25 +191,18 @@ public class TransactionMarkerChannelManager {
         }
         ensureDrainQueuedTransactionMarkersActivity();
         CompletableFuture<TransactionMarkerChannelHandler> result =
-                handlerMap.computeIfAbsent(socketAddress, address -> {
-            CompletableFuture<TransactionMarkerChannelHandler> handlerFuture = new CompletableFuture<>();
-            ChannelFutures.toCompletableFuture(bootstrap.connect(socketAddress))
-                    .thenAccept(channel -> {
-                        handlerFuture.complete(
-                                (TransactionMarkerChannelHandler) channel.pipeline().get("txnHandler"));
-                    }).exceptionally(e -> {
-                        handlerFuture.completeExceptionally(e);
-                        return null;
-                    });
+                handlerMap.computeIfAbsent(socketAddress, address -> new CompletableFuture<>());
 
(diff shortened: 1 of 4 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/ProducerStateManager.java
```diff
@@ -178,7 +178,7 @@ void updateAbortedTxnsPurgeOffset(long abortedTxnsPurgeOffset) {
     }
 
     long maybePurgeAbortedTx() {
-        if (mapEndOffset == -1) {
+        if (mapEndOffset == -1 || kafkaTxnPurgeAbortedTxnIntervalSeconds <= 0) {
             return 0;
         }
         long now = System.currentTimeMillis();
```
Demogorgon314, line 184: I find the `deltaFromLast` has been divided by 1000 twice. We should only divide it once.

gaoran10: It seems that this issue had been fixed.

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionMarkerChannelManager.java
```diff
@@ -200,8 +201,21 @@ public CompletableFuture<TransactionMarkerChannelHandler> getChannel(InetSocketA
                         handlerFuture.completeExceptionally(e);
                         return null;
                     });
+
+            // do not cache failed futures
+            handlerFuture.exceptionally(error -> {
+                log.error("getChannel failed {} {}", socketAddress, error.getMessage(), error);
+                handlerMap.remove(socketAddress, handlerFuture);
```
Demogorgon314, line 208: Removing the value in the map may cause `Recursive update` error.

Demogorgon314 (changes requested):

kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/coordinator/transaction/TransactionMarkerChannelManager.java
Demogorgon314: Here is an example:
```java
ConcurrentHashMap<String, String> map = new ConcurrentHashMap<>();
        map.computeIfAbsent("test", key -> {
            map.remove("test");
            return "test";
        });
```

PR #1982, opened 2023-07-27
Do not throw recursive update exception when producer state recovery failed

BewareMyPower:
### Motivation

When transaction is enabled, `PartitionLog#initialise` will recover the state from the local snapshot. It's an asynchronous operation that could fail. In this case, an "recursive update" `IllegalStateException` will be thrown, which is unexpected.

```
Suppressed: java.lang.IllegalStateException: Recursive update
    at java.util.concurrent.ConcurrentHashMap.replaceNode(ConcurrentHashMap.java:1167) ~[?:?]
    at java.util.concurrent.ConcurrentHashMap.remove(ConcurrentHashMap.java:1552) ~[?:?]
    at io.streamnative.pulsar.handlers.kop.storage.PartitionLogManager.lambda$getLog$0(PartitionLogManager.java:88) ~[?:?]
```

The reason is that in `PartitionLogManager#getLog`, `logMap.remove` is called in the callback of `whenComplete`, which could be called in the same thread. Then the `remove` method is just called in the 2nd argument of `computeIfAbsent`.

https://github.com/streamnative/kop/blob/3602c9e826d903d97091af1cc608b9d88c1b8cf3/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLogManager.java#L88

### Modifications

Store the future of `PartitionLog` in `PartitionLogManager`, move the `remove` call out of the `computeIfAbsent` in the `exceptionally` callback of `ReplicaManager#getPartitionLog`.

### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLogManager.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLogManager.java
@@ -13,15 +13,15 @@
  */
 package io.streamnative.pulsar.handlers.kop.storage;
 
-import com.google.common.collect.Maps;
+import com.google.common.annotations.VisibleForTesting;
 import io.streamnative.pulsar.handlers.kop.KafkaServiceConfiguration;
 import io.streamnative.pulsar.handlers.kop.KafkaTopicLookupService;
 import io.streamnative.pulsar.handlers.kop.RequestStats;
 import io.streamnative.pulsar.handlers.kop.utils.KopTopic;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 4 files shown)


kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLogManager.java
```diff
@@ -60,62 +60,43 @@ public PartitionLogManager(KafkaServiceConfiguration kafkaConfig,
...
+                kafkaConfig, requestStats, time, topicPartition, key, entryFilters, kafkaTopicLookupService,
+                prodPerTenant, recoveryExecutor
+        ).initialise());
+        // We need to wait here to avoid the message disorder
+        try {
+            future.get();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        } catch (ExecutionException e) {
+            log.error("Failed to get PartitionLog for {} under {}", topicPartition, namespacePrefix, e.getCause());
```
gaoran10, line 84: Do we need to remove the failed future?

BewareMyPower: It will be removed in the `exceptionally` callback of the future returned.

BewareMyPower: This PR has a serious problem so that it's draft now.

PR #1992, opened 2023-08-03
Fix wrong offset increment for Sarama message set

BewareMyPower:
### Motivation

When Sarama producers send messages to KoP with message set, e.g.

```golang
conf := sarama.NewConfig()
conf.Version = sarama.V0_10_2_2
conf.Producer.Compression = sarama.CompressionLZ4
producer, err := sarama.NewAsyncProducer([]string{"localhost:15003"}, conf)
```

The `PartitionLog#analyzeAndValidateRecords` method could parse a wrong `LogAppendInfo#numMessages` result. It's because this method uses `RecordBatch#lastOffset` to get the latest offset.

https://github.com/streamnative/kop/blob/3b22e79764ca22107228ec2f74590f0769bd9fd9/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java#L1028

However, when Sarama client handles the message set (v0 or v1 records), it only writes a offset of zero value:

https://github.com/IBM/sarama/blob/c10bd1e5709a7b47729b445bc98f2e41bc7cc0a8/produce_request.go#L107

In this case, every batch is counted as 1 message, which might results in a wrong offset. Assuming there are 2 batches whose size is 5, the offsets written in the entries could be:

- Entry 0: index=0
- Entry 1: index=1
- LEO: 2

The correct offsets should be:

- Entry 0: index=0
- Entry 1: index=5
- LEO: 10

The wrong LEO could make the offset check in
`PartitionLog#checkOffsetOutOfRange` fail and then consumers will reset the offset and consume duplicated messages.

### Modifications

When the `lastOffset` is 0, iterate over the records in the batch to compute the number of messages.

Ideally we should add a Golang client test to avoid the regression, in this patch, we only add a `SaramaCompressedV1Records` class to simulate the behavior of Sarama. Then add the unit tests.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/storage/PartitionLog.java
@@ -78,6 +78,7 @@ import org.apache.kafka.common.errors.UnknownServerException;
 import org.apache.kafka.common.protocol.Errors;
 import org.apache.kafka.common.record.CompressionType;
 import org.apache.kafka.common.record.MemoryRecords;
+import org.apache.kafka.common.record.Record;
 import org.apache.kafka.common.record.RecordBatch;
 import org.apache.kafka.common.record.Records;
 import org.apache.kafka.common.requests.FetchRequest;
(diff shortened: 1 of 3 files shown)


BewareMyPower:
I think for clients of other languages, we can add the tests into our own private test repo in future.

https://github.com/streamnative/kop-tests

@Demogorgon314 @gaoran10 @codelipenghui

Demogorgon314 (approved):
Good catch!

PR #1996, opened 2023-08-12
Add Time based metric exipration.

### LINKED ISSUE #1956: [BUG] Memory leak of the map in ThreadLocalAccessor
**Describe the bug**

There are many entries in `ThreadLocalAccessor#map`, whose size is 2885059.

https://github.com/streamnative/kop/blob/ae5d500b293795e9f3b7680cfb07d05366409860/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/stats/ThreadLocalAccessor.java#L21-L23

**To Reproduce**

It happened in SN's test environment with KoP 2.11.1.3. The process ran with `-XX:ActiveProcessorCount=4` so it does not hit the issue here: https://github.com/streamnative/kop/issues/1278. From the thread dump, we can see only one thread (`metrics-90-1`) for it.

**Expected behavior**

The map should not take so much memory.

**Screenshots**

image

**Additional context**

lifepuzzlefun:
Fixes #1956

### Motivation
FastThreadLocal wont cleanup until thread exit. so we can just check if the stats is not recorded a long time if expired just remove from threadlocal

### Modifications
add time check logic if long time have no record just remove the stats.

### Documentation

Check the box below.

Need to update docs? 

- [ ] `doc-required` 
  
- [x] `no-need-doc` 
  
  (Please explain why)
  
- [ ] `doc` 
  
### DIFF
--- a/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/stats/DataSketchesOpStatsLogger.java
+++ b/kafka-impl/src/main/java/io/streamnative/pulsar/handlers/kop/stats/DataSketchesOpStatsLogger.java
@@ -101,7 +101,7 @@ public class DataSketchesOpStatsLogger implements OpStatsLogger {
         throw new UnsupportedOperationException();
     }
 
-    public void rotateLatencyCollection() {
+    public void rotateLatencyCollection(long expiredTimeSeconds) {
         // Swap current with replacement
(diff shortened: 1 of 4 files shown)


BewareMyPower (approved):
Though it does not fix the issue directly, it's a good way to prevent such issues. So it LGTM.

PR #2035, opened 2023-12-29
[CI] Upload the NAR package when pushing a tag

BewareMyPower:


BewareMyPower:
An example run: https://github.com/BewareMyPower/kop/actions/runs/7357165859

It only packages the NAR file, not including the `oauth-clients` and `kafka-payload-processor` packages, which should be used by importing a dependency to Maven.

codelipenghui (approved):

Demogorgon314 (approved):

